PR #137, opened 2020-03-04
Fix optional parameter for AddAzureAppConfiguration

abhilasharora:
Currently, calling `AddAzureAppConfiguration` with `optional:true` does not work. It still throws any exception thrown for failures in the initial load. This pull request contains the code change to fix that bug and adds a unit test to validate it.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -177,9 +177,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 // Block current thread for the initial load of key-values registered for refresh that are not already loaded
                 await Task.Run(() => LoadKeyValuesRegisteredForRefresh(data).ConfigureAwait(false).GetAwaiter().GetResult()).ConfigureAwait(false);
             }
-            catch (Exception exception) when (exception.InnerException is RequestFailedException ||
-                                              exception.InnerException is HttpRequestException ||
-                                              exception.InnerException is OperationCanceledException)
+            catch (Exception exception) when (exception is KeyVaultReferenceException ||
+                                              exception is RequestFailedException ||
+                                              ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -177,9 +177,10 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
                 // Block current thread for the initial load of key-values registered for refresh that are not already loaded
                 await Task.Run(() => LoadKeyValuesRegisteredForRefresh(data).ConfigureAwait(false).GetAwaiter().GetResult()).ConfigureAwait(false);
             }
-            catch (Exception exception) when (exception.InnerException is RequestFailedException ||
-                                              exception.InnerException is HttpRequestException ||
-                                              exception.InnerException is OperationCanceledException)
+            catch (Exception exception) when (exception is KeyVaultReferenceException ||
+                                              exception is RequestFailedException ||
+                                              ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
+                                              exception is OperationCanceledException)
```
zhenlan: Do we know why `OperationCanceledException` is not part of [`TryRefreshAsync`](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/master/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L110)?

abhilasharora: For `TryRefreshAsync`, we plan to add `CancellationToken` as a parameter to the method, so we may need to include it in the catch block for that method too.

We currently use `CancellationToken.None` as the cancellation token for calls to the AppConfig SDK, so catching this exception is not necessary at this time. In future, we may plan to replace it with a timeout cancellation token for all calls to the SDK so we limit the maximum time per call to allow additional retry attempts.

abhilasharora: I have added the handling of `OperationCancelledException` in `TryRefreshAsync` so we don't miss it once we start adding a non-default cancellation token.

PR #141, opened 2020-03-17
Fix bug that causes feature flag refresh even when there is no change

abhilasharora:
When a refresh operation is invoked for feature flags, the etags for the feature flags are fetched first. In case there is a change in the etag for any of the key-values, a call to fetch all the feature flags along with values is made to the configuration store. The current logic has a bug that causes the call to fetch all feature flags along with values to be made to the configuration store even when the etag hasn't changed for any of the key-values. This bug was introduced starting version `2.1.0-preview-010380001-1099` with the change to migrate to the `Azure.Data.AppConfiguration` SDK.

This pull request fixes the bug and would significantly reduce the volume of calls made to App Configuration when refresh operation is invoked for feature flags.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
@@ -119,7 +119,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
                 {
                     await foreach(ConfigurationSetting setting in client.GetConfigurationSettingsAsync(selector).ConfigureAwait(false))
                     {
-                        if (!eTagMap.TryGetValue(setting.Key, out string etag) || !etag.Equals(setting.ETag))
+                        if (!eTagMap.TryGetValue(setting.Key, out string etag) || !setting.ETag.Equals(etag))
                         {
                             hasKeyValueCollectionChanged = true;
                             break;
@@ -152,7 +152,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
                     {
                         await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAsync(selector).ConfigureAwait(false))
                         {
-                            if (!eTagMap.TryGetValue(setting.Key, out string etag) || !etag.Equals(setting.ETag))
+                            if (!eTagMap.TryGetValue(setting.Key, out string etag) || !setting.ETag.Equals(etag))
                             {
                                 changes.Add(new KeyValueChange
                                 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
```diff
@@ -119,7 +119,7 @@ public static async Task<IEnumerable<KeyValueChange>> GetKeyValueChangeCollectio
                 {
                     await foreach(ConfigurationSetting setting in client.GetConfigurationSettingsAsync(selector).ConfigureAwait(false))
                     {
-                        if (!eTagMap.TryGetValue(setting.Key, out string etag) || !etag.Equals(setting.ETag))
+                        if (!eTagMap.TryGetValue(setting.Key, out string etag) || !setting.ETag.Equals(etag))
```
jimmyca15: >setting.ETag.Equals(etag) [](start = 83, length = 25)

It fixes the issue, but it looks like the issue stems from the fact that we are converting the `ETag` to a string in eTagMap. Should we keep it as an ETag instead and use it as intended for etag comparison?

tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
```diff
@@ -284,7 +284,7 @@ public void ServerRequestIsMadeWhenDefaultCredentialIsSet()
             mockClient.Setup(c => c.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()))
                 .Returns(new MockAsyncPageable(new List<ConfigurationSetting> { _kv }));
 
-            KeyVaultReferenceException ex = Assert.Throws<KeyVaultReferenceException>(() =>
+            Assert.Throws<AuthenticationFailedException>(() =>
```
jimmyca15: >AuthenticationFailedException [](start = 26, length = 29)

This looks unrelated to the PR, but I'm assuming this test randomly started failing. The only way I can think of that happening was that it was missed in a previous PR, or we depend on a "latest-patch-version" of key vault client which now throws a different PR. 

Shouldn't we catch this new `AuthenticationFailedException` and convert it to `KeyVaultReferenceException` as we were doing before?

abhilasharora: I don't think its due to those reasons. I believe we had managed identity working for that specific key-vault in the past, so it would succeed with authentication but fail to find the secret and throw `KeyVaultReferenceException`. Now the managed identity endpoint doesn't seem to work, not sure why, so the test fails with `Azure.Identity.AuthenticationFailedException`.

Unfortunately, we can't handle this exception in the configuration provider since this belongs to the `Azure.Identity` package and we don't want a dependency on that package.

abhilasharora: As discussed offline, I have excluded this change from current PR and will send a PR for it separately along with other related changes.

PR #144, opened 2020-03-19
Update TryRefreshAsync to return false for exceptions from Azure.Identity

abhilasharora:
The package `Azure.Identity` throws `Azure.Identity.AuthenticationFailedException` when it fails to use a `TokenCredential`. For example, when `new DefaultAzureCredentials()` is used to connect to Azure App Configuration but no valid identity could be found, then it throws `AuthenticationFailedException`. This pull request updates `TryRefreshAsync` to handle all exceptions from `Azure.Identity` package without adding an explicit dependency, so the method returns false as the result.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -113,7 +113,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 e is RequestFailedException ||
                 ((e as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
                 e is OperationCanceledException ||
-                e.Source.Equals(AzureIdentityExceptionSource, StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -111,7 +112,8 @@ public async Task<bool> TryRefreshAsync()
                 e is KeyVaultReferenceException ||
                 e is RequestFailedException ||
                 ((e as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
-                e is OperationCanceledException)
+                e is OperationCanceledException ||
+                e.Source.Equals(AzureIdentityExceptionSource, StringComparison.OrdinalIgnoreCase))
```
jimmyca15: Source can be null right? looks like it could throw a null-ref.

abhilasharora: If the Source property is null, the .NET runtime automatically sets it to the assembly name in which the exception originated. So I don't expect it to be null for any top-level exceptions. It could be null for inner exceptions. I would still add the null check here just to be safe.

PR #146, opened 2020-03-23
Wrap exceptions from Azure.Identity with KeyVaultReferenceException

abhilasharora:
Check related PR https://github.com/Azure/AppConfiguration-DotnetProvider/pull/144 for context


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationRefresher.cs
```diff
@@ -21,6 +21,7 @@ public interface IConfigurationRefresher
         /// The refresh operation failed with one or more errors. Check <see cref="AggregateException.InnerExceptions"/> for more details.
         /// </exception>
         /// <exception cref="InvalidOperationException">The refresh operation was invoked before Azure App Configuration Provider was initialized.</exception>
+        /// <exception>An Azure.Identity.AuthenticationFailedException is thrown if authentication using TokenCredential passed to <see cref="AzureAppConfigurationKeyVaultOptions.SetCredential"/> method failed.</exception>
```
jimmyca15: We can't document this we don't own it.

zhenlan: Not sure if there is a better way to do this. I understand we don't reference the package that has this exception type. With this, customers can handle it if they want at least. The `AuthenticationFailedException` should really be from either Key Vault client or Azure.Core. We should open an issue against the SDK.

abhilasharora: @jimmyca15 We document this exception since there are valid scenarios that will cause `RefreshAsync` method to throw it and we expect the users to handle it. I know this is not ideal, since we don't own it. It still seems to be better than crashing user's application even when a user handles all the documented exceptions.

abhilasharora: @zhenlan I already have issue https://github.com/Azure/azure-sdk-for-net/issues/10759 open against the SDK team. It seems they are considering moving the exception to `Azure.Core` similar to what we were discussing offline. However, I don't expect that change to be available before our next release.

jimmyca15: @abhilasharora documenting the exception like this is similar to taking a dependency on it. If we feel the need to do that perhaps we should just catch it like we are doing right now in TryRefreshAsync and wrap it in a KeyVaultReference Exception.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/AssemblyNames.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Constants
```
jimmyca15: The name spacing strategy here seems a little backwards. I usually see namespaces being used to determine the domain for which the types in the namespace belong. If we were going to come up with a new namespace for code dealing with Azure Identity I would expect that the namespace is something like `Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureIdentity` then it would have a class named `Constants`. Then the using statement might be a bit more clear for users who pull in the namespace since it shows they have some dependency on AzureIdentity.

Another option is to not have a new namespace at all, or to just put the string as a private in the class like it was before. It's just the `constants` namespace that throws me off.

abhilasharora: I don't think we should have a separate sub-namespace for `AzureIdentity`. The class `AssemblyNames` that I added is just a regular class for constants similar to the one we have for `ErrorMessages`. It has no relation to AzureIdentity. The constant that I created is just one of the assembly names. If we need to add more assembly names, we can add it to this class. It doesn't make the class specific to AzureIdentity.

We only need this value at a few places, so I can move this as a constant if you'd like. We already have the `Constants` namespace for error messages and request tracing constants, so that will be left unchanged.

jimmyca15: I'm fine with not having a separate sub-namespace. Just as long as we don't have the constants namespace. It doesn't need a namespace. It would be like having a namespace `Something.Something.Exceptions`

abhilasharora: Sure, that makes sense. I will go ahead and remove the `Constants` sub-namespace.

PR #147, opened 2020-03-23
Update user agent to use product version instead of assembly version

abhilasharora:
This pull request updates the user agent string to include the package version (informational version) instead of the assembly version. Check out [this blog](https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/) for more details on the different versions in .NET.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
@@ -6,7 +6,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -23,14 +22,13 @@ internal static class TracingUtils
         public static string GenerateUserAgent(string currentUserAgent = null)
         {
             Assembly assembly = typeof(AzureAppConfigurationOptions).Assembly;
-            var userAgent = new StringBuilder($"{assembly.GetName().Name}/{assembly.GetName().Version}");
+            string informationalVersion = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
```
zhenlan: Just to understand, who will set the `AssemblyInformationalVersionAttribute`?

tests/Tests.AzureAppConfiguration/Tests.cs
```diff
@@ -208,6 +208,10 @@ public void TestUserAgentHeader()
             MockRequest request = mockTransport.SingleRequest;
 
             string appUserAgent = TracingUtils.GenerateUserAgent("SdkUserAgent");
+
+            // Validate the user agent has information version (3.9.9999) instead of assembly version (3.9.9999.0)
+            Assert.Equal("Microsoft.Extensions.Configuration.AzureAppConfiguration/3.9.9999 SdkUserAgent", appUserAgent);
```
zhenlan: I am wondering if this test will pass in an official build.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
abhilasharora: This attribute is automatically generated by the `.NET Core SDK` based on the `Version` property in the `csproj` file. When I build the project, I can see the `*AssemblyInfo.cs` file in the obj directory with the `AssemblyInformationalVersionAttribute` attribute along with other auto-generated attributes, and these are also applied to the corresponding assembly. More information can be found [here](https://compiledexperience.com/blog/posts/assemblyinfo-generation).

zhenlan: Thanks.

PR #151, opened 2020-03-27
Update RefreshAsync to not ignore exceptions when optional parameter is set in AddAzureAppConfiguration

abhilasharora:
This pull request contains changes to update `RefreshAsync` to not ignore exceptions when optional parameter is set in `AddAzureAppConfiguration`.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             try
             {
-                LoadAll().ConfigureAwait(false).GetAwaiter().GetResult();
+                LoadAll(_optional).ConfigureAwait(false).GetAwaiter().GetResult();
             }
             catch (ArgumentException)
             {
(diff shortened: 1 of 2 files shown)


jimmyca15:
This change seems good to me. The optional parameter is  used to control the behavior of startup rather than the behavior of individual refresh calls. However, this change affects those who are currently using optional. Their code will now be able to throw when they call RefreshAsync whereas it would never throw before. This is a noticable behavioral change. Will this be taken into account for release versioning?

abhilasharora:
> This change seems good to me. The optional parameter is used to control the behavior of startup rather than the behavior of individual refresh calls. However, this change affects those who are currently using optional. Their code will now be able to throw when they call RefreshAsync whereas it would never throw before. This is a noticable behavioral change. Will this be taken into account for release versioning?

@jimmyca15 I don't quite agree with `Their code will now be able to throw when they call RefreshAsync whereas it would never throw before`. Even when the `optional` parameter is set, the current behavior is that `RefreshAsync` throws the documented exceptions. This change fixes the inconsistent behavior where `RefreshAsync` would hide exceptions that were thrown in a **subset of the cases**, specifically when `refreshAll` was set and the corresponding `LoadAll` had failed, which is a bug in the current code.

With this bug fix, the public interface and the documented behavior remains unchanged, so I consider it a non-breaking change.

jimmyca15:
@abhilasharora good point. This is only for refreshAll, and since the error could have happened in the initial check for etag change then it was possible for the exception to occur before. No breaking change then.

PR #152, opened 2020-03-31
Update RefreshAsync to auto-recover from initial load failure

abhilasharora:
When the `optional: true` parameter is passed to `AddAzureAppConfigurationProvider` and the initial configuration load fails, a valid instance of `IConfigurationRefresher` is returned and the exceptions are suppressed. In this case, a user could invoke `RefreshAsync` or `TryRefreshAsync` method to refresh the configuration. The current implementation has a bug that throws `NullReferenceException` when these methods are invoked.

This pull request contains changes to update the implementation of `RefreshAsync` to fix the bug and attempt auto-recover by reloading the initial configuration. In order to accomplish that, it uses the minimum cache expiration from the `CacheExpirationTime` value of for each key-value or feature flag, and fetches the configuration settings from scratch in the same manner as it is done when a `refreshAll` is triggered.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -31,6 +31,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private AzureAppConfigurationOptions _options;
         private ConcurrentDictionary<string, ConfigurationSetting> _settings;
 
+        private static TimeSpan? MinCacheExpirationTime;
+        private static DateTimeOffset LastRefreshAttemptTime = default;
+        private static readonly SemaphoreSlim RefreshAttemptSemaphore = new SemaphoreSlim(1);
+
         private static readonly TimeSpan MinDelayForUnhandledFailure = TimeSpan.FromSeconds(5);
 
         public AzureAppConfigurationProvider(ConfigurationClient client, AzureAppConfigurationOptions options, bool optional)
@@ -95,6 +99,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         public async Task RefreshAsync()
         {
+            // Check if initial configuration load had failed
+            if (_settings == null)
+            {
+                await LoadInitialConfiguration().ConfigureAwait(false);
+                return;
+            }
+
             await RefreshIndividualKeyValues().ConfigureAwait(false);
             await RefreshKeyValueCollections().ConfigureAwait(false);
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -31,6 +31,10 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
         private AzureAppConfigurationOptions _options;
         private ConcurrentDictionary<string, ConfigurationSetting> _settings;
 
+        private static TimeSpan? MinCacheExpirationTime;
```
jimmyca15, line 34: >TimeSpan [](start = 23, length = 8)

Do we need to cache this thing at the class level? If so then it could probably be set in the constructor.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -31,6 +31,10 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
         private AzureAppConfigurationOptions _options;
         private ConcurrentDictionary<string, ConfigurationSetting> _settings;
 
+        private static TimeSpan? MinCacheExpirationTime;
+        private static DateTimeOffset LastRefreshAttemptTime = default;
```
jimmyca15, line 35: >static [](start = 16, length = 6)

Doesn't look like a static property to me.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,6 +219,41 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
...
+            }
+
+            var timeElapsedSinceLastLoadAttempt = DateTimeOffset.UtcNow - LastRefreshAttemptTime;
+
+            if (timeElapsedSinceLastLoadAttempt < MinCacheExpirationTime || !RefreshAttemptSemaphore.Wait(0))
+            {
+                return;
+            }
+
+            LastRefreshAttemptTime = DateTimeOffset.UtcNow;
```
jimmyca15, line 245: >LastRefreshAttemptTime [](start = 12, length = 22)

looks like it's loadAttempt in the local variable. Load attempt makes more sense to me.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,6 +219,41 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
...
+                    // No registration was found for refresh
+                    return;
+                }
+
+                MinCacheExpirationTime = watchers.Min(w => w.CacheExpirationTime);
+            }
+
+            var timeElapsedSinceLastLoadAttempt = DateTimeOffset.UtcNow - LastRefreshAttemptTime;
+
+            if (timeElapsedSinceLastLoadAttempt < MinCacheExpirationTime || !RefreshAttemptSemaphore.Wait(0))
```
jimmyca15, line 240: >RefreshAttemptSemaphore [](start = 77, length = 23)

Also I would expect the term load here rather than refresh.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,6 +219,41 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
...
+        {
+            // Check if this is the first attempt to load the configuration for failed startup load as a part of refresh
+            if (MinCacheExpirationTime == null)
+            {
+                IEnumerable<KeyValueWatcher> watchers = _options.ChangeWatchers.Union(_options.MultiKeyWatchers);
+
+                if (!watchers.Any())
+                {
+                    // No registration was found for refresh
+                    return;
```
jimmyca15, line 232: >return [](start = 20, length = 6)

If the code returns here then we never mark any state to prevent it from happening every time refresh is called. Intended?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,6 +219,41 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
             }
         }
 
+        private async Task LoadInitialConfiguration()
```
jimmyca15, line 222: >LoadInitialConfiguration [](start = 27, length = 24)

The method doesn't always LoadInitialConfiguration as it says. Seems like the caller needs to be smarter about when to invoke this or the maybe the name needs attended to. Maybe splitting it into `IsLoadAllowed()` and `LoadInitial` ?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Let's put a comment here to mark that this is the shared refresh attempt tracker.

jimmyca15, line 245: Please check the condition for acquiring the semaphore again after having acquired it.

PR #153, opened 2020-04-01
Update package version for provider to 3.0.1

abhilasharora:
This change updates the version for the package `Microsoft.Extensions.Configuration.AzureAppConfiguration` to 3.0.1.


jimmyca15:
What updates are included ?

abhilasharora:
> What updates are included ?

The updates include fixing the optional parameter, feature flag refresh using etag, handling exceptions from identity in TryRefreshAsync, package version in user agent, optional parameter handling in RefreshAsync, and auto-recovery for initial load failure in refresh. I will update the release notes with more details.

jimmyca15:
Got it. Thanks for the summary. Just couldn't remember what the last update was since previous release.

PR #161, opened 2020-04-17
Update refresh to use SetCacheExpiration time to recover from initial load failures

abhilasharora:
The `RefreshAsync` and `TryRefreshAsync` methods were updated to be able to detect a failure and re-attempt to load the initial settings from App Configuration. It is possible to leverage this feature even when there are no settings registered for refresh. In this case, the current behavior is to use a default cache expiration time of 30 seconds for auto-recovery of the initial configuration load.

In such cases, if a custom value of the cache expiration time has been specified by the user through the `SetCacheExpiration` method, we would like to use it as the cache expiration time for recovery of initial configuration load, instead of the default cache expiration time. For example, a cache expiration time of `10 seconds` would be used for the following code snippet, instead of the default cache expiration time of `30 seconds`.

```` csharp
IConfiguration configuration = builder.Build();

{
    options.Connect(configuration["connection_string"])
            .ConfigureRefresh(refresh =>
            {
                refresh.SetCacheExpiration(TimeSpan.FromSeconds(10));
            });

    _refresher = options.GetRefresher();
});
````

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -89,6 +89,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         internal ConfigurationClientOptions ClientOptions { get; } = GetDefaultClientOptions();
 
         /// <summary>
+        /// Cache expiration time for refresh when no settings were registered for refresh and initial configuration failed to load.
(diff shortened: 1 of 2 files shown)


jimmyca15:
If we do this can we remove this code: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/d19492dd64f379240194eba945468e52f9ec37ff/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L52 and just use this? I'd like to avoid having the min cache logic in multiple places.

PR #165, opened 2020-06-03
Add ResetCache to IConfigurationRefresher for Push Refresh

abhilasharora:
This pull request adds `ResetCache` method to the interface `IConfigurationRefresher` to allow users to invalidate the cache for the key-values. This will ensure that the subsequence call to `RefreshAsync` or `TryRefreshAsync` would not use the cached value and instead retrieve the updated configuration from the server. When a user received a push notification from the event grid indicating that a key-value has been updated or deleted, this method will reset the cache to allow the configuration to be updated.

This change also renames `TimeSpan CacheExpirationTime` to `TimeSpan CacheExpirationInterval` and instead uses `DateTimeOffset CacheExpirationTime` to indicate the time when the cached value for a specific key would expire.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -141,7 +141,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             return true;
         }
 
-        public void ResetCache()
+        public void SetDirty()
         {
             DateTimeOffset currentTime = DateTimeOffset.UtcNow;
 
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -141,6 +141,19 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             return true;
         }
 
+        public void ResetCache()
+        {
+            foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
+            {
+                changeWatcher.LastRefreshTime = DateTimeOffset.MinValue;
+            }
+
+            foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
+            {
+                changeWatcher.LastRefreshTime = DateTimeOffset.MinValue;
+            }
+        }
+
         private async Task LoadAll(bool ignoreFailures)
         {
             IDictionary<string, ConfigurationSetting> data = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -31,11 +31,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private AzureAppConfigurationOptions _options;
         private ConcurrentDictionary<string, ConfigurationSetting> _settings;
 
-        private readonly TimeSpan MinCacheExpirationTime;
+        private readonly TimeSpan MinCacheExpirationInterval;
         private readonly SemaphoreSlim InitializationSemaphore = new SemaphoreSlim(1);
 
         // The most-recent time when the refresh operation attempted to load the initial configuration
-        private DateTimeOffset LastRefreshAttemptTime = default;
+        private DateTimeOffset InitializationCacheExpirationTime = default;
 
         private static readonly TimeSpan MinDelayForUnhandledFailure = TimeSpan.FromSeconds(5);
 
@@ -49,11 +49,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             if (watchers.Any())
             {
-                MinCacheExpirationTime = watchers.Min(w => w.CacheExpirationTime);
+                MinCacheExpirationInterval = watchers.Min(w => w.CacheExpirationInterval);
             }
             else
             {
-                MinCacheExpirationTime = AzureAppConfigurationRefreshOptions.DefaultCacheExpirationTime;
+                MinCacheExpirationInterval = AzureAppConfigurationRefreshOptions.DefaultCacheExpirationInterval;
             }
 
             // Enable request tracing if not opt-out
@@ -143,14 +143,16 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         public void ResetCache()
         {
+            DateTimeOffset currentTime = DateTimeOffset.UtcNow;
+
             foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
             {
-                changeWatcher.LastRefreshTime = DateTimeOffset.MinValue;
+                changeWatcher.CacheExpirationTime = currentTime;
             }
 
             foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
             {
-                changeWatcher.LastRefreshTime = DateTimeOffset.MinValue;
+                changeWatcher.CacheExpirationTime = currentTime;
             }
         }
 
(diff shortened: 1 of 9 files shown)


tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -771,6 +771,37 @@ public async Task RefreshTests_UpdatesAllSettingsIfInitialLoadFails()
...
+                .Build();
+
+            Assert.Equal("TestValue1", config["TestKey1"]);
+            FirstKeyValue.Value = "newValue";
+
+            // Invalidate the cache to force next refresh
+            refresher.ResetCache();
+
+            refresher.RefreshAsync().Wait();
+            Assert.Equal("newValue", config["TestKey1"]);
```
zhenlan, line 802: Can we add these 2 lines before the `ResetCache()` too to validate values are not changed?

abhilasharora: Added

abhilasharora:
> It's because I responded on GitHub. It's in regards to this one
> 
> > This is unrelated to this PR, but I just thought of it. Should we initialize CacheExpirationTime for all of the watchers here? otherwise they will be initialized to default which will mean they're ready to refresh even though they were just loaded.
> 
> In reply to: [642304788](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/165#issuecomment-642304788) [](ancestors = 642304788)

To be able to set cache expiration time for key-values that were loaded on startup and registered for refresh, we can find intersection of the `KeyValueWatcher` list and the loaded settings by comparing (Key, Label) pairs and then set `CacheExpirationTime` for all of them as current timestamp. I don't think this is necessary, but I can add that code if you'd like to ensure that settings loaded on startup cannot be updated before the cache expiration interval is reached.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -141,6 +141,21 @@ public async Task<bool> TryRefreshAsync()
             return true;
         }
 
+        public void SetDirty()
+        {
+            DateTimeOffset currentTime = DateTimeOffset.UtcNow;
+
+            foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
+            {
+                changeWatcher.CacheExpirationTime = currentTime;
```
drago-draganov, line 150: Can we name it CacheExpires instead of CacheExpirationTime?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -141,6 +141,21 @@ public async Task<bool> TryRefreshAsync()
             return true;
         }
 
+        public void SetDirty()
+        {
+            DateTimeOffset currentTime = DateTimeOffset.UtcNow;
```
drago-draganov, line 146: now

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -381,10 +393,8 @@ private async Task RefreshKeyValueCollections()
         {
             foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
             {
-                var timeElapsedSinceLastRefresh = DateTimeOffset.UtcNow - changeWatcher.LastRefreshTime;
-
                 // Skip the refresh for this key-prefix if the cached value has not expired or a refresh operation is in progress
-                if (timeElapsedSinceLastRefresh < changeWatcher.CacheExpirationTime || !changeWatcher.Semaphore.Wait(0))
+                if (DateTimeOffset.UtcNow < changeWatcher.CacheExpirationTime || !changeWatcher.Semaphore.Wait(0))
```
drago-draganov, line 397: We should check against same point in time.

jimmyca15:
@abhilasharora I thought all we would have to do is add a for loop right here: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/d19492dd64f379240194eba945468e52f9ec37ff/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L201 to ensure all the change watchers were updated.

abhilasharora:
At line 201, it is not guaranteed that settings for all key watchers have been fetched, so we can't reset cache expiration time there.

If we move line 280 before continue, then we are setting the cache expiration time even when we haven't fetched the value which may fail. We don't want to update this time in case of failures.

---
In reply to: [642312541](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/165#issuecomment-642312541) [](ancestors = 642312541)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
abhilasharora: From `CacheExpires` it isn't clear what this field is meant to store. `CacheExpirationTime` makes it explicit that it stores the time when the cache expires.

---
In reply to: [438453096](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/165#discussion_r438453096) [](ancestors = 438453096)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
abhilasharora: If the current time equals cache expiration time, we do want to refresh, so this condition is correct.

---
In reply to: [438453710](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/165#discussion_r438453710) [](ancestors = 438453710)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
drago-draganov: LastModified, Created, Expires, etc. is the common naming pattern across platforms and frameworks (see also RFC)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
drago-draganov: That's my point. Take a snapshot of Now and evaluate based on it, instead of moving Now in each iteration.

abhilasharora: If user invokes RefreshAsync at 12:00:00, first registered key-label has cache expiration time 11:59:59 and second key-label has cache expiration time 12:00:02, then taking a snapshot at 12:00:00 would not refresh the second key-label even when we take 60 seconds to refresh the first key-value. Checking the current time enables us to fetch the updated value in case it is stale by the time we reach it. The goal is to refresh as often as possible without violating the minimum cache expiration interval.

drago-draganov: Make sense. Wanted to make sure it's by design.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
drago-draganov: We also have SetCacheExpiration, not SetCacheExpirationTime.

abhilasharora: `CacheExpirationTime` is a field that we populate based on current time and cache interval. It isn't set by `SetCacheExpiration` method. `Expires` as a field name seems fine, but would be more appropriate in an object that represents the cache. `CacheExpires` seems odd and I couldn't find any component that uses such a name.

drago-draganov: Can you find a component that uses CacheExpirationTime?

zhenlan: `CacheExpires` -> `CacheExpiration`

The `SetCacheExpiration` API sets the value of `CacheExpiration` property :)

abhilasharora: The `SetCacheExpiration` sets the cache expiration interval (`TimeSpan`). The `CacheExpirationTime` here is different (`DateTimeOffset`).

drago-draganov: The current SetCacheExpiration is effectively SetCacheExpirationInterval

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
abhilasharora: Updated

zhenlan:
@abhilasharora will you add jitter in a separate PR?

abhilasharora:
> @abhilasharora will you add jitter in a separate PR?

Added in PR https://github.com/Azure/AppConfiguration-DotnetProvider/pull/176

PR #177, opened 2020-06-23
Allow users to configure max delay for SetDirty

abhilasharora:
This pull request contains the changes to allow users to override the default maximum randomized delay for calls to the `SetDirty` method in `IConfigurationRefresher`.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationRefresher.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationRefresher.cs
@@ -31,7 +31,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         /// <summary>
         /// Sets the cached value for key-values registered for refresh as dirty.
         /// A random delay is added before the cached value is marked as dirty to reduce potential throttling in case multiple instances refresh at the same time.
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationRefresher.cs
```diff
@@ -30,8 +30,10 @@ public interface IConfigurationRefresher
 
         /// <summary>
         /// Sets the cached value for key-values registered for refresh as dirty.
-        /// The next call to <see cref="RefreshAsync"/> or <see cref="TryRefreshAsync"/> would cause cached values to be revalidated.
+        /// A random delay is added before the cached value is marked as dirty to reduce potential throttling in case multiple instances refresh at the same time.
+        /// After the cache is marked as dirty, the subsequent call to <see cref="RefreshAsync"/> or <see cref="TryRefreshAsync"/> would cause cached values to be revalidated.
```
jimmyca15, line 34: I didn't really understand the last line. I thought the first two lines gave a good summary though. Do we need the last line?

abhilasharora: I will remove the last line. The only additional information it adds is that the refresh method needs to be invoked for the revalidation to happen, and that could be considered to be implicit from the method name and the interface context.

Issue #178, opened 2020-06-29
Inconsistent cached configuration after failed refreshAll operations


abhilasharora:
When a change is detected in a sentinel key-value with refreshAll set to true, a refreshAll operation is triggered to populate the entire configuration from scratch. However, in case we fail to complete that operation and all retry attempts fail, the subsequent calls to TryRefreshAsync fail to detect changes and do not attempt to retry the failed operation. This issue tracks the changes required to ensure that failed refreshAll operations do not leave the internal configuration cache in an inconsistent state.

PR #179, opened 2020-06-29
Fix issue with refresh on refreshAll failure

abhilasharora:
When a change is detected in a sentinel key-value with `refreshAll` set to `true`, a `refreshAll` operation is triggered to populate the entire configuration from scratch. However, in case we fail to complete that operation and all retry attempts fail, the subsequent calls to `TryRefreshAsync` fail to detect changes and do not attempt to retry the failed operation. This pull request updates code to ensure that failed `refreshAll` operations do not leave the internal configuration cache in an inconsistent state. https://github.com/Azure/AppConfiguration-DotnetProvider/issues/178

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -334,8 +334,14 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                         // Check if a change has been detected in the key-value registered for refresh
                         if (keyValueChange.ChangeType != KeyValueChangeType.None)
                         {
-                            ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
+                            if (changeWatcher.RefreshAll)
+                            {
+                                shouldRefreshAll = true;
+                                break;
+                            }
+
                             hasChanged = true;
+                            ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
                         }
                     }
                     else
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -356,25 +362,22 @@ private async Task RefreshIndividualKeyValues()
...
-                            shouldRefreshAll = true;
-
-                            // Skip refresh for other key-values since refreshAll will populate configuration from scratch
-                            break;
-                        }
-                        else
-                        {
-                            await SetData(_settings).ConfigureAwait(false);
-                        }
+                        await SetData(_settings).ConfigureAwait(false);
```
jimmyca15, line 380: >await SetData [](start = 24, length = 13)

Not relevant to this PR, but it looks like calling this in a loop could cause us to do a lot of key value processor calls. That could mean a lot of key vault calls. Maybe we could optimize to build a list of KeyValuePair and then update settings all at once instead of multiple calls.

Just not sure how common it is to watch a lot of single key-values.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -334,8 +334,14 @@ private async Task RefreshIndividualKeyValues()
...
                         {
-                            ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
+                            if (changeWatcher.RefreshAll)
+                            {
+                                shouldRefreshAll = true;
+                                break;
+                            }
+
                             hasChanged = true;
+                            ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
```
drago-draganov, line 344: when do we call ProcessChanges in case of changeWatcher.RefreshAll?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
abhilasharora: Yes, we can move it outside the loop to call it once instead of once per watcher and reduce the calls to adapters. One downside I can think of that approach is that if we exhaust both retry attempts to load a specific watcher, we lose the settings fetched previously. In order to preserve those settings, we might need some refactoring for the exception logic.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
abhilasharora: `ProcessChanges` is used to incrementally update or delete a setting in the cached configuration. In case of `refreshAll`, we load all configuration from scratch, so we don't need to call `ProcessChanges`.

drago-draganov: What about if the watch key isn't part of the select filters?

jimmyca15: `LoadAll()` is called on line 392. `LoadAll()` ensures that all the keys registered for refresh, like sentinel keys, are obtained.

drago-draganov: Thanks!

PR #181, opened 2020-07-10
Adding JSON Content type support for key-values

avanigupta:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -179,7 +179,14 @@ private async Task LoadAll(bool ignoreFailures)
                     {
                         await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsAsync(selector, CancellationToken.None).ConfigureAwait(false))
                         {
-                            data[setting.Key] = setting;
+                            if (JsonConfigurationParser.IsJsonContentType(setting.ContentType))
+                            {
```
abhilasharora, line 183: This is only one of the multiple places where settings are retrieved from the configuration store. For this change, we can add an adapter that implements `IKeyValueAdapter`, which is designed to run for each key-value, irrespective of where it is retrieved or updated. For examples, you can look at the classes `AzureKeyVaultKeyValueAdapter` and `FeatureManagementKeyValueAdapter`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,108 @@
...
+                    || Equals(type, KeyVaultConstants.ContentType))
+                {
+                    return false;
+                }
+
+                if (type.Contains('/'))
+                {
+                    string mainType = type.Split('/')[0];
+                    string subType = type.Split('/')[1];
+                    if (Equals(mainType, acceptedMainType))
```
abhilasharora, line 33: >Equals [](start = 24, length = 6)

This would ignore JSON parsing if the content type is `APPLICATION/JSON` (upper case).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,108 @@
...
+                            }
+                        }
+                        else if (Equals(subType, acceptedSubType))
+                        {
+                            return true;
+                        }
+                    }
+                }
+            }
+            catch (Exception ex) when (ex is ArgumentException || ex is FormatException || ex is NullReferenceException)
```
abhilasharora, line 50: >NullReferenceException [](start = 97, length = 22)

Instead of catching `NullReferenceException`, we can use null conditional checks.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,108 @@
...
+                    {
+                        if (subType.Contains('+'))
+                        {
+                            var subTypes = subType.Split('+');
+                            if (Array.Exists(subTypes, x => x == acceptedSubType))
+                            {
+                                return true;
+                            }
+                        }
+                        else if (Equals(subType, acceptedSubType))
```
abhilasharora, line 43: >Equals(subType, acceptedSubType) [](start = 33, length = 32)

Case-insensitive comparison : subType.Equals(acceptedSubType, StringComparison.OrdinalIgnoreCase)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,108 @@
...
+                if (type.Contains('/'))
+                {
+                    string mainType = type.Split('/')[0];
+                    string subType = type.Split('/')[1];
+                    if (Equals(mainType, acceptedMainType))
+                    {
+                        if (subType.Contains('+'))
+                        {
+                            var subTypes = subType.Split('+');
+                            if (Array.Exists(subTypes, x => x == acceptedSubType))
```
abhilasharora, line 38: >acceptedSubType [](start = 65, length = 15)

Case-insensitive comparison

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,108 @@
...
+                        }
+                        break;
+
+                    default:
+                        keyValueDict[currentKey] = json_data.ToString();
+                        break;
+                }
+
+            }
+            catch (Exception ex) when (ex is ArgumentNullException || ex is JsonException || ex is NotSupportedException)
```
abhilasharora, line 101: >ArgumentNullException [](start = 45, length = 21)

Where do expect this exception to be thrown?

abhilasharora, line 101: >NotSupportedException [](start = 99, length = 21)

Where do we expect this exception to be thrown?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,108 @@
...
+            catch (Exception ex) when (ex is ArgumentException || ex is FormatException || ex is NullReferenceException)
+            {
+                // not a valid JSON content type
+            }
+            return false;
+        }
+
+        public static void ParseJsonSetting(ConfigurationSetting setting, IDictionary<string, ConfigurationSetting> data)
+        {
+            SortedDictionary<string, string> keyValueDict = new SortedDictionary<string, string>();
```
abhilasharora, line 59: >SortedDictionary [](start = 12, length = 16)

Do these settings need to be sorted in a specific order?

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
```diff
@@ -12,6 +12,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.1.0" />
     <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.0.0" />
```
abhilasharora, line 15: >Microsoft.Extensions.Configuration.Json [](start = 31, length = 39)

Where is this package being used?

abhilasharora, line 15: >2.0.0 [](start = 81, length = 5)

We could use the updated patched version `2.0.2`

abhilasharora:
> using Azure.Data.AppConfiguration;

License header?

---
Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs:1 in c30fe2f. [](commit_id = c30fe2f240199e46ba0c049456725a8400ea1d67, deletion_comment = False)

tests/Tests.AzureAppConfiguration/TestHelper.cs
```diff
@@ -61,6 +62,22 @@ public static ConfigurationSetting CloneSetting(ConfigurationSetting setting)
         {
             return ConfigurationModelFactory.ConfigurationSetting(setting.Key, setting.Value, setting.Label, setting.ContentType, setting.ETag, setting.LastModified);
         }
+
+        public static List<ConfigurationSetting> LoadJsonSettingsFromfile(string path)
```
abhilasharora, line 66: >LoadJsonSettingsFromfile [](start = 49, length = 24)

File? (casing)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
abhilasharora: Just noticed you convert it to lowercase before using it.

---
In reply to: [453139448](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453139448) [](ancestors = 453139448)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,108 @@
...
+    internal static class JsonConfigurationParser
+    {
+        public static bool IsJsonContentType(string contentType)
+        {
+            try
+            {
+                string acceptedMainType = "application";
+                string acceptedSubType = "json";
+
+                ContentType ct = new ContentType(contentType.Trim().ToLower());
```
abhilasharora, line 21: >contentType [](start = 49, length = 11)

We can add null check for `contentType` before dereferencing it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,108 @@
...
+                var type = ct.MediaType;
+                if (Equals(type, FeatureManagementConstants.ContentType)
```
abhilasharora, line 23: >Equals [](start = 20, length = 6)

How about using `==` in place of `Equals`?

tests/Tests.AzureAppConfiguration/MockTestData/appconfig-settings.json
```diff
@@ -0,0 +1,132 @@
...
+      [ "NestedArray3" ]
+    ],
+    "contentType": "application/json"
+  },
+  {
+    "key": "key17",
+    "value": "2020-01-01T12:34:56.789Z",
+    "contentType": "application/json"
+  }
+]
```
abhilasharora, line 132: Can we add test cases for key-vault reference and feature flag content types? That will help validate those values remain unaffected, even though they are valid json types.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Thanks for pointing this out! I was looking at KeyVault and FeatureManagement adapters. It looks like they are added to AzureAppConfigurationOptions only when users explicitly configure them with AddAzureAppConfiguration method. In the case of JSON values, we should always add this adapter by default right?

---
In reply to: [453138843](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453138843) [](ancestors = 453138843)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
avanigupta: From the docs, JsonSerializer.Deserialize can throw these three exceptions: [https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize?view=netcore-3.1](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize?view=netcore-3.1) 

I can reproduce ArgumentNullException when the a setting has null value and json content type. For example, this setting was created through CLI:
Az appconfig kv set --key foobar --content-type application/json -y
{
  "contentType": "application/json",
  "etag": "RAHNsqogncZZdEIkBgyTM1in7FT",
  "key": "foobar",
  "label": null,
  "lastModified": "2020-07-13T19:50:20+00:00",
  "locked": false,
  "tags": {},
  "value": null
}

Notice that the value is null. Since the value is invalid JSON, going forward, we will not allow users to create new settings with null value and JSON content type, but there could be some customers with existing settings like this.

I was not able to reproduce NotSupportedException but still kept it to be on the safer side.


---
In reply to: [453141145](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453141145) [](ancestors = 453141145)

tests/Tests.AzureAppConfiguration/MockTestData/appconfig-settings.json
avanigupta: I added a test for feature flags (JsonContentTypeTests_DontFlattenFeatureFlagAsJsonObject) when users don't configure "UseFeatureFlags". In that case, we should be able to read the value of feature flags as regular string (no object flattening).

I did not add tests for when users do configure Feature Flags and KeyVault refs because we already have explicit tests for these. As long as those tests pass, we don't need to add duplicate tests here. I think it's best to keep them decoupled in case we make changes to Feature Flags or KeyVault functionality in future.


---
In reply to: [453143374](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453143374) [](ancestors = 453143374)

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
avanigupta: It's used to load config settings from JSON file ( new ConfigurationBuilder().AddJsonFile(jsonFilePath) )

---
In reply to: [453141525](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453141525) [](ancestors = 453141525)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
avanigupta: We don't really need to sort the keys. If the value is a nested JSON object, we could just store keys in the order that they appear inside the value. 
The only thing is, if we don't sort nested objects here, the "data" dictionary will not have sorted keys. When we load settings from config store, we get settings sorted by key. I am not sure if there would be a side effect of having unsorted dictionary of config settings.
Moreover, sorting here will also give the same config output as we would get from reading appsettings.json file (which is not a deciding factor though).

---
In reply to: [453141260](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453141260) [](ancestors = 453141260)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
avanigupta: Is == better than Equals in this case? I found that there is no difference when comparing strings and == actually calls Equals function: [https://docs.microsoft.com/en-us/dotnet/api/system.string.op_equality?view=netcore-3.1#remarks](https://docs.microsoft.com/en-us/dotnet/api/system.string.op_equality?view=netcore-3.1#remarks) 

---
In reply to: [453142883](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453142883) [](ancestors = 453142883)

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
avanigupta: It looks like the latest one is 3.1.5. Should we just update it to the latest version?
[https://www.nuget.org/packages/Microsoft.Extensions.Configuration.Json](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.Json) 

---
In reply to: [453142063](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453142063) [](ancestors = 453142063)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
abhilasharora: `AzureKeyVaultKeyValueAdapter` is added by default, and `FeatureManagementKeyValueAdapter` is added only when user uses `UseFeatureFlags` in their code. For JSON-related adapter, we can enable it by default. It will provide an abstraction over the code to detect if it is JSON and actual processing, so the logic can be at once place, instead of all places where we update settings.

---
In reply to: [453876591](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453876591) [](ancestors = 453876591,453138843)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonKeyValueAdapter.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, CancellationToken cancellationToken)
+        {
+            var keyValues = JsonConfigurationParser.ParseJsonSetting(setting);
+            return Task.FromResult<IEnumerable<KeyValuePair<string, string>>>(keyValues);
+        }
+
+        public bool CanProcess(ConfigurationSetting setting)
+        {
+            return JsonConfigurationParser.IsJsonContentType(setting?.ContentType);
```
abhilasharora, line 21: >IsJsonContentType [](start = 43, length = 17)

We can skip calling this method if `setting` is `null`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Net.Mime;
+using System.Text.Json;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.JsonConfiguration
+{
+    internal static class JsonConfigurationParser
+    {
+        public static bool IsJsonContentType(string contentType)
+        {
+            try
```
abhilasharora, line 19: >try [](start = 12, length = 3)

We can reduce the `try` block scope to only include code that can throw the exceptions in the `catch` block.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                if (string.IsNullOrEmpty(contentType))
```
abhilasharora, line 21: >IsNullOrEmpty [](start = 27, length = 13)

How about `IsNullOrWhitespace` instead of `IsNullOrEmpty`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
abhilasharora: We can update `CanProcess` to return `false` if the value is `null`. Then we wouldn't need to handle it in the parser.

---
In reply to: [453895459](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453895459) [](ancestors = 453895459,453141145)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
abhilasharora: I would expect the settings to preserve the original ordering from the source instead of being sorted.

---
In reply to: [453933488](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453933488) [](ancestors = 453933488,453141260)

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
abhilasharora: We generally avoid using any package version that has updated patched versions available. If this was product code, we would prefer using the older version with the latest patch to include any fixes or security updates. For test code, you can choose `2.0.2` or most-recent patch for any newer version.

---
In reply to: [453945711](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453945711) [](ancestors = 453945711,453142063)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
abhilasharora: Both of these are value comparisons and can be used interchangeably. We currently use `==` for string comparisons in this solution, so using it would help maintain consistency.

---
In reply to: [453940673](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453940673) [](ancestors = 453940673,453142883)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            string acceptedMainType = "application";
+            string acceptedSubType = "json";
+
+            try
+            {
+                ContentType ct = new ContentType(contentType.Trim().ToLower());
+                var type = ct.MediaType;
+                if (type == FeatureManagementConstants.ContentType ||
+                    type == KeyVaultConstants.ContentType)
```
abhilasharora, line 31: We can extract them in a list as follows, so it is easier to find this list and update it in future if needed.
private static IEnumerable<string> _excludedJsonContentTypes = new [] { FeatureManagementConstants.ContentType, KeyVaultConstants.ContentType };

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                if (type.Contains('/'))
+                {
+                    string mainType = type.Split('/')[0];
+                    string subType = type.Split('/')[1];
+                    if (mainType == acceptedMainType)
+                    {
+                        if (subType.Contains('+'))
+                        {
+                            var subTypes = subType.Split('+');
+                            if (Array.Exists(subTypes, x => x == acceptedSubType))
```
abhilasharora, line 45: >if (Array.Exists(subTypes, x => x == acceptedSubType)) [](start = 28, length = 54)

How about `if (subTypes.Contains(acceptedSubType))` instead of using a predicate?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                            {
+                                return true;
+                            }
+                        }
+                        else if (subType == acceptedSubType)
```
abhilasharora, line 50: >else if (subType == acceptedSubType) [](start = 24, length = 36)

Wouldn't this be covered by lines 44-48 if the `if` check is removed?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                        {
+                            return true;
+                        }
+                    }
+                }
+            }
+            catch (Exception ex) when (ex is ArgumentException || ex is FormatException)
```
abhilasharora, line 57: >ArgumentException [](start = 45, length = 17)

We don't need to catch `ArgumentException` since content type cannot be empty.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                return false;
+            }
+            string acceptedMainType = "application";
+            string acceptedSubType = "json";
+
+            try
+            {
+                ContentType ct = new ContentType(contentType.Trim().ToLower());
+                var type = ct.MediaType;
```
abhilasharora, line 29: Since this is the only statement that could throw exception(s) in the catch block, we can get move the other statements outside the try block.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                            var newKey = ConfigurationPath.Combine(new List<string> { currentKey, index.ToString() });
+                            ParseSetting(newKey, valueArray.ElementAt(index).GetRawText(), keyValueDict);
+                        }
+                        break;
+
+                    case JsonValueKind.Object:
+                        var valueObject = json_data.EnumerateObject();
+                        foreach(JsonProperty entry in valueObject)
+                        {
+                            var newKey = ConfigurationPath.Combine(new List<string> { currentKey, entry.Name.ToString() });
```
abhilasharora, line 97: >entry.Name [](start = 98, length = 10)

Could `entry.Name` be `null`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                    default:
+                        keyValueDict[currentKey] = json_data.ToString();
+                        break;
+                }
+
+            }
+            catch (Exception ex) when (ex is JsonException || ex is NotSupportedException)
+            {
+                // If it's not a valid JSON, treat it like regular string value
+                keyValueDict[currentKey] = currentValue;
```
abhilasharora, line 111: >keyValueDict [](start = 16, length = 12)

Is there a scenario where we update the value for the same key in this dictionary?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                        }
+                        break;
+
+                    default:
+                        keyValueDict[currentKey] = json_data.ToString();
+                        break;
+                }
+
+            }
+            catch (Exception ex) when (ex is JsonException || ex is NotSupportedException)
```
abhilasharora, line 108: >NotSupportedException [](start = 68, length = 21)

This might hide any deserialization bugs and cause silent failures. We might want to let it bubble up.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        public static void ParseSetting(string currentKey, string currentValue, Dictionary<string, string> keyValueDict)
+        {
+            try
+            {
+                var json_data = JsonSerializer.Deserialize<JsonElement>(currentValue);
+                switch (json_data.ValueKind)
+                {
+                    case JsonValueKind.Array:
+                        var valueArray = json_data.EnumerateArray();
```
abhilasharora, line 85: >var [](start = 24, length = 3)

We can specify explicit type `ArrayEnumerator` to increase readability. Same for `ObjectEnumerator` type in the next case block.

abhilasharora, line 85: >EnumerateArray [](start = 51, length = 14)

Return type for this method implements `IDisposable`, so it needs to be manually disposed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        public static List<KeyValuePair<string, string>> ParseJsonSetting(ConfigurationSetting setting)
+        {
+            List<KeyValuePair<string, string>> keyValues = new List<KeyValuePair<string, string>>();
+            Dictionary<string, string> keyValueDict = new Dictionary<string, string>();
+            ParseSetting(setting.Key, setting.Value, keyValueDict);
+            foreach (KeyValuePair<string, string> entry in keyValueDict)
+            {
+                keyValues.Add(entry);
+            }
```
abhilasharora, line 72: You could replace this loop with `keyValueDict.ToList()`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+            return keyValues;
+        }
+
+
+        public static void ParseSetting(string currentKey, string currentValue, Dictionary<string, string> keyValueDict)
```
abhilasharora, line 77: >public [](start = 8, length = 6)

This method could be `private` instead of `public`.

tests/Tests.AzureAppConfiguration/JsonContentTypeTests.cs
```diff
@@ -0,0 +1,215 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Xunit;
+
+namespace Tests.AzureAppConfiguration
+{
+    public class JsonContentTypeTests
+    {
+        List<ConfigurationSetting> _kvCollection = new List<ConfigurationSetting>();
```
abhilasharora, line 21: >List<ConfigurationSetting> _kvCollection = new List<ConfigurationSetting>(); [](start = 8, length = 76)

Since I see most test cases are clearing this collection in the beginning and populating it, how about creating it inside each test instead?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
jimmyca15, line 29: >var [](start = 16, length = 3)

What type is that? `var` adds a bit of obscurity if the type isn't readily available in the assignment.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                var type = ct.MediaType;
+                if (type == FeatureManagementConstants.ContentType ||
+                    type == KeyVaultConstants.ContentType)
+                {
+                    return false;
+                }
+
+                if (type.Contains('/'))
+                {
+                    string mainType = type.Split('/')[0];
```
jimmyca15, line 38: >type.Split [](start = 38, length = 10)

We can split once, and better yet we can avoid a split.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                    string subType = type.Split('/')[1];
```
jimmyca15, line 39: >subType [](start = 27, length = 7)

Not needed unless the if statement is true.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
jimmyca15, line 45: >Array.Exists(subTypes, x => x == acceptedSubType) [](start = 32, length = 49)

A common way to write this pattern is:

`subtypes.Any(s => s == acceptedSubType);`

jimmyca15: Your way here works fine if you like. That other way could save code in the future though.

---
In reply to: [454684293](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454684293) [](ancestors = 454684293)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
jimmyca15, line 57: >Exception [](start = 19, length = 9)

Rather than catching exception can we catch the specific exceptions.

jimmyca15, line 57: >ArgumentException [](start = 45, length = 17)

what can throw this?

jimmyca15, line 57: >catch [](start = 12, length = 5)

Can we avoid calling anything that can throw? I'd like to avoid creation of exceptions in this code path if possible.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            return keyValues;
+        }
+
+
+        public static void ParseSetting(string currentKey, string currentValue, Dictionary<string, string> keyValueDict)
+        {
+            try
+            {
+                var json_data = JsonSerializer.Deserialize<JsonElement>(currentValue);
```
jimmyca15, line 81: >json_data [](start = 20, length = 9)

This doesn't look like c# variable naming.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
jimmyca15, line 85: >var [](start = 24, length = 3)

Type please.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                        break;
+
+                    default:
+                        keyValueDict[currentKey] = json_data.ToString();
+                        break;
+                }
+
+            }
+            catch (Exception ex) when (ex is JsonException || ex is NotSupportedException)
+            {
```
jimmyca15, line 109: The best bet with a try catch is to only put it around the piece that can throw it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
jimmyca15, line 97: >ConfigurationPath.Combine(new List<string> { currentKey, entry.Name.ToString() }); [](start = 41, length = 82)

You can avoid a list allocation.

`$"{currentKey}{ConfigurationPath.KeyDelimiter}{entry.Name.ToString()}"`

jimmyca15, line 97: >Name.ToString() [](start = 104, length = 15)

Isn't `Name` a string?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                            ParseSetting(newKey, valueArray.ElementAt(index).GetRawText(), keyValueDict);
+                        }
+                        break;
+
+                    case JsonValueKind.Object:
+                        var valueObject = json_data.EnumerateObject();
+                        foreach(JsonProperty entry in valueObject)
+                        {
+                            var newKey = ConfigurationPath.Combine(new List<string> { currentKey, entry.Name.ToString() });
+                            ParseSetting(newKey, entry.Value.GetRawText(), keyValueDict);
```
jimmyca15, line 98: >entry.Value.GetRawText() [](start = 49, length = 24)

If this is another object we're going to be deserializing -> serializing -> deserializing. Can't we avoid that?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
jimmyca15, line 77: >ParseSetting [](start = 27, length = 12)

This looks very similar to [https://github.com/Azure/AppConfiguration-DotnetProvider/blob/master/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/JsonFlattener.cs](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/master/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/JsonFlattener.cs)

jimmyca15: Maybe we can just re-use that.

---
In reply to: [454694319](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454694319) [](ancestors = 454694319)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+            catch (Exception ex) when (ex is ArgumentException || ex is FormatException)
+            {
+                // not a valid JSON content type
+            }
+            return false;
+        }
+
+        public static List<KeyValuePair<string, string>> ParseJsonSetting(ConfigurationSetting setting)
```
jimmyca15, line 64: >List [](start = 22, length = 4)

IEnumerable.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
jimmyca15, line 77: >public [](start = 8, length = 6)

Looks like this can be private.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            List<KeyValuePair<string, string>> keyValues = new List<KeyValuePair<string, string>>();
```
jimmyca15, line 66: >keyValues [](start = 47, length = 9)

Can't we just return `keyValueDict`? It should implement `IEnumerable<KeyValuePair<string, string>>`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,115 @@
...
+                    type == KeyVaultConstants.ContentType)
+                {
+                    return false;
+                }
+
+                if (type.Contains('/'))
+                {
+                    string mainType = type.Split('/')[0];
+                    string subType = type.Split('/')[1];
+                    if (mainType == acceptedMainType)
```
jimmyca15, line 40: >mainType == acceptedMainType [](start = 24, length = 28)

Content types are case insensitive.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
jimmyca15, line 64: >ParseJsonSetting [](start = 57, length = 16)

`Parse` should be good enough for the name here. The type name ties everything here to json and the parameter shows that it's a ConfigurationSetting.

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
avanigupta: Changed it to 2.0.2

---
In reply to: [454039139](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454039139) [](ancestors = 454039139,453945711,453142063)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
avanigupta: Done

---
In reply to: [454041425](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454041425) [](ancestors = 454041425,453940673,453142883)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs

---
In reply to: [454034139](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454034139) [](ancestors = 454034139,453876591,453138843)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfigurationParser.cs
avanigupta: Done

---
In reply to: [454037685](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454037685) [](ancestors = 454037685,453895459,453141145)

tests/Tests.AzureAppConfiguration/MockTestData/appconfig-settings.json
avanigupta: Added more tests.

---
In reply to: [453921593](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r453921593) [](ancestors = 453921593,453143374)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonKeyValueAdapter.cs
avanigupta: Done

---
In reply to: [454035078](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454035078) [](ancestors = 454035078)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Done

---
In reply to: [454615335](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454615335) [](ancestors = 454615335)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Done

---
In reply to: [454616697](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454616697) [](ancestors = 454616697)

avanigupta: As suggested by Abhilash, changed it to "(subTypes.Contains(acceptedSubType))" which also works fine.

---
In reply to: [454684520](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454684520) [](ancestors = 454684520,454684293)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Yes, removed

---
In reply to: [454617283](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454617283) [](ancestors = 454617283)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Done

---
In reply to: [454618216](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454618216) [](ancestors = 454618216)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Done

---
In reply to: [454619359](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454619359) [](ancestors = 454619359)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Removed the ToString() call because entry.Name  is already string.

---
In reply to: [454622297](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454622297) [](ancestors = 454622297)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Done

---
In reply to: [454626098](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454626098) [](ancestors = 454626098)

tests/Tests.AzureAppConfiguration/JsonContentTypeTests.cs
avanigupta: Done

---
In reply to: [454634233](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454634233) [](ancestors = 454634233)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Exception can be thrown by System.Net.Mime.ContentType class while parsing the content type. Since this is a standard way to parse content type, I didn't want to duplicate the code that does content type validation.

---
In reply to: [454687415](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454687415) [](ancestors = 454687415)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Fixed

---
In reply to: [454689036](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454689036) [](ancestors = 454689036)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Good idea! Thanks!

---
In reply to: [454692298](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454692298) [](ancestors = 454692298)

avanigupta: Yes, fixed.

---
In reply to: [454692807](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454692807) [](ancestors = 454692807)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Thanks for pointing it out! We can reuse the same flattener :)

---
In reply to: [454694903](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454694903) [](ancestors = 454694903,454694319)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Done

---
In reply to: [454695520](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454695520) [](ancestors = 454695520)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: I am converting content type to lowercase while parsing with System.Net.Mime.ContentType 

---
In reply to: [454696570](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454696570) [](ancestors = 454696570)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
avanigupta: Done

---
In reply to: [454698062](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r454698062) [](ancestors = 454698062)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -25,7 +26,7 @@ public class AzureAppConfigurationOptions
 
         private Dictionary<string, KeyValueWatcher> _changeWatchers = new Dictionary<string, KeyValueWatcher>();
         private List<KeyValueWatcher> _multiKeyWatchers = new List<KeyValueWatcher>();
-        private List<IKeyValueAdapter> _adapters = new List<IKeyValueAdapter>() { new AzureKeyVaultKeyValueAdapter(new AzureKeyVaultSecretProvider()) };
+        private List<IKeyValueAdapter> _adapters = new List<IKeyValueAdapter>() { new AzureKeyVaultKeyValueAdapter(new AzureKeyVaultSecretProvider()), new JsonKeyValueAdapter() };
```
jimmyca15, line 29: This line is a bit long, off screen. Can we do it like

```
adapters = new List {
  a,
  b
};
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Mime;
+using System.Text.Json;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.JsonConfiguration
+{
+    internal static class JsonConfigurationParser
+    {
+        private static IEnumerable<string> _excludedJsonContentTypes = new[] { FeatureManagementConstants.ContentType, KeyVaultConstants.ContentType };
```
jimmyca15, line 17: >FeatureManagementConstants.ContentType, KeyVaultConstants.ContentType [](start = 79, length = 69)

Can we do it like
```
new[] {
  a,
  b
}
```

Makes it a bit easier for me to see the items, also more scalable.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,80 @@
...
+                    }
+                }
+            }
+
+            return false;
+        }
+
+        public static IEnumerable<KeyValuePair<string, string>> Parse(ConfigurationSetting setting)
+        {
+            string rootJson = $"{{\"{setting.Key}\":{setting.Value}}}";
```
jimmyca15, line 65: >: [](start = 51, length = 1)

Isn't `ConfigurationPath.Delimiter` available for that?

jimmyca15: Nvm, different purpose here.

---
In reply to: [456048962](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r456048962) [](ancestors = 456048962)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,80 @@
...
+            JsonElement jsonData;
+            try
+            {
+                jsonData = JsonSerializer.Deserialize<JsonElement>(rootJson);
+            }
+            catch (JsonException)
+            {
+                // If the value is not a valid JSON, treat it like regular string value
+                return new List<KeyValuePair<string, string>> { new KeyValuePair<string, string>(setting.Key, setting.Value) };
```
jimmyca15, line 74: >List [](start = 27, length = 4)

Array would be fine, no need for list interface here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonKeyValueAdapter.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.JsonConfiguration
+{
+    internal class JsonKeyValueAdapter : IKeyValueAdapter
+    {
+        public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, CancellationToken cancellationToken)
+        {
+            IEnumerable<KeyValuePair<string, string>> keyValues = JsonConfigurationParser.Parse(setting);
+            return Task.FromResult(keyValues);
```
jimmyca15, line 16: Can be simplified to 1 line.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
jimmyca15, line 65: Can we add parameter validation. Throwing argumentnullexception if setting is null.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -5,6 +5,7 @@
 using Azure.Data.AppConfiguration;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.JsonConfiguration;
```
jimmyca15, line 8: >JsonConfiguration [](start = 63, length = 17)

I'm not sure if this new namespace is valuable. Would it be fine if we just put the types in the `Microsoft.Extensions.Configuration.AzureAppConfiguration` namespace?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            string acceptedMainType = "application";
+            string acceptedSubType = "json";
+            string mediaType;
+
+            try
+            {
+                ContentType ct = new ContentType(contentType.Trim().ToLower());
+                mediaType = ct.MediaType;
```
jimmyca15, line 33: Could be one line.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+            catch (FormatException)
+            {
+                return false;
+            }
+            catch (IndexOutOfRangeException)
+            {
+                // Bug in System.Net.Mime.ContentType throws this if contentType is "xyz/"
```
jimmyca15, line 41: >Bug in System.Net.Mime.ContentType [](start = 19, length = 34)

Is there an issue filed for this?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonConfiguration/JsonConfigurationParser.cs
```diff
@@ -0,0 +1,80 @@
...
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Mime;
+using System.Text.Json;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.JsonConfiguration
+{
+    internal static class JsonConfigurationParser
```
jimmyca15, line 15: >JsonConfigurationParser [](start = 26, length = 23)

Would it make sense to move the functionality of this type into JsonKeyValueAdapter? Parse looks like it should be the implementation for `JsonKeyValueAdapter.ProcessKeyValue` and IsJsonContentType could probably be a private method.

jimmyca15: On second thought `IsJsonContentType ` could just be inlined into CanProcess.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: Done

---
In reply to: [456047463](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r456047463) [](ancestors = 456047463)

jimmyca15: Whoops my bracket placement was inconsistent in my suggestion. To be consistent with the rest of the code we would need like

```
  a = new List
  {
    b,
    c
  }
```

---
In reply to: [456090082](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/181#discussion_r456090082) [](ancestors = 456090082,456047463)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Net.Mime;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal class JsonKeyValueAdapter : IKeyValueAdapter
+    {
+        private static readonly IEnumerable<string> _excludedJsonContentTypes = new[] {
```
jimmyca15, line 19: >{ [](start = 86, length = 1)

same

abhilasharora, line 19: >_excludedJsonContentTypes [](start = 52, length = 25)

Since this is `static`, the naming convention would be to use `ExcludedJsonContentTypes` instead of `_excludedJsonContentTypes`.

PR #183, opened 2020-07-17
Revert "Adding JSON Content type support for key-values"

avanigupta:
Reverts Azure/AppConfiguration-DotnetProvider#181


avanigupta:
Created for a VSO merge test only

PR #187, opened 2020-07-21
Update logic for Nuget package version generation

abhilasharora:
This pull request adds a `Preview` property to allow generation of preview versions for Nuget packages generated in official builds.

### DIFF
--- a/build/NugetProperties.props
+++ b/build/NugetProperties.props
@@ -8,7 +8,6 @@
     <PackageProjectUrl>https://github.com/Azure/AppConfiguration</PackageProjectUrl>
     <PackageLicenseUrl>https://aka.ms/AzureAppConfigurationDotnetCorePublicPreviewEula</PackageLicenseUrl>
     <PackageIconUrl>https://aka.ms/AzureAppConfigurationPackageIcon</PackageIconUrl>
-    <Preview>true</Preview>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'==''">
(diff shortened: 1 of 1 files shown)


build/NugetProperties.props
```diff
@@ -18,7 +19,7 @@
     <Version>4.0.0</Version>
```
zhenlan, line 19: When we build official preview release, can we just hard code the version here, for example, `4.0.0-preview` here?

abhilasharora: This code auto-generates a random string to append to the preview version name, leading to versions such as `3.0.0-preview-010550001-251` and `3.0.0-preview-011100001-1152`. We could hard code the version for official builds and use a different naming convention like `4.0.0-preview1`, `4.0.0-preview2` instead.

jimmyca15: The random generated numbers were helpful to prevent collisions when we were doing a bunch of dev builds and pushing to our dev NuGet feed. Now that we have a more stable release process and we don't collide, I think it would be great to see the clear version like `4.0.0-preview`

jimmyca15: The random numbers could still be tacked onto the end for dev builds if we want.

abhilasharora: What if we want to release multiple preview versions? We can name them as `4.0.0-preview1`, `4.0.0-preview2` etc.

zhenlan: We can do `4.0.0-preview` first. If we find we need another preview release later, we can still do `4.0.0-preview2`. At this point, I think we only plan to do one preview release.

jimmyca15: I don't see anything wrong with starting with `preview1` but I don't see it as necessary either. I would personally just put preview and not assume that we will have a `preview2`, fine either way though.

zhenlan:
Approved, but this should be the last PR merged before we have the `4.0.0-preview` release.

Have all other changes merged already? Do we have a release candidate build tested?

abhilasharora:
> Approved, but this should be the last PR merged before we have the `4.0.0-preview` release.
> 
> Have all other changes merged already? Do we have a release candidate build tested?

The order of PRs does not matter for the build and release process. I will still wait to submit this change after other changes are merged, so the commit history is more logical.

Issue #173, opened 2020-06-18
Add multi-targeting support to Microsoft.Azure.AppConfiguration.AspNetCore for .NET Core 3.1


abhilasharora:
A large number of NuGet packages aren't produced for ASP.NET Core 3.1. The package `Microsoft.Azure.AppConfiguration.AspNetCore` has a dependency on the package `Microsoft.AspNetCore.Http.Abstractions` that is included in the [list](https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-3.1&tabs=visual-studio#remove-obsolete-package-references) of obsolete packages.

This issue tracks the changes needed to add multi-targeting support for ASP.NET Core 3.1.

abhilasharora:
These changes are now available in version `4.0.0-preview` of the package.

PR #189, opened 2020-07-22
Add multi-targeting support for .NET Core 3.1

abhilasharora:
This pull request updates the package `Microsoft.Azure.AppConfiguration.AspNetCore` to target .NET Core 3.1 in addition to .NET Standard 2.0 https://github.com/Azure/AppConfiguration-DotnetProvider/issues/173

### DIFF
--- a/examples/ConfigStoreDemo/ConfigStoreDemo.csproj
+++ b/examples/ConfigStoreDemo/ConfigStoreDemo.csproj
@@ -1,11 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
     <IsPackable>false</IsPackable>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


jimmyca15:
@abhilasharora so we're not adding the multi-targeting to the `Microsoft.Extensions.Configuration.AzureAppConfiguration` package since we would have the same dependencies for both right? Wanted to make sure it was considered since other `Microsoft.Extensions...` packages [do do multi-targeting](https://www.nuget.org/packages/Microsoft.Extensions.Configuration/3.1.0).

abhilasharora:
> @abhilasharora so we're not adding the multi-targeting to the `Microsoft.Extensions.Configuration.AzureAppConfiguration` package since we would have the same dependencies for both right? Wanted to make sure it was considered since other `Microsoft.Extensions...` packages [do do multi-targeting](https://www.nuget.org/packages/Microsoft.Extensions.Configuration/3.1.0).

We have two packages that have multiple targets - `Microsoft.Extensions.Configuration` (2 targets) and `System.Text.Json` (13 targets). I have evaluated them and we don't need multi-targeting for them.

PR #193, opened 2020-08-06
Fix issue with user agent header policy

abhilasharora:
The change in the user agent header policy resulted in [our code](https://msazure.visualstudio.com/Azure%20AppConfig/_git/Azconfig-DotnetProvider?path=%2Fsrc%2FMicrosoft.Extensions.Configuration.AzureAppConfiguration%2FTracingUtils.cs&version=GBmaster) in the configuration provider to add this substring to the user agent. With this change, the order of user agent substring generation in the configuration provider and the SDK was reversed.

https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/Pipeline/HttpPipelineBuilder.cs

image
image
image

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
@@ -6,10 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
-using System.Runtime.InteropServices;
-using System.Runtime.Versioning;
 using System.Security;
-using System.Text;
 using System.Threading.Tasks;
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -18,32 +15,11 @@ internal static class TracingUtils
     {
         static private HostType? _hostType = null;
 
-        public static string GenerateUserAgent(string currentUserAgent = null)
+        public static string GenerateUserAgent()
```
jimmyca15, line 18: Do we call this anywhere else or can we just move it into the UserAgentHeaderPolicy?

jimmyca15: I guess we use it in tests, but could always make it internal so it's visible to tests.

abhilasharora: We create an instance of the class `UserAgentHeaderPolicy` and send it to the SDK to include it in its HTTP pipeline. It currently only has methods that override the base class methods. From the design perspective, I'm not sure if we should include utility methods in the same class. Perhaps, that was the original reasoning this code was extracted out in the first place, but let me know if you think otherwise.

jimmyca15: Having `GenerateUserAgent` in a utility method doesn't sound right to me. I don't see a problem moving it into `UserAgentHeaderPolicy` even though right now it just overrides base class methods.

jimmyca15 (approved):

Issue #136, opened 2020-03-03
The optional:true parameter for AddAzureAppConfiguration does not work


zhenlan:
I have code below and a key vault reference in the store.
```cs
builder.AddAzureAppConfiguration(options =>
{
    options.Connect(configuration["ConnectionString"])
}, optional:true);
```

The app does not have access to my key vault, but since I specified `optional:true`, the app is not expected to throw. However, it does:

```
Unhandled exception. Microsoft.Extensions.Configuration.AzureAppConfiguration.KeyVaultReferenceException: No key vault credential configured and no matching secret client could be found.. ErrorCode:, Key:foo, Label:, Etag:oyI5kwp9Dp68pxoux30v7G6lVQJ, SecretIdentifier:https://keyvaultreference.vault.azure.net/secrets/MySecret
 ---> System.UnauthorizedAccessException: No key vault credential configured and no matching secret client could be found.
   at Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault.AzureKeyVaultSecretProvider.GetSecretValue(Uri secretUri, CancellationToken cancellationToken)
   at Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault.AzureKeyVaultKeyValueAdapter.ProcessKeyValue(ConfigurationSetting setting, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault.AzureKeyVaultKeyValueAdapter.ProcessKeyValue(ConfigurationSetting setting, CancellationToken cancellationToken)
   at Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureAppConfigurationProvider.ProcessAdapters(ConfigurationSetting setting, CancellationToken cancellationToken)
   at Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureAppConfigurationProvider.SetData(IDictionary`2 data, CancellationToken cancellationToken)
   at Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureAppConfigurationProvider.LoadAll()
   at Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureAppConfigurationProvider.Load()
   at Microsoft.Extensions.Configuration.ConfigurationRoot..ctor(IList`1 providers)
   at Microsoft.Extensions.Configuration.ConfigurationBuilder.Build()
[...]

PR #196, opened 2020-08-31
Don't throw KeyValueReferenceException when optional is set to true

abhilasharora:
This pull request fixes issue https://github.com/Azure/AppConfiguration-DotnetProvider/issues/136.


jimmyca15:
:shipit:

Issue #156, opened 2020-04-08
Allow refresh to be registered for multiple key-values with same key but different labels


abhilasharora:
This issue tracks the changes needed to allow users to register refresh for multiple key-values with the same key but different labels.

```` csharp
var configBuilder = new ConfigurationBuilder().AddAzureAppConfiguration(options =>
{
    options.Connect(connectionString)
        .Select(KeyFilter.Any, LabelFilter.Null)
        .Select(KeyFilter.Any, "some_label")
        .ConfigureRefresh(refresh =>
        {
            refresh.Register("some_key", LabelFilter.Null, true)
                   .Register("some_key", "some_label", true);
        })
});

**Comment by avanigupta** [comment]
## Proposed Solution:

Consider the following example to better understand the solution:

Let's say that your App Configuration has the following key-values:

Key | Label | Value
--- |--- |---
TestKey | Label1 | Value1
TestKey | Label2 | Value2

### Existing behavior:

Key-values registered for refresh overwrite key-values loaded as part of `Select` clause(s). This means that if some key-value is loaded by `Select` clause and the same key (with a different label) is registered for refresh, your application configuration will always contain the value corresponding to the refresh registered key-value. 

```cs
var config = new ConfigurationBuilder()
    .AddAzureAppConfiguration(options =>
    {
        options.Select(KeyFilter.Any, "Label1");
        options.Select(KeyFilter.Any, "Label2");
        options.ConfigureRefresh(refreshOptions =>
        {
            refreshOptions.Register("TestKey", "Label1")
                          .SetCacheExpiration(TimeSpan.FromSeconds(10));
        });

        refresher = options.GetRefresher();
    })
    .Build();
```

Your application configuration will have the following key-value: 

Key | Value
--- |---
TestKey |  Value1


### New functionality when `refreshAll` is `true`:

Maintaining the behavior that currently exists, refresh registered key-values overwrite key-values loaded by the `Select` clause(s). In addition, if there are multiple refresh registrations of same key with different labels, only the last key-value registered for refresh will be added to your application configuration. However, the refresh registered sentinel keys which are not added to application configuration can still be used to trigger a `refreshAll` operation, in case the sentinel key-value changes.

```cs
var config = new ConfigurationBuilder()
    .AddAzureAppConfiguration(options =>
    {
        options.Select(KeyFilter.Any, "Label1");
        options.Select(KeyFilter.Any, "Label2");
[...]

PR #199, opened 2020-09-09
Allow refresh for key-values with same key but different labels

abhilasharora:
This pull request contains changes to address issues https://github.com/Azure/AppConfiguration-DotnetProvider/issues/156 and https://github.com/Azure/AppConfiguration-DotnetProvider/issues/171.


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefreshOptions.cs
```diff
@@ -38,12 +38,17 @@ public AzureAppConfigurationRefreshOptions Register(string key, bool refreshAll)
         /// <param name="refreshAll">If true, a change in the value of this key refreshes all key-values being used by the configuration provider.</param>
         public AzureAppConfigurationRefreshOptions Register(string key, string label = LabelFilter.Null, bool refreshAll = false)
         {
-            RefreshRegistrations[key] = new KeyValueWatcher
+            if (string.IsNullOrEmpty(key))
```
jimmyca15, line 41: Guess we can put this in the release notes. Shouldn't effect anybody but technically a change.

jimmyca15:
Did you do any testing to see how the provider behaves with multiple keys with different labels? I looked at the code path and it seems to me it would cause refresh on every single call to RefreshAsync, but maybe I'm missing something.

PR #204, opened 2020-10-02
Set initial cache expiry time after all settings have been loaded

### LINKED ISSUE #172: Cache expiration is ignored on first call to RefreshAsync
After initial load of the configuration, if a setting is included in `Select` and also registered for refresh, the first call to `RefreshAsync` would ignore the cache expiration interval specified by the user and always revalidate and refresh the that setting. The expected behavior is to respect the cache expiration interval even for the first call to refresh.

avanigupta:
Fix #172

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -133,18 +133,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 throw;
             }
 
-            // Set the cache expiration time for all refresh registered settings
-            var initialLoadTime = DateTimeOffset.UtcNow;
-            foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
-            {
-                changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
-            }
-
-            foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
-            {
-                changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
-            }
-
             // Mark all settings have loaded at startup.
             _isInitialLoadComplete = true;
         }
@@ -198,6 +186,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private async Task LoadAll(bool ignoreFailures)
         {
             IDictionary<string, ConfigurationSetting> data = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);
+            bool isLoadAllComplete = false;
 
             try
             {
@@ -252,6 +241,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
                 // Block current thread for the initial load of key-values registered for refresh that are not already loaded
                 await Task.Run(() => LoadKeyValuesRegisteredForRefresh(data).ConfigureAwait(false).GetAwaiter().GetResult()).ConfigureAwait(false);
+
+                if (_options.OfflineCache != null)
+                {
+                    _options.OfflineCache.Export(_options, JsonSerializer.Serialize(data));
+                }
+
+                isLoadAllComplete = true;
             }
             catch (Exception exception) when (exception is RequestFailedException ||
                                               ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
@@ -259,28 +255,37 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             {
                 if (_options.OfflineCache != null)
                 {
-                    data = JsonSerializer.Deserialize<IDictionary<string, ConfigurationSetting>>(_options.OfflineCache.Import(_options));
-
-                    if (data != null)
+                    // During startup or refreshAll scenario, we'll try to populate config from offline cache, if available
+                    string cachedData = _options.OfflineCache.Import(_options);
+                    if (cachedData != null)
                     {
-                        await SetData(data, ignoreFailures).ConfigureAwait(false);
-                        return;
+                        data = JsonSerializer.Deserialize<IDictionary<string, ConfigurationSetting>>(cachedData);
+                        isLoadAllComplete = true;
                     }
                 }
 
-                if (!ignoreFailures)
+                // If we're unable to load data from offline cache, check if we need to ignore or rethrow the exception 
+                if (!isLoadAllComplete && !ignoreFailures)
                 {
                     throw;
                 }
-
-                return;
             }
 
-            await SetData(data, ignoreFailures).ConfigureAwait(false);
-
-            if (_options.OfflineCache != null)
+            if (isLoadAllComplete)
             {
-                _options.OfflineCache.Export(_options, JsonSerializer.Serialize(data));
+                await SetData(data, ignoreFailures).ConfigureAwait(false);
+
+                // Set the cache expiration time for all refresh registered settings
+                var initialLoadTime = DateTimeOffset.UtcNow;
+                foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
+                {
+                    changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
+                }
+
+                foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
+                {
+                    changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
+                }
             }
         }
 
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -186,6 +186,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private async Task LoadAll(bool ignoreFailures)
         {
             IDictionary<string, ConfigurationSetting> data = null;
+            string cachedData = null;
 
             try
             {
@@ -242,12 +243,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
                 // Block current thread for the initial load of key-values registered for refresh that are not already loaded
                 await Task.Run(() => LoadKeyValuesRegisteredForRefresh(serverData).ConfigureAwait(false).GetAwaiter().GetResult()).ConfigureAwait(false);
-
-                if (_options.OfflineCache != null)
-                {
-                    _options.OfflineCache.Export(_options, JsonSerializer.Serialize(serverData));
-                }
-
                 data = serverData;
             }
             catch (Exception exception) when (exception is RequestFailedException ||
@@ -257,7 +252,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 if (_options.OfflineCache != null)
                 {
                     // During startup or refreshAll scenario, we'll try to populate config from offline cache, if available
-                    string cachedData = _options.OfflineCache.Import(_options);
+                    cachedData = _options.OfflineCache.Import(_options);
+                    
                     if (cachedData != null)
                     {
                         data = JsonSerializer.Deserialize<IDictionary<string, ConfigurationSetting>>(cachedData);
@@ -277,6 +273,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
                 // Set the cache expiration time for all refresh registered settings
                 var initialLoadTime = DateTimeOffset.UtcNow;
+                
                 foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
                 {
                     changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
@@ -285,6 +282,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
                 {
                     changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
+                }
+
+                if (_options.OfflineCache != null && cachedData == null)
+                {
+                    _options.OfflineCache.Export(_options, JsonSerializer.Serialize(data));
                 }
             }
         }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -185,11 +185,12 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         private async Task LoadAll(bool ignoreFailures)
         {
-            IDictionary<string, ConfigurationSetting> data = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);
-            bool isLoadAllComplete = false;
+            IDictionary<string, ConfigurationSetting> data = null;
 
             try
             {
+                var serverData = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);
+
                 // Use default query if there are no key-values specified for use other than the feature flags
                 bool useDefaultQuery = !_options.KeyValueSelectors.Any(selector => !selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker));
 
@@ -206,7 +207,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     {
                         await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsAsync(selector, CancellationToken.None).ConfigureAwait(false))
                         {
-                            data[setting.Key] = setting;
+                            serverData[setting.Key] = setting;
                         }
                     }).ConfigureAwait(false);
                 }
@@ -234,20 +235,20 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     {
                         await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsAsync(selector, CancellationToken.None).ConfigureAwait(false))
                         {
-                            data[setting.Key] = setting;
+                            serverData[setting.Key] = setting;
                         }
                     }).ConfigureAwait(false);
                 }
 
                 // Block current thread for the initial load of key-values registered for refresh that are not already loaded
-                await Task.Run(() => LoadKeyValuesRegisteredForRefresh(data).ConfigureAwait(false).GetAwaiter().GetResult()).ConfigureAwait(false);
+                await Task.Run(() => LoadKeyValuesRegisteredForRefresh(serverData).ConfigureAwait(false).GetAwaiter().GetResult()).ConfigureAwait(false);
 
                 if (_options.OfflineCache != null)
                 {
-                    _options.OfflineCache.Export(_options, JsonSerializer.Serialize(data));
+                    _options.OfflineCache.Export(_options, JsonSerializer.Serialize(serverData));
                 }
 
-                isLoadAllComplete = true;
+                data = serverData;
             }
             catch (Exception exception) when (exception is RequestFailedException ||
                                               ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
@@ -260,18 +261,17 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     if (cachedData != null)
                     {
                         data = JsonSerializer.Deserialize<IDictionary<string, ConfigurationSetting>>(cachedData);
-                        isLoadAllComplete = true;
                     }
                 }
 
                 // If we're unable to load data from offline cache, check if we need to ignore or rethrow the exception 
-                if (!isLoadAllComplete && !ignoreFailures)
+                if (data == null && !ignoreFailures)
                 {
                     throw;
                 }
             }
 
-            if (isLoadAllComplete)
+            if (data != null)
             {
                 await SetData(data, ignoreFailures).ConfigureAwait(false);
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -133,6 +133,17 @@ public override void Load()
                 throw;
             }
 
+            // Set the cache expiration time for all refresh registered settings
+            foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
+            {
+                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
```
abhilasharora, line 139: >DateTimeOffset.UtcNow [](start = 45, length = 21)

Can we call `DateTimeOffset.UtcNow` once instead of one call per change watcher?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -229,7 +240,7 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
                         LabelFilter = loadOption.LabelFilter
                     };
 
-                    // Load all key-values with the null label.
+                    // Load all key-values with the specified label.
```
abhilasharora, line 243: >// Load all key-values with the specified label. [](start = 20, length = 48)

We could just remove this comment since it could be confusing. We aren't loading all key-values here, just the ones corresponding to one selector.

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -101,10 +121,43 @@ public void RefreshTests_RefreshIsSkippedIfCacheIsNotExpired()
                 .AddAzureAppConfiguration(options =>
                 {
                     options.Client = mockClient.Object;
-                    options.Select("TestKey*");
+                    // NOTE: This select clause is redundant, we are loading all settings by default in GetMockConfigurationClient()
+                    options.Select("TestKey*");
```
abhilasharora, line 125: We could just remove these two lines, since these are not relevant to the scenario we're testing.

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -920,6 +833,33 @@ void action() => new ConfigurationBuilder()
             Assert.Throws<ArgumentException>(action);
         }
 
+        [Fact]
+        public void RefreshTests_CacheExpiresIsNotDefaultValue()
```
jimmyca15, line 837: This seems like an implementation detail that is covered by the other test you added.

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -784,7 +697,7 @@ public void RefreshTests_SetDirtyForcesNextRefresh()
                     options.Select("TestKey*");
                     options.ConfigureRefresh(refreshOptions =>
                     {
-                        refreshOptions.Register("TestKey1")
+                        refreshOptions.Register("TestKey1", "label")
```
jimmyca15, line 700: Why are we changing this test, it seems unrelated.

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -101,10 +121,41 @@ public void RefreshTests_RefreshIsSkippedIfCacheIsNotExpired()
...
+            // Wait for some time but not enough to let the cache expire
+            Thread.Sleep(5000);
+
+            refresher.RefreshAsync().Wait();
+
+            Assert.Equal("TestValue1", config["TestKey1"]);
+        }
+
+        [Fact]
+        public void RefreshTests_RefreshIsSkippedIfKvNotInSelectAndCacheIsNotExpired()
```
jimmyca15, line 146: The test updates in this PR got really big. Can we shorten it to just this method and GetMockConfigurationClientSelectKeyLabel?

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -101,10 +121,41 @@ public void RefreshTests_RefreshIsSkippedIfCacheIsNotExpired()
...
+
+                    refresher = options.GetRefresher();
+                })
+                .Build();
+
+            Assert.Equal("TestValue1", config["TestKey1"]);
+            FirstKeyValue.Value = "newValue1";
+
+            // Wait for some time but not enough to let the cache expire
+            Thread.Sleep(5000);
```
jimmyca15, line 138: It won't let me put a comment in `RefreshTests_RefreshIsSkippedIfKvNotInSelectAndCacheIsNotExpired`, but we don't need a sleep there right? The issue it is trying to test is that we don't skip keys not in select.

avanigupta: RefreshTests_RefreshIsSkippedIfKvNotInSelectAndCacheIsNotExpired is testing the same functionality as RefreshTests_RefreshIsSkippedIfCacheIsNotExpired. 
The only difference is that in RefreshIsSkippedIfCacheIsNotExpired, the refresh registered key is already loaded (default behavior is to load all keys when `Select` clause is missing), whereas in RefreshIsSkippedIfKvNotInSelectAndCacheIsNotExpired, the refresh registered key is not already loaded by the `Select` clause.

---
In reply to: [500698506](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/204#discussion_r500698506) [](ancestors = 500698506)

jimmyca15: Any  test that uses arbitrary numbers for delays is suspect to me. I would say that both of these tests would be fine without the sleep at all.

avanigupta: Makes sense, I'll remove sleep from both tests. Thanks!

---
In reply to: [501235350](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/204#discussion_r501235350) [](ancestors = 501235350)

tests/Tests.AzureAppConfiguration/RefreshTests.cs
avanigupta: I found that some refresh scenario tests were actually not testing the intended functionality because of the following reasons:

- We were selecting/registering KVs based on just key name, whereas all the KVs in mock data also had a label. That led to unintentionally registering/selecting non-existent keys with null labels, and hence the test took a different code path than expected. 
For Eg: 

    - When `RefreshAsync `call triggers `RefreshIndividualKeyValues`, instead of calling `GetKeyValueChange `and processing the "changes", we load the setting again with `GetConfigurationSettingAsync`. That misses the intention of testing the `GetKeyValueChange `method.

    - When `LoadKeyValuesRegisteredForRefresh `is called, we should skip loading KV that have already loaded as part of `Select` clause. But since we registered a KV without its label, it's treated as a new KV and loaded again. This is actually why `RefreshTests_RefreshIsSkippedIfCacheIsNotExpired` was passing instead of catching the bug addressed in this PR.

- We were always returning the entire mock data collection of KV, regardless of what is present in the `Select` clause. This led to unintentional behavior in tests like `RefreshTests_RefreshRegisteredKeysAreLoadedOnStartup_DefaultUseQuery` and `RefreshTests_RefreshRegisteredKeysAreLoadedOnStartup_CustomUseQuery`. These tests would always pass regardless of `ConfigureRefresh `because the `Select` clause (or a lack thereof), ensures that all KVs from mock data collection are loaded.


To fix this, I have made the following changes to these tests:

1. I've explicitly added labels to every key that is registered for refresh. 
2. Because of # 1, I also had to register a new mock client setup for `GetConfigurationSettingAsync(ConfigurationSetting setting, bool cond, CancellationToken ct)` method.
3. Because of # 2, I was able to refactor all the tests that had the same mock client setup and replace the duplicated code with the shared `GetMockConfigurationClient()` method.
4. I've also added a new method `GetMockConfigurationClientSelectKeyLabel` for the tests that absolutely rely on loading only the specified KV with `Select` clause.

The only reason I included all these changes in this PR is because I wanted to fix the test that is related to this PR (`RefreshTests_RefreshIsSkippedIfCacheIsNotExpired`) by adding the correct label for refresh registered KV, and that small change triggered all the other changes that are mentioned above. 

But I understand that these are a lot of changes for one PR which are unrelated to the bug fix, so in this PR, I can add the new test `RefreshTests_RefreshIsSkippedIfKvNotInSelectAndCacheIsNotExpired` and `GetMockConfigurationClientSelectKeyLabel`, and create a new PR for all the other test changes.

jimmyca15: That sounds like a good idea. Thanks Avani.

jimmyca15 (approved):

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -180,6 +183,76 @@ public void WatchesFeatureFlags()
...
+                        }
+                        ",
+                label: default,
+                contentType: FeatureManagementConstants.ContentType + ";charset=utf-8",
+                eTag: new ETag("c3c231fd-39a0-4cb6-3237-4614474b92c1" + "f"));
+
+            featureFlags.Add(_kv2);
+
+            // Sleep for some time but not enough to let the cache expire
+            Thread.Sleep(1000);
```
jimmyca15, line 247: Looks like another arbitrary sleep that can be removed.

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -172,6 +172,9 @@ public void WatchesFeatureFlags()
                 eTag: new ETag("c3c231fd-39a0-4cb6-3237-4614474b92c1" + "f"));
 
             featureFlags.Add(_kv2);
+
+            // Sleep to let the cache expire
+            Thread.Sleep(1000);
```
jimmyca15, line 177: The sleep here looks like it's necessary since we are waiting for the cache to expire. But can we update the test to use the same timespan that we pass into `options.UseFeatureFlags(o => o.CacheExpirationInterval` so they don't get out of sync.

jimmyca15: After I saw that you added this sleep I was wondering to myself, how did this ever pass in the first place? Then I realize it was only passing because of a bug

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -316,8 +327,6 @@ private async Task LoadKeyValuesRegisteredForRefresh(IDictionary<string, Configu
                     watchedKv = null;
                 }
 
-                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
```
abhilasharora, line 319: >changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval); [](start = 16, length = 94)

I though a bit more about this line and it seems this would cause issues in the scenario when there is a change in the sentinel key with `refreshAll` set. Removing this line would cause the next refresh within 30 seconds to ignore the cached values that were just retrieved from the server.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -241,35 +241,51 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
...
                     throw;
                 }
-
-                return;
             }
 
-            await SetData(data, ignoreFailures).ConfigureAwait(false);
-
-            if (_options.OfflineCache != null)
+            if (isLoadAllComplete)
```
avanigupta, line 274: After thinking more about the refactoring, I think it makes sense to use a boolean flag to track if we were able to load the settings (from AppConfig or OfflineCache) in order to remove the return statements from the catch block. In my opinion, not using the flag would complicate this part of code and make it harder to understand next time.

jimmyca15: Did you consider initializing data to null and only setting it to non null in the places you are setting isLoadAllComplete to true? Then that would take the place of the bool.

avanigupta: Yes, but Abhilash pointed out (and I agree) that we would essentially be using `data` to do the work of a bool flag, and change the intended responsibility of `data`, which would make the code less readable. Having an explicit bool flag makes the functionality more obvious.

jimmyca15: I won't block on it, but I don't agree in the necessity of introducing a new variable to track the state of a method. Always that will make me think something is wrong with the design. I would say that the pattern below would only be more complicated with the introduction of a new variable to track. 

```

object some = null;

try
{
    some = Load();
}
catch (Exception)
{
    if (!ignoreFailures) throw;
}

if (some != null)
{
    Use(some);
}

```

avanigupta: With slight modification to your pattern, the two possible approaches can be simplified to this:

**1. Use `some` object to track its own loading status:**

```
object some = new object();

try
{
    some = Load();
}
catch (Exception)
{
    some = null;    // this looks sneaky, considering it was never null to begin with
    if (canLoadDifferently())
        some = LoadDifferently();
    if (!ignoreFailures && some == null) throw;
}

if (some != null) // this check unintentionally hides the fact that some can be null only if there is an exception
{
    Use(some);
}
```

**2. Use a bool flag to track `some` object's loading status:**

```
// this could introduce a bug in future if the flag is not in sync with the loading status
bool isLoadComplete = false;  
object some = new object();

try
{
    some = Load();
    isLoadComplete = true;
}
catch (Exception)
{
    if (canLoadDifferently())
    {
         some = LoadDifferently();
         isLoadComplete = true;
    }
    if (!ignoreFailures && !isLoadComplete) throw;
}

if (isLoadComplete)
{
    Use(some);
}
```

I can see the downsides of both approaches, and I am not strongly in favor of one over the other. However, I'd like to go with the second approach because it's slightly more readable.

jimmyca15: Your first example has a problem. The object `some` should be initialized to null. Yes, it would be odd if we reset it to null

avanigupta: But in our LoadAll() method, we never initialize `data` to null.
This is what we have:

`IDictionary<string, ConfigurationSetting> data = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);`

jimmyca15: We can just change it to initialize to null.

avanigupta: That would trigger changes in both `LoadAll` and `LoadKeyValuesRegisteredForRefresh`, and an additional overhead of null checks on `data` in various places.

jimmyca15: Not really, just need to have `serverData` inside of the `try` block. Just like you have `cacheData` in the `catch` block.

`var serverData = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);`
.
.
.
data = serverData;

jimmyca15: or move everything inside the try to get data into it's own method.
```
data = null;
try
{
    data = LoadFromServer();
}
```

avanigupta: In that case, we'll have something like this:

```
object some = null;

try
{
    object serverSome = new object();
    serverSome = Load();
    some = serverSome;
}
catch (Exception)
{
    some = null; 
    if (canLoadDifferently())
        some = LoadDifferently();
    if (!ignoreFailures && some == null) throw;
}

if (some != null)
{
    Use(some);
}

```

jimmyca15: That example seems to have some unnecessary pieces? The logic can be simplified to

```
object some = null;

try
{
    some = Load();
}
catch (Exception)
{
    if (canLoadDifferently())
        some = LoadDifferently();
    if (!ignoreFailures && some == null) throw;
}

if (some != null)
{
    Use(some);
}
```

avanigupta: Lets say we get an exception after loading a few key-values from AppConfig. At this point, data will not be null because it has some key-values that were loaded successfully.

Lets say offline cache is not configured and ignoreFailures is false.

That means, we will end up hitting `Use(some)`, and load partial config for the user without throwing any errors.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -229,47 +231,61 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
...
                 }
 
                 // Block current thread for the initial load of key-values registered for refresh that are not already loaded
-                await Task.Run(() => LoadKeyValuesRegisteredForRefresh(data).ConfigureAwait(false).GetAwaiter().GetResult()).ConfigureAwait(false);
+                await Task.Run(() => LoadKeyValuesRegisteredForRefresh(serverData).ConfigureAwait(false).GetAwaiter().GetResult()).ConfigureAwait(false);
+
+                if (_options.OfflineCache != null)
+                {
+                    _options.OfflineCache.Export(_options, JsonSerializer.Serialize(serverData));
+                }
```
abhilasharora, line 249: Could we keep the existing behavior to export only after the configuration load is complete? This will ensure that we don't delay loading of settings just to save the offline copy.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -229,47 +231,61 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
...
                                               exception is OperationCanceledException)
             {
                 if (_options.OfflineCache != null)
                 {
-                    data = JsonSerializer.Deserialize<IDictionary<string, ConfigurationSetting>>(_options.OfflineCache.Import(_options));
-
-                    if (data != null)
+                    // During startup or refreshAll scenario, we'll try to populate config from offline cache, if available
+                    string cachedData = _options.OfflineCache.Import(_options);
+                    if (cachedData != null)
```
abhilasharora, line 261: >if (cachedData != null) [](start = 20, length = 23)

I would suggest leaving a blank line before each block statement (such as if, foreach etc.) unless it is the first statement in the method. This will improve readability and maintain consistency with existing code.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Since we call `SetData` only once after the catch block, we don't know if this data was loaded from AppConfig or from OfflineCache. Moving the export here ensures that if there is an exception, and if we end up loading data from existing offline cache - we don't need to export same data back to offline cache.

abhilasharora: The existing code handles this case without exporting the offline cache data again in case of an exception. I wouldn't change that behavior since it affects the performance observed by the customer. One approach could be to only export if `cachedData` is `null`.

jimmyca15 (approved):
Looks good. One question. What's the purpose of the last commit to initialize cachedData to string.Empty?

avanigupta:
> Looks good. One question. What's the purpose of the last commit to initialize cachedData to string.Empty?

As Abhilash suggested in one of the previous comment, I moved the cache exporting logic back to the end of LoadAll. So before exporting, we just need to check if the data was loaded from server or from offline cache. If it was loaded from offline cache, we don't need to export the same data again. That's why I initialized cachedData at the beginning of LoadAll so that it can be accessed both inside and outside the catch block.

jimmyca15:
> > Looks good. One question. What's the purpose of the last commit to initialize cachedData to string.Empty?
> 
> As Abhilash suggested in this comment: [#204 (comment)](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/204#discussion_r502704628)
> 
> I moved the cache exporting logic back to the end of LoadAll. So before exporting, we just need to check if the data was loaded from server or from offline cache. If it was loaded from offline cache, we don't need to export the same data again. That's why I initialized cachedData at the beginning of LoadAll so that it can be accessed both inside and outside the catch block.

I get that. Not sure why it is initialized to string.Empty instead of null.

avanigupta:
> 
> 
> > > Looks good. One question. What's the purpose of the last commit to initialize cachedData to string.Empty?
> > 
> > 
> > As Abhilash suggested in this comment: [#204 (comment)](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/204#discussion_r502704628)
> > I moved the cache exporting logic back to the end of LoadAll. So before exporting, we just need to check if the data was loaded from server or from offline cache. If it was loaded from offline cache, we don't need to export the same data again. That's why I initialized cachedData at the beginning of LoadAll so that it can be accessed both inside and outside the catch block.
> 
> I get that. Not sure why it is initialized to string.Empty instead of null. The default initialization of a string is null so the fact that it's explicitly initialized to an empty string makes it look like it's for some purpose.

Good point, thanks! Reverted the last commit. We also don't need to check if `OfflineCache.Import` returned empty string because it always returns null if the encrypted data in offline cache file is invalid (or throws an exception if the file cannot be accessed). So a simple null check will suffice.

abhilasharora (approved):

PR #205, opened 2020-10-12
Update refresh scenario tests

avanigupta:
I found that some refresh scenario tests were actually not testing the intended functionality because of the following reasons:

- We were selecting/registering KVs based on just key name, whereas all the KVs in mock data also had a label. That led to unintentionally registering/selecting non-existent keys with null labels, and hence the test took a different code path than expected. 
For Eg: 

    - When `RefreshAsync `call triggers `RefreshIndividualKeyValues`, instead of calling `GetKeyValueChange `and processing the "changes", we load the setting again with `GetConfigurationSettingAsync`. That misses the intention of testing the `GetKeyValueChange `method.

    - When `LoadKeyValuesRegisteredForRefresh `is called, we should skip loading KV that have already loaded as part of `Select` clause. But since we registered a KV without its label, it's treated as a new KV and loaded again. 

- We were always returning the entire mock data collection of KV, regardless of what is present in the `Select` clause. This led to unintentional behavior in tests like `RefreshTests_RefreshRegisteredKeysAreLoadedOnStartup_DefaultUseQuery` and `RefreshTests_RefreshRegisteredKeysAreLoadedOnStartup_CustomUseQuery`. These tests would always pass regardless of `ConfigureRefresh` because the `Select` clause (or a lack thereof), ensures that all KVs from mock data collection are loaded.

To fix this, I have made the following changes to these tests:

1. I've explicitly added labels to every key that is registered for refresh. 
2. Because of # 1, I also had to register a new mock client setup for `GetConfigurationSettingAsync(ConfigurationSetting setting, bool cond, CancellationToken ct)` method in `GetMockConfigurationClient()`.
3. Because of # 2, I was able to refactor all the tests that had the same mock client setup and replace the duplicated code with the shared `GetMockConfigurationClient()` method.

_Originally posted by @avanigupta in https://github.com/Azure/AppConfiguration-DotnetProvider/pull/204#discussion_r501315115_

### DIFF
--- a/tests/Tests.AzureAppConfiguration/RefreshTests.cs
+++ b/tests/Tests.AzureAppConfiguration/RefreshTests.cs
@@ -19,8 +19,6 @@ namespace Tests.AzureAppConfiguration
 {
     public class RefreshTests
     {
-        string _connectionString = TestHelpers.CreateMockEndpointString();
-
         List<ConfigurationSetting> _kvCollection = new List<ConfigurationSetting>
         {
             ConfigurationModelFactory.ConfigurationSetting(
@@ -50,7 +48,30 @@ namespace Tests.AzureAppConfiguration
         [Fact]
         public void RefreshTests_RefreshRegisteredKeysAreLoadedOnStartup_DefaultUseQuery()
         {
-            var mockClient = GetMockConfigurationClient();
+            var serviceCollection = new List<ConfigurationSetting>(_kvCollection);
+            var mockResponse = new Mock<Response>();
+            var mockClient = new Mock<ConfigurationClient>(MockBehavior.Strict, TestHelpers.CreateMockEndpointString());
+
+            Response<ConfigurationSetting> GetTestKey(string k, string l, CancellationToken ct)
+            {
+                return Response.FromValue(serviceCollection.FirstOrDefault(s => (s.Key == k && s.Label == l)), mockResponse.Object);
+            }
+
+            Response<ConfigurationSetting> GetIfChanged(ConfigurationSetting setting, bool cond, CancellationToken ct)
+            {
+                var newSetting = serviceCollection.FirstOrDefault(s => (s.Key == setting.Key && s.Label == setting.Label));
+                var unchanged = (newSetting.Key == setting.Key && newSetting.Label == setting.Label && newSetting.Value == setting.Value);
+                var response = new MockResponse(unchanged ? 304 : 200);
+                return Response.FromValue(newSetting, response);
+            }
+
+            // Load all settings except the one registered for refresh - this test is to ensure that it will be loaded later
+            mockClient.Setup(c => c.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()))
+                .Returns(new MockAsyncPageable(serviceCollection.Where(s => s.Key != "TestKey1" && s.Label != "label").ToList()));
+            mockClient.Setup(c => c.GetConfigurationSettingAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()))
+                .ReturnsAsync((Func<string, string, CancellationToken, Response<ConfigurationSetting>>)GetTestKey);
+            mockClient.Setup(c => c.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()))
+                .ReturnsAsync((Func<ConfigurationSetting, bool, CancellationToken, Response<ConfigurationSetting>>)GetIfChanged);
 
             var config = new ConfigurationBuilder()
                 .AddAzureAppConfiguration(options =>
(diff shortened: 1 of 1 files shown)


tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -50,37 +48,59 @@ public class RefreshTests
         [Fact]
         public void RefreshTests_RefreshRegisteredKeysAreLoadedOnStartup_DefaultUseQuery()
         {
-            var mockClient = GetMockConfigurationClient();
+            var serviceCollection = new List<ConfigurationSetting>(_kvCollection);
+            var mockResponse = new Mock<Response>();
+            var mockClient = new Mock<ConfigurationClient>(MockBehavior.Strict, TestHelpers.CreateMockEndpointString());
+
+            Response<ConfigurationSetting> GetTestKey(string k, string l, CancellationToken ct)
```
abhilasharora, line 55: >string k, string l, CancellationToken ct [](start = 54, length = 40)

More descriptive names for the method arguments would make the code easier to read.
(k, l, ct) => (key, label, cancallationToken)

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -559,7 +510,7 @@ public void RefreshTests_RefreshAsyncThrowsOnRequestFailedException()
                     options.Select("TestKey*");
                     options.ConfigureRefresh(refreshOptions =>
                     {
-                        refreshOptions.Register("TestKey1")
+                        refreshOptions.Register("NonExistentKey") // Exception will be thrown only if a non-existent key is registered for refresh
```
abhilasharora, line 513: >Exception will be thrown only if a non-existent key is registered for refresh [](start = 69, length = 77)

I don't think it matters what key is used here. if we have configured the mock to throw an exception, it will be thrown. The comment could be confusing.

abhilasharora:
>             Assert.Equal("TestValue1", config["TestKey1"]);

We could use the same key registered in refresh for further validation, instead of using two separate keys.

---
Refers to: tests/Tests.AzureAppConfiguration/RefreshTests.cs:521 in 11b77fd. [](commit_id = 11b77fdf5e5515a1ec34c571f28cc3c62f1f7a1f, deletion_comment = False)

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -595,7 +546,7 @@ public void RefreshTests_TryRefreshAsyncReturnsFalseOnRequestFailedException()
...
```
abhilasharora, line 549: >Exception will be thrown only if a non-existent key is registered for refresh [](start = 69, length = 77)

Same comment as above test

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -959,13 +910,32 @@ private Mock<ConfigurationClient> GetMockConfigurationClient()
...
+            // We don't actually select KV based on SettingSelector, we just return a deep copy of _kvCollection
             mockClient.Setup(c => c.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()))
-                .Returns(new MockAsyncPageable(_kvCollection));
+                .Returns(() =>
+                {
+                    var copy = new List<ConfigurationSetting>();
+                    foreach (var setting in _kvCollection)
+                    {
+                        copy.Add(TestHelpers.CloneSetting(setting));
+                    };
```
abhilasharora, line 932: Here's an alternate way to clone the list.
List<ConfigurationSetting> copy = _kvCollection.Select(setting => TestHelpers.CloneSetting(setting)).ToList();

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -959,13 +910,32 @@ private Mock<ConfigurationClient> GetMockConfigurationClient()
...
+                    return new MockAsyncPageable(copy);
+                });
             mockClient.Setup(c => c.GetConfigurationSettingAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()))
                 .ReturnsAsync((Func<string, string, CancellationToken, Response<ConfigurationSetting>>)GetTestKey);
+            mockClient.Setup(c => c.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()))
+                .ReturnsAsync((Func<ConfigurationSetting, bool, CancellationToken, Response<ConfigurationSetting>>)GetIfChanged);
```
abhilasharora, line 938: I would leaving a blank line between each mock setup to improve readability.

tests/Tests.AzureAppConfiguration/RefreshTests.cs
avanigupta: In `RefreshIndividualKeyValues`, we check if the refresh registered KV exists in _settings cache. 

If it exists, we will call `GetKeyValueChange` which calls `GetConfigurationSettingAsync(ConfigurationSetting, bool, CancellationToken))`. 

If it doesn't exist, we call `GetConfigurationSettingAsync(string, string, CancellationToken))` - which is the method configured to throw the exception.

So we have two options - either register a non existent key or setup a different method to throw the exception. Since it doesn't really matter which method throws exception for the purpose of these two tests, I just changed the key to be explicitly non existent.

abhilasharora: I prefer the former approach over the latter for three reasons.

1. A user is more likely to use a key that exists as compared to one that does not.
2. Using a non-existent key distracts from the original test case that is being validated and is not necessary.
3. Implementation that requires no comments is generally easier to understand than one that requires a comment (despite the explanation).

---
In reply to: [503572923](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/205#discussion_r503572923) [](ancestors = 503572923)

avanigupta:
> 
> 
> > ```
> >         Assert.Equal("TestValue1", config["TestKey1"]);
> > ```
> 
> We could use the same key registered in refresh for further validation, instead of using two separate keys.
> 
> Refers to: tests/Tests.AzureAppConfiguration/RefreshTests.cs:521 in [11b77fd](https://github.com/Azure/AppConfiguration-DotnetProvider/commit/11b77fdf5e5515a1ec34c571f28cc3c62f1f7a1f). [](commit_id = [11b77fd](https://github.com/Azure/AppConfiguration-DotnetProvider/commit/11b77fdf5e5515a1ec34c571f28cc3c62f1f7a1f), deletion_comment = False)

What code does this refer to? I'm not able to find context for this comment.

tests/Tests.AzureAppConfiguration/RefreshTests.cs
avanigupta: Makes sense. Updated both tests.

abhilasharora (approved):

PR #208, opened 2020-10-26
Support refresh for multiple key-values with same key but different labels

### LINKED ISSUE #156: Allow refresh to be registered for multiple key-values with same key but different labels

avanigupta:
Fix #156 and #171

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -30,6 +30,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private readonly ConfigurationClient _client;
         private AzureAppConfigurationOptions _options;
         private ConcurrentDictionary<string, ConfigurationSetting> _settings;
+        private ConcurrentDictionary<KeyLabelIdentifier, ConfigurationSetting> _refreshRegisteredSettings = new ConcurrentDictionary<KeyLabelIdentifier, ConfigurationSetting>();
 
         private readonly TimeSpan MinCacheExpirationInterval;
         private readonly SemaphoreSlim InitializationSemaphore = new SemaphoreSlim(1);
@@ -316,10 +317,12 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             {
                 string watchedKey = changeWatcher.Key;
                 string watchedLabel = changeWatcher.Label;
+                KeyLabelIdentifier watchedKeyLabel = new KeyLabelIdentifier(watchedKey, watchedLabel);
 
                 // Skip the loading for the key-value in case it has already been loaded
                 if (data.ContainsKey(watchedKey) && data[watchedKey].Label == watchedLabel.NormalizeNull())
                 {
+                    _refreshRegisteredSettings[watchedKeyLabel] = data[watchedKey];
                     continue;
                 }
 
@@ -338,6 +341,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 if (watchedKv != null)
                 {
                     data[watchedKey] = watchedKv;
+                    _refreshRegisteredSettings[watchedKeyLabel] = watchedKv;
                 }
             }
         }
@@ -361,10 +365,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 {
                     bool hasChanged = false;
                     ConfigurationSetting watchedKv = null;
+                    KeyLabelIdentifier watchedKeyLabel = new KeyLabelIdentifier(watchedKey, watchedLabel);
 
-                    if (_settings.ContainsKey(watchedKey) && _settings[watchedKey].Label == watchedLabel.NormalizeNull())
+                    if (_refreshRegisteredSettings.ContainsKey(watchedKeyLabel))
                     {
-                        watchedKv = _settings[watchedKey];
+                        watchedKv = _refreshRegisteredSettings[watchedKeyLabel];
 
                         KeyValueChange keyValueChange = default;
                         await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, RequestType.Watch, _hostType,
@@ -381,8 +386,18 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                                 break;
                             }
 
-                            hasChanged = true;
-                            ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
+                            // If this key-value exists in _settings cache, it needs to be updated.
+                            if (_settings.ContainsKey(watchedKey) && _settings[watchedKey].Label == watchedLabel.NormalizeNull())
+                            {
+                                hasChanged = true;
+                                ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
+                            }
+                            else
+                            {
+                                // This is only a refresh registered setting with 'RefreshAll=false'.
+                                // Change in this key-value will not affect any other key-values in IConfiguration.
+                                _refreshRegisteredSettings[watchedKeyLabel] = keyValueChange.Current;
+                            }
                         }
                     }
                     else
@@ -410,6 +425,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                             }
 
                             hasChanged = true;
+                            _refreshRegisteredSettings[watchedKeyLabel] = watchedKv;
 
                             // Add the key-value if it is not loaded, or update it if it was loaded with a different label
                             _settings[watchedKey] = watchedKv;
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private readonly ConfigurationClient _client;
         private AzureAppConfigurationOptions _options;
         private ConcurrentDictionary<string, ConfigurationSetting> _applicationSettingsCache;
-        private ConcurrentDictionary<KeyLabelIdentifier, ConfigurationSetting> _watchedSettingsCache = new ConcurrentDictionary<KeyLabelIdentifier, ConfigurationSetting>();
+        private ConcurrentDictionary<KeyLabel, ConfigurationSetting> _watchedSettingsCache = new ConcurrentDictionary<KeyLabel, ConfigurationSetting>();
 
         private readonly TimeSpan MinCacheExpirationInterval;
         private readonly SemaphoreSlim InitializationSemaphore = new SemaphoreSlim(1);
@@ -317,14 +317,14 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             {
                 string watchedKey = changeWatcher.Key;
                 string watchedLabel = changeWatcher.Label;
-                KeyLabelIdentifier watchedKeyLabel = new KeyLabelIdentifier(watchedKey, watchedLabel);
+                KeyLabel watchedKeyLabel = new KeyLabel(watchedKey, watchedLabel);
 
                 // Skip the loading for the key-value in case it has already been loaded
-                if (data.ContainsKey(watchedKey) 
-                    && string.Equals(data[watchedKey].Key, watchedKey, StringComparison.Ordinal)
-                    && string.Equals(data[watchedKey].Label, watchedLabel.NormalizeNull(), StringComparison.Ordinal))
+                if (data.TryGetValue(watchedKey, out ConfigurationSetting loadedKv)
+                    && loadedKv.Key == watchedKey
+                    && loadedKv.Label == watchedLabel.NormalizeNull())
                 {
-                    _watchedSettingsCache[watchedKeyLabel] = data[watchedKey];
+                    _watchedSettingsCache[watchedKeyLabel] = loadedKv;
                     continue;
                 }
 
@@ -366,13 +366,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 try
                 {
                     bool hasChanged = false;
-                    ConfigurationSetting watchedKv = null;
-                    KeyLabelIdentifier watchedKeyLabel = new KeyLabelIdentifier(watchedKey, watchedLabel);
+                    KeyLabel watchedKeyLabel = new KeyLabel(watchedKey, watchedLabel);
 
-                    if (_watchedSettingsCache.ContainsKey(watchedKeyLabel))
+                    if (_watchedSettingsCache.TryGetValue(watchedKeyLabel, out ConfigurationSetting watchedKv))
                     {
-                        watchedKv = _watchedSettingsCache[watchedKeyLabel];
-
                         KeyValueChange keyValueChange = default;
                         await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, RequestType.Watch, _hostType,
                             async () => keyValueChange = await _client.GetKeyValueChange(watchedKv, CancellationToken.None).ConfigureAwait(false)).ConfigureAwait(false);
@@ -390,7 +387,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
                             if (keyValueChange.ChangeType == KeyValueChangeType.Deleted)
                             {
-                                _watchedSettingsCache.TryRemove(watchedKeyLabel, out ConfigurationSetting watchedSeting);
+                                _watchedSettingsCache.TryRemove(watchedKeyLabel, out ConfigurationSetting _);
                             }
                             else if (keyValueChange.ChangeType == KeyValueChangeType.Modified)
                             {
@@ -398,11 +395,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                             }
 
                             // If this key-value exists in _applicationSettingsCache, it needs to be updated.
-                            // Else, this is only a refresh registered setting which would be overwritten by another refresh registered setting with a different label.
-                            // Changes in this key-value will not affect any other key-values in application configuration.
-                            if (_applicationSettingsCache.ContainsKey(watchedKey)
-                                && string.Equals(_applicationSettingsCache[watchedKey].Key, watchedKey, StringComparison.Ordinal)
-                                && string.Equals(_applicationSettingsCache[watchedKey].Label, watchedLabel.NormalizeNull(), StringComparison.Ordinal))
+                            // Else, this is only a refresh registered setting which would be overwritten by another refresh registered setting with
+                            // a different label. Changes in this key-value will not affect any other key-values in application configuration.
+                            if (_applicationSettingsCache.TryGetValue(watchedKey, out ConfigurationSetting appSetting)
+                                && appSetting.Key == watchedKey
+                                && appSetting.Label == watchedLabel.NormalizeNull())
                             {
                                 hasChanged = true;
                                 ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -29,8 +29,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private readonly HostType _hostType;
         private readonly ConfigurationClient _client;
         private AzureAppConfigurationOptions _options;
-        private ConcurrentDictionary<string, ConfigurationSetting> _settings;
-        private ConcurrentDictionary<KeyLabelIdentifier, ConfigurationSetting> _refreshRegisteredSettings = new ConcurrentDictionary<KeyLabelIdentifier, ConfigurationSetting>();
+        private ConcurrentDictionary<string, ConfigurationSetting> _applicationSettingsCache;
+        private ConcurrentDictionary<KeyLabelIdentifier, ConfigurationSetting> _watchedSettingsCache = new ConcurrentDictionary<KeyLabelIdentifier, ConfigurationSetting>();
 
         private readonly TimeSpan MinCacheExpirationInterval;
         private readonly SemaphoreSlim InitializationSemaphore = new SemaphoreSlim(1);
@@ -141,7 +141,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         public async Task RefreshAsync()
         {
             // Check if initial configuration load had failed
-            if (_settings == null)
+            if (_applicationSettingsCache == null)
             {
                 await RefreshInitialConfiguration().ConfigureAwait(false);
                 return;
@@ -320,13 +320,15 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 KeyLabelIdentifier watchedKeyLabel = new KeyLabelIdentifier(watchedKey, watchedLabel);
 
                 // Skip the loading for the key-value in case it has already been loaded
-                if (data.ContainsKey(watchedKey) && data[watchedKey].Label == watchedLabel.NormalizeNull())
+                if (data.ContainsKey(watchedKey) 
+                    && string.Equals(data[watchedKey].Key, watchedKey, StringComparison.Ordinal)
+                    && string.Equals(data[watchedKey].Label, watchedLabel.NormalizeNull(), StringComparison.Ordinal))
                 {
-                    _refreshRegisteredSettings[watchedKeyLabel] = data[watchedKey];
+                    _watchedSettingsCache[watchedKeyLabel] = data[watchedKey];
                     continue;
                 }
 
-                // Send a request to retrieve key-value since it may be either not loaded or loaded with a different label
+                // Send a request to retrieve key-value since it may be either not loaded or loaded with a different label or different casing
                 ConfigurationSetting watchedKv = null;
                 try
                 {
@@ -341,7 +343,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 if (watchedKv != null)
                 {
                     data[watchedKey] = watchedKv;
-                    _refreshRegisteredSettings[watchedKeyLabel] = watchedKv;
+                    _watchedSettingsCache[watchedKeyLabel] = watchedKv;
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefreshOptions.cs
```diff
@@ -38,12 +38,17 @@ public AzureAppConfigurationRefreshOptions Register(string key, bool refreshAll)
         /// <param name="refreshAll">If true, a change in the value of this key refreshes all key-values being used by the configuration provider.</param>
         public AzureAppConfigurationRefreshOptions Register(string key, string label = LabelFilter.Null, bool refreshAll = false)
         {
-            RefreshRegistrations[key] = new KeyValueWatcher
+            if (string.IsNullOrEmpty(key))
+            {
+                throw new ArgumentNullException(nameof(key));
+            }
```
abhilasharora, line 44: This would be a breaking change.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,7 +208,14 @@ private async Task LoadAll(bool ignoreFailures)
                     {
                         await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsAsync(selector, CancellationToken.None).ConfigureAwait(false))
                         {
-                            serverData[setting.Key] = setting;
+                            if (serverData.ContainsKey(setting.Key))
```
abhilasharora, line 211: >if (serverData.ContainsKey(setting.Key)) [](start = 28, length = 40)

When would this condition be true in this code path?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -256,7 +270,12 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
                     
                     if (cachedData != null)
                     {
-                        data = JsonSerializer.Deserialize<IDictionary<string, ConfigurationSetting>>(cachedData);
+                        IDictionary<string, ConfigurationSetting> offlineCacheData = JsonSerializer.Deserialize<IDictionary<string, ConfigurationSetting>>(cachedData);
+
+                        foreach (KeyValuePair<string, ConfigurationSetting> setting in offlineCacheData)
+                        {
+                            data[setting.Key] = new List<ConfigurationSetting>() { setting.Value };
```
abhilasharora, line 277: >data [](start = 28, length = 4)

Have we validated this scenario? Where is the dictionary being initialized?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -286,7 +305,14 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
 
                 if (_options.OfflineCache != null && cachedData == null)
                 {
-                    _options.OfflineCache.Export(_options, JsonSerializer.Serialize(data));
+                    IDictionary<string, ConfigurationSetting> offlineCacheData = new Dictionary<string, ConfigurationSetting>();
+
+                    foreach (KeyValuePair<string, List<ConfigurationSetting>> setting in data)
+                    {
+                        offlineCacheData[setting.Key] = setting.Value.Last();
```
abhilasharora, line 312: >offlineCacheData[setting.Key] = setting.Value.Last(); [](start = 24, length = 53)

Do we need to create another dictionary and copy the last values, or could we re-use the current dictionary and serialize only the last values?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -310,16 +336,25 @@ private async Task RefreshInitialConfiguration()
...
 
                 // Skip the loading for the key-value in case it has already been loaded
-                if (data.ContainsKey(watchedKey) && data[watchedKey].Label == watchedLabel.NormalizeNull())
+                if (data.ContainsKey(watchedKey) && data[watchedKey].Any(kv => kv.Label == watchedLabel.NormalizeNull()))
                 {
+                    int index = data[watchedKey].FindIndex(kv => kv.Label == watchedLabel.NormalizeNull());
+
+                    if (index >= 0 && index < data[watchedKey].Count() - 1)
+                    {
+                        // Move this setting to the last position because refresh registered setting should overwrite previously loaded settings
```
abhilasharora, line 353: >// Move this setting to the last position because refresh registered setting should overwrite previously loaded settings [](start = 24, length = 120)

If the value for a key is coming from three sources (Select, Register 1, Register 2), then the value from last source (Register 2) should always be the last one. This code snippet would instead move the value being processed to the last position.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Good point, since we fetch everything with null label first, this condition would never be true. It would always go to the else block. I'll remove this check.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: If a setting is loaded multiple times, the data cache always needs to be updated with the most recently loaded value at the end of the value list. 
Say we have a TestKey with labels A, B, C, and all of them are loaded by Select in order (A,B,C). 
If key with label A is loaded by Register 1 and label B is loaded by Register 2 - we want the final value to be the one corresponding to label B.
The first iteration over ChangeWatchers will change the order of settings in `data` cache to (B,C,A).
The second iteration over ChangeWatchers will change the order of settings in `data` cache to (C,A,B).

abhilasharora: In your example, I would expect the key to always return the value corresponding to label B, instead of returning value for A at some times and B at other times, since the latter would result in inconsistent behavior. In case you feel that we might want to reorder them in some scenario, let's discuss it over a call.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Initialized and added test for validating this scenario.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefreshOptions.cs
avanigupta: Removed

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Fixed

jimmyca15:
Would we be able to avoid list operations and satisfy the provider if we used a dictionary key format of `$"{setting.Key}\n{setting.Label}"` ?

jimmyca15:
Does this cause a breaking change for offline file cache?

avanigupta:
> 
> 
> Would we be able to avoid list operations and satisfy the provider if we used a dictionary key format of `$"{setting.Key}\n{setting.Label}"` ?

If we store (Key+Label) as dictionary key, how do we track the order in which they were loaded, and therefore, how do we know which value ends up in IConfiguration if there are multiple KV with same key different label?

avanigupta:
> 
> 
> Does this cause a breaking change for offline file cache?

We Import and Export to Offline cache in the LoadAll() method only, and that has been modified such that offline cache functionality is unaffected by this change.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyLabelIdentifier.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using System;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Models
+{
+    internal struct KeyLabelIdentifier
```
abhilasharora, line 8: >KeyLabelIdentifier [](start = 20, length = 18)

How about `KeyLabel`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyLabelIdentifier.cs
```diff
@@ -0,0 +1,42 @@
...
+            Key = key;
+            Label = label;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj is KeyLabelIdentifier keyLabel)
+            {
+                return string.Equals(Key, keyLabel.Key, StringComparison.Ordinal)
+                    && string.Equals(Label, keyLabel.Label, StringComparison.Ordinal);
```
abhilasharora, line 31: How about `return Key == keyLabel.Key && Label == keyLabel.Label`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyLabelIdentifier.cs
abhilasharora, line 8: >KeyLabelIdentifier [](start = 20, length = 18)

Since we now have a separate dictionary to store watched settings, is it possible for us to reuse the `KeyValueWatcher` reference as the dictionary key and avoid introducing a new type?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -316,14 +317,18 @@ private async Task LoadKeyValuesRegisteredForRefresh(IDictionary<string, Configu
             {
                 string watchedKey = changeWatcher.Key;
                 string watchedLabel = changeWatcher.Label;
+                KeyLabelIdentifier watchedKeyLabel = new KeyLabelIdentifier(watchedKey, watchedLabel);
 
                 // Skip the loading for the key-value in case it has already been loaded
-                if (data.ContainsKey(watchedKey) && data[watchedKey].Label == watchedLabel.NormalizeNull())
+                if (data.ContainsKey(watchedKey) 
+                    && string.Equals(data[watchedKey].Key, watchedKey, StringComparison.Ordinal)
+                    && string.Equals(data[watchedKey].Label, watchedLabel.NormalizeNull(), StringComparison.Ordinal))
```
abhilasharora, line 325: We could use `==` operator to maintain consistency with existing code. It performs ordinal comparison by default.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -316,14 +317,18 @@ private async Task LoadKeyValuesRegisteredForRefresh(IDictionary<string, Configu
             {
                 string watchedKey = changeWatcher.Key;
                 string watchedLabel = changeWatcher.Label;
+                KeyLabelIdentifier watchedKeyLabel = new KeyLabelIdentifier(watchedKey, watchedLabel);
 
                 // Skip the loading for the key-value in case it has already been loaded
-                if (data.ContainsKey(watchedKey) && data[watchedKey].Label == watchedLabel.NormalizeNull())
+                if (data.ContainsKey(watchedKey) 
+                    && string.Equals(data[watchedKey].Key, watchedKey, StringComparison.Ordinal)
```
abhilasharora, line 324: >data[watchedKey] [](start = 37, length = 16)

The expression `data[watchedKey]` performs hash map lookup. We could do this once and re-use it in other places.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -316,14 +317,18 @@ private async Task LoadKeyValuesRegisteredForRefresh(IDictionary<string, Configu
             {
                 string watchedKey = changeWatcher.Key;
                 string watchedLabel = changeWatcher.Label;
+                KeyLabelIdentifier watchedKeyLabel = new KeyLabelIdentifier(watchedKey, watchedLabel);
 
                 // Skip the loading for the key-value in case it has already been loaded
-                if (data.ContainsKey(watchedKey) && data[watchedKey].Label == watchedLabel.NormalizeNull())
+                if (data.ContainsKey(watchedKey)
```
abhilasharora, line 323: >ContainsKey [](start = 25, length = 11)

We could use `TryGetValue` instead of `ContainsKey` to retrieve value at the same time.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -338,6 +343,7 @@ private async Task LoadKeyValuesRegisteredForRefresh(IDictionary<string, Configu
                 if (watchedKv != null)
                 {
                     data[watchedKey] = watchedKv;
+                    _watchedSettingsCache[watchedKeyLabel] = watchedKv;
```
abhilasharora, line 346: In case of Select("TestKey", "Label2") and Register("TestKey", "Label1").Register("TestKey", "Label2"), this implementation would fetch the value of ("TestKey", "Label2") again even though it was retrieved using Select.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -381,8 +388,25 @@ private async Task RefreshIndividualKeyValues()
                                 break;
                             }
 
-                            hasChanged = true;
-                            ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
+                            if (keyValueChange.ChangeType == KeyValueChangeType.Deleted)
+                            {
+                                _watchedSettingsCache.TryRemove(watchedKeyLabel, out ConfigurationSetting watchedSeting);
```
abhilasharora, line 393: >watchedSeting [](start = 106, length = 13)

`watchedSeting` -> Typo? We could also use a discard `_` since we aren't using this value.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -381,8 +388,25 @@ private async Task RefreshIndividualKeyValues()
...
+                            {
+                                _watchedSettingsCache[watchedKeyLabel] = keyValueChange.Current;
+                            }
+
+                            // If this key-value exists in _applicationSettingsCache, it needs to be updated.
+                            // Else, this is only a refresh registered setting which would be overwritten by another refresh registered setting with a different label.
+                            // Changes in this key-value will not affect any other key-values in application configuration.
+                            if (_applicationSettingsCache.ContainsKey(watchedKey)
+                                && string.Equals(_applicationSettingsCache[watchedKey].Key, watchedKey, StringComparison.Ordinal)
+                                && string.Equals(_applicationSettingsCache[watchedKey].Label, watchedLabel.NormalizeNull(), StringComparison.Ordinal))
```
abhilasharora, line 405: The dictionary lookup here can be optimized from three times to once.

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -854,6 +876,219 @@ Response<ConfigurationSetting> GetIfChanged(ConfigurationSetting setting, bool o
             Assert.Equal("newValue", config["TestKey3"]);
         }
 
+        [Fact]
+        public void RefreshTests_RefreshAllTrueForOverwrittenSentinelUpdatesEntireConfiguration()
+        {
+            var serviceCollection = new List<ConfigurationSetting>(_kvCollection);
```
abhilasharora, line 882: We may want to use a different name for this variable. `ServiceCollection` is a well-known class in .NET.
[https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.servicecollection?view=dotnet-plat-ext-5.0](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.servicecollection?view=dotnet-plat-ext-5.0)

avanigupta: I am using `serviceCollection` as the variable name here in order to maintain consistency with all other tests in this class.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyLabelIdentifier.cs
avanigupta: Since we need something to serve as dictionary key, I think `KeyLabel` helps us create a smaller object every time we do a dictionary lookup. `KeyValueWatcher` has a lot of other fields that we don't need for this scenario.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: That is correct. There is no clean way to keep track of which refresh registered key ends up in application configuration, apart from overwriting and letting the last KV win. Therefore, if there are same keys with multiple labels registered for refresh, we will fetch all of them again even if one of them was already fetched by the `Select` clause.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -316,14 +317,18 @@ private async Task LoadKeyValuesRegisteredForRefresh(IDictionary<string, Configu
             {
                 string watchedKey = changeWatcher.Key;
                 string watchedLabel = changeWatcher.Label;
+                KeyLabel watchedKeyLabel = new KeyLabel(watchedKey, watchedLabel);
```
jimmyca15, line 320: >KeyLabel [](start = 16, length = 8)

It has properties `Key` and `Label`, but I'm thinking the type itself is more of `KeyValueIdentifier`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -316,14 +317,18 @@ private async Task LoadKeyValuesRegisteredForRefresh(IDictionary<string, Configu
             {
                 string watchedKey = changeWatcher.Key;
                 string watchedLabel = changeWatcher.Label;
+                KeyLabel watchedKeyLabel = new KeyLabel(watchedKey, watchedLabel);
 
                 // Skip the loading for the key-value in case it has already been loaded
-                if (data.ContainsKey(watchedKey) && data[watchedKey].Label == watchedLabel.NormalizeNull())
+                if (data.TryGetValue(watchedKey, out ConfigurationSetting loadedKv)
+                    && loadedKv.Key == watchedKey
+                    && loadedKv.Label == watchedLabel.NormalizeNull())
```
jimmyca15, line 325: >watchedLabel.NormalizeNull() [](start = 41, length = 28)

I'm thinking we can move this into the `KeyLabel` class equals comparison and use that here to see if they are queal.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -381,8 +385,25 @@ private async Task RefreshIndividualKeyValues()
                                 break;
                             }
 
-                            hasChanged = true;
-                            ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
+                            if (keyValueChange.ChangeType == KeyValueChangeType.Deleted)
```
jimmyca15, line 388: >keyValueChange.ChangeType == KeyValueChangeType.Deleted [](start = 32, length = 55)

Don't we need to do this before breaking from the loop? Seems like otherwise we will be storing an outdated kv.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -412,13 +433,14 @@ private async Task RefreshIndividualKeyValues()
                             hasChanged = true;
 
                             // Add the key-value if it is not loaded, or update it if it was loaded with a different label
-                            _settings[watchedKey] = watchedKv;
+                            _applicationSettingsCache[watchedKey] = watchedKv;
```
jimmyca15, line 436: >_applicationSettingsCache[watchedKey] = watchedKv [](start = 28, length = 49)

Same as my previous comment.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -381,8 +385,25 @@ private async Task RefreshIndividualKeyValues()
...
+                            }
+                            else if (keyValueChange.ChangeType == KeyValueChangeType.Modified)
+                            {
+                                _watchedSettingsCache[watchedKeyLabel] = keyValueChange.Current;
+                            }
+
+                            // If this key-value exists in _applicationSettingsCache, it needs to be updated.
+                            // Else, this is only a refresh registered setting which would be overwritten by another refresh registered setting with
+                            // a different label. Changes in this key-value will not affect any other key-values in application configuration.
+                            if (_applicationSettingsCache.TryGetValue(watchedKey, out ConfigurationSetting appSetting)
```
jimmyca15, line 400: >_applicationSettingsCache.TryGetValue(watchedKey, out ConfigurationSetting appSetting [](start = 32, length = 85)

I'm not sure if this is necessary. Might need a bit of help on this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -29,7 +29,8 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
         private readonly HostType _hostType;
         private readonly ConfigurationClient _client;
         private AzureAppConfigurationOptions _options;
-        private ConcurrentDictionary<string, ConfigurationSetting> _settings;
+        private ConcurrentDictionary<string, ConfigurationSetting> _applicationSettingsCache;
```
jimmyca15, line 32: >_applicationSettingsCache [](start = 67, length = 25)

`Cache` isn't needed in the name.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: I thought more about this. If we change the cached values before the `break` statement, and lets say `refreshAll` operation fails due to some `RequestFailedException`, we have now cached the latest value of a sentinel key without updating application configuration. On the next call to `TryRefreshAsync`, since we already have the latest value of sentinel key, we will not be able to observe any changes to the sentinel key.
Instead of changing the logic here, I think it would be better to clear the `_watchedSettings` cache every time `LoadKeyValuesRegisteredForRefresh` is called (since we are always going to fetch the latest values of refresh registered keys anyway).

jimmyca15: Won't we have that problem because of this line already? https://github.com/Azure/AppConfiguration-DotnetProvider/blob/2ab6bd32470523d918cee545bc2e135d2b20fb20/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L377

avanigupta: Yes, the next refresh operation will be triggered only after the cache expires again. This is our behavior even today. If the first refresh failed, wait for `CacheExpires` time before triggering another refresh.

jimmyca15: I remember Abhilash and I discussed this in the past and ultimately this is a known behavior.

tests/Tests.AzureAppConfiguration/RefreshTests.cs
jimmyca15: I don't see any problem fixing it in just this new test.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Talked with Avani. It may look like the update being here could cause us to maintain stale key-values in the watchedsettings list, but actually the refreshAll operation will cause the updated values to be used.

Secondly, if we were to move it above the refreshAll + break, then we would never trigger a refresh again on a failure. Right now, if there is a failure in reload all, we would trigger the refresh again when the watcher's cache expires.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyLabelIdentifier.cs
avanigupta: We'll stick with KeyValueIdentifier for now and we can discuss it again when you are back.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Discussed offline with Jimmy. From users' point of view, every refresh registered setting will be fetched in a new server call. Avoiding some of these calls is an optimization we provide, but don't necessarily promise. In this case, the optimization would make the code would overly complicated and is not worth it.

Issue #209, opened 2020-11-05
Allow users to configure the behavior when a Key Vault reference cannot be resolved


avanigupta:
### Problem

The recommended way to load key-values from App Configuration is to specify only what your application needs to use in the `Select` clause(s). This can be achieved by name-spacing key-values with a prefix or placing key-values into separate labels and loading only the required labels. 

This rule of thumb also applies to any Key Vault references that are stored in App Configuration. If your application loads a Key Vault reference, and you haven't configured the right credentials needed to fetch the secret value from Key Vault, an exception will be thrown. However, in certain situations, you may not be able to rely on filtering the Key Vault references and need more control over what happens after Key Vault references are fetched from App Configuration. Common scenarios when you might need a workaround for avoiding this exception:

- Your application doesn't need any secrets from Key Vault.
- Your application is configured to resolve some secrets from Key Vault, but you don't want your application to resolve any new Key Vault references that may be added to your App Config store in future.

### Proposed Solution

Configuration provider should not implicitly hide the exception because that would leave the application configuration in an undefined state. In order to have a deterministic value for any unresolved Key Vault references, it's best that users explicitly control what they want to do when their application loads a Key Vault reference whose value should not be fetched from Key Vault.
For this, we can add a new API that allows users to set a callback which will be invoked to resolve Key Vault references that do not have a matching SecretClient registered. 

```cs
public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTask<string>> secretResolver)
```

This callback takes the Key Vault secret identifier (URI) as input and returns the value that will be added to your application configuration corresponding to the Key Vault reference key.

### Usage Example

If your App Configuration has the following Key Vault references:

Key | Value
--- |---
KeyVaultRef1 |  {"uri":"https://abc.vault.azure.net/secrets/SecretName"}
KeyVaultRef2 |  {"uri":"https://xyz.vault.azure.net/secrets/SecretName"}
KeyVaultRef3 |  {"uri":"https://unknown1.vault.azure.net/secrets/SecretName"}
KeyVaultRef4 |  {"uri":"https://unknown2.vault.azure.net/secrets/SecretName"}

The secret resolver callback can be configured as follows:

```cs
var builder = new ConfigurationBuilder();
builder.AddAzureAppConfiguration(options =>
{
    options.Connect(connectionString)
           .Select("KeyVaultRef*")
           .ConfigureKeyVault(kv =>
           {
[...]

PR #210, opened 2020-11-06
Add SetSecretResolver API for configuring default value of Key Vault references

avanigupta:
Implementation for proposed design in #209

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
@@ -39,11 +39,19 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         }
 
         /// <summary>
-        /// Sets the callback to be invoked for resolving those key vault references that have no registered <see cref="SecretClient"/>.
+        /// Sets the callback used to resolve key vault references that have no registered <see cref="SecretClient"/>.
         /// </summary>
         /// <param name="secretResolver">A callback which returns a value for the given Key Vault reference.</param>
+        /// <remarks>
+        /// The input to the callback will be a <see cref="Uri"/> in key vault reference format: "https://{vault-name}.vault.azure.net/{type}/{name}/{version}". 
+        /// </remarks>
         public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTask<string>> secretResolver)
         {
+            if (secretResolver == null)
+            {
+                throw new ArgumentNullException(nameof(secretResolver));
+            }
+
             SecretResolver = secretResolver;
             return this;
         }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         /// </summary>
         /// <param name="secretResolver">A callback which returns a value for the given Key Vault reference.</param>
         /// <remarks>
-        /// The input to the callback will be a <see cref="Uri"/> in key vault reference format: "https://{vault-name}.vault.azure.net/{type}/{name}/{version}". 
+        /// The input to the callback will be a key vault secret <see cref="Uri"/>, for example "https://{vault-name}.vault.azure.net/secrets/{name}/{version}"
         /// </remarks>
         public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTask<string>> secretResolver)
         {
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
@@ -3,7 +3,9 @@
 //
 using Azure.Core;
 using Azure.Security.KeyVault.Secrets;
+using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
@@ -14,6 +16,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     {
         internal TokenCredential Credential;
         internal List<SecretClient> SecretClients = new List<SecretClient>();
+        internal Func<Uri, ValueTask<string>> SecretResolver;
 
         /// <summary>
         /// Sets the credentials used to authenticate to key vaults that have no registered <see cref="SecretClient"/>.
@@ -32,6 +35,16 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         public AzureAppConfigurationKeyVaultOptions Register(SecretClient secretClient)
         {
             SecretClients.Add(secretClient);
+            return this;
+        }
+
+        /// <summary>
+        /// Sets the callback to be invoked for resolving those key vault references that have no registered <see cref="SecretClient"/>.
+        /// </summary>
+        /// <param name="secretResolver">A callback which returns a value for the given Key Vault reference.</param>
+        public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTask<string>> secretResolver)
+        {
+            SecretResolver = secretResolver;
             return this;
         }
     }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         /// <summary>
         /// Sets the callback used to resolve key vault references that have no registered <see cref="SecretClient"/>.
         /// </summary>
-        /// <param name="secretResolver">A callback which returns a value for the given Key Vault reference.</param>
+        /// <param name="secretResolver">A callback that maps the <see cref="Uri"/> of the secret to its value.</param>
         /// <remarks>
         /// The input to the callback will be a key vault secret <see cref="Uri"/>, for example "https://{vault-name}.vault.azure.net/secrets/{name}/{version}"
         /// </remarks>
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
@@ -41,10 +41,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         /// <summary>
         /// Sets the callback used to resolve key vault references that have no registered <see cref="SecretClient"/>.
         /// </summary>
-        /// <param name="secretResolver">A callback that maps the <see cref="Uri"/> of the secret to its value.</param>
-        /// <remarks>
-        /// The input to the callback will be a key vault secret <see cref="Uri"/>, for example "https://{vault-name}.vault.azure.net/secrets/{name}/{version}"
-        /// </remarks>
+        /// <param name="secretResolver">A callback that maps the <see cref="Uri"/> of the key vault secret to its value.</param>
         public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTask<string>> secretResolver)
         {
             if (secretResolver == null)
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -296,6 +296,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             var keyVaultOptions = new AzureAppConfigurationKeyVaultOptions();
             configure?.Invoke(keyVaultOptions);
 
+            if (keyVaultOptions.Credential != null && keyVaultOptions.SecretResolver != null)
+            {
+                throw new InvalidOperationException($"Cannot configure both default credentials and secret resolver for Key Vault references. Please call only one of {nameof(keyVaultOptions.SetCredential)}() or {nameof(keyVaultOptions.SetSecretResolver)}() methods.");
+            }
+
             _adapters.RemoveAll(a => a is AzureKeyVaultKeyValueAdapter);
             _adapters.Add(new AzureKeyVaultKeyValueAdapter(new AzureKeyVaultSecretProvider(keyVaultOptions.Credential, keyVaultOptions.SecretClients, keyVaultOptions.SecretResolver)));
 
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -297,7 +297,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             configure?.Invoke(keyVaultOptions);
 
             _adapters.RemoveAll(a => a is AzureKeyVaultKeyValueAdapter);
-            _adapters.Add(new AzureKeyVaultKeyValueAdapter(new AzureKeyVaultSecretProvider(keyVaultOptions.Credential, keyVaultOptions.SecretClients)));
+            _adapters.Add(new AzureKeyVaultKeyValueAdapter(new AzureKeyVaultSecretProvider(keyVaultOptions.Credential, keyVaultOptions.SecretClients, keyVaultOptions.SecretResolver)));
 
             return this;
         }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -298,7 +298,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             if (keyVaultOptions.Credential != null && keyVaultOptions.SecretResolver != null)
             {
-                throw new InvalidOperationException($"Cannot configure both default credentials and secret resolver for Key Vault references. Please call only one of {nameof(keyVaultOptions.SetCredential)}() or {nameof(keyVaultOptions.SetSecretResolver)}() methods.");
+                throw new InvalidOperationException($"Cannot configure both default credentials and secret resolver for Key Vault references. Please call either {nameof(keyVaultOptions.SetCredential)} or {nameof(keyVaultOptions.SetSecretResolver)} method, not both.");
             }
 
             _adapters.RemoveAll(a => a is AzureKeyVaultKeyValueAdapter);
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
@@ -19,11 +19,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault
 
         public AzureKeyVaultSecretProvider(TokenCredential credential = null, IEnumerable<SecretClient> secretClients = null, Func<Uri, ValueTask<string>> secretResolver = null)
         {
-            if (credential != null && secretResolver != null)
-            {
-                throw new InvalidOperationException($"Cannot configure both default credentials and secret resolver for Key Vault references. Please call only one of 'SetCredential()' or 'SetSecretResolver()' methods.");
-            }
-
             _credential = credential;
             _secretClients = new Dictionary<string, SecretClient>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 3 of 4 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -45,7 +52,14 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
-                throw new UnauthorizedAccessException("No key vault credential configured and no matching secret client could be found.");
+                if (_secretResolver == null)
+                {
+                    throw new UnauthorizedAccessException("No key vault credential or secret resolver callback configured, and no matching secret client could be found.");
+                }
+                else
+                {
+                    return await _secretResolver(secretUri);
+                }
             }
```
abhilasharora, line 63: We could refactor this a little bit to avoid nested if statement.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -15,12 +15,19 @@ internal class AzureKeyVaultSecretProvider
...
         private readonly TokenCredential _credential;
+        private readonly Func<Uri, ValueTask<string>> _secretResolver;
 
-        public AzureKeyVaultSecretProvider(TokenCredential credential = null, IEnumerable<SecretClient> secretClients = null)
+        public AzureKeyVaultSecretProvider(TokenCredential credential = null, IEnumerable<SecretClient> secretClients = null, Func<Uri, ValueTask<string>> secretResolver = null)
         {
+            if (credential != null && secretResolver != null)
+            {
+                throw new InvalidOperationException($"Cannot configure both default credentials and secret resolver for Key Vault references. Please call only one of 'SetCredential()' or 'SetSecretResolver()' methods.");
+            }
```
abhilasharora, line 25: Could this check be moved earlier in the SetCredential and SetSecretResolver methods? Or alternatively in the ConfigureKeyVault method?

tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
```diff
@@ -328,7 +328,7 @@ public void ThrowsWhenNoMatchingSecretClientIsFound()
         }
 
         [Fact]
-        public void ThrowsWhenSecretClientAndCredentialAreMissing()
+        public void ThrowsWhenSecretClientAndSecretResolverAndDefaultCredentialAreMissing()
```
abhilasharora, line 331: >ThrowsWhenSecretClientAndSecretResolverAndDefaultCredentialAreMissing [](start = 20, length = 69)

We could also use `ThrowsWhenConfigureKeyVaultIsMissing` or something similar, so we don't have to maintain it as we keep adding new parameters.

tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
```diff
@@ -367,5 +367,122 @@ public void DoesNotThrowKeyVaultExceptionWhenProviderIsOptional()
...
+                    {
+                        kv.SetSecretResolver((secretUri) =>
+                        {
+                            return new ValueTask<string>(secretUri.ToString());
+                        });
+                    });
+                })
+                .Build();
+
+            Assert.Equal("https://keyvault-theclassics.vault.azure.net/secrets/TheTrialSecret", config["TestKey1"]);
```
abhilasharora, line 393: >https://keyvault-theclassics.vault.azure.net/secrets/TheTrialSecret [](start = 26, length = 67)

We can extract this out in a variable to maintain a single source of truth.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -34,5 +37,15 @@ public AzureAppConfigurationKeyVaultOptions Register(SecretClient secretClient)
             SecretClients.Add(secretClient);
             return this;
         }
+
+        /// <summary>
+        /// Sets the callback to be invoked for resolving those key vault references that have no registered <see cref="SecretClient"/>.
+        /// </summary>
+        /// <param name="secretResolver">A callback which returns a value for the given Key Vault reference.</param>
+        public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTask<string>> secretResolver)
```
abhilasharora, line 45: >secretResolver [](start = 99, length = 14)

We can throw `ArgumentNullException` when `secretResolver` is `null`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -34,5 +37,15 @@ public AzureAppConfigurationKeyVaultOptions Register(SecretClient secretClient)
             SecretClients.Add(secretClient);
             return this;
         }
+
+        /// <summary>
+        /// Sets the callback to be invoked for resolving those key vault references that have no registered <see cref="SecretClient"/>.
```
abhilasharora, line 42: > Sets the callback to be invoked for resolving those key vault references that have no registered <see cref="SecretClient"/>. [](start = 11, length = 125)

This can be rephrased as follows to keep it consistent with the summary comments for other methods.

`Sets the callback used to resolve key vault references that have no registered <see cref="SecretClient"/>.`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
abhilasharora, line 45: >Uri [](start = 75, length = 3)

We could share a sample Uri format in the summary comments, so users understand the input to the callback.

tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
```diff
@@ -367,5 +367,122 @@ public void DoesNotThrowKeyVaultExceptionWhenProviderIsOptional()
...
+                        });
+
+                        kv.SetCredential(new DefaultAzureCredential());
+                    });
+                })
+                .Build();
+            });
+
+            Assert.NotNull(ex.InnerException);
+            Assert.IsType<InvalidOperationException>(ex.InnerException);
```
abhilasharora, line 423: >Assert.IsType<InvalidOperationException>(ex.InnerException); [](start = 12, length = 60)

Why would it throw `InvalidOperationException` as inner exception instead of top-level exception?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
jimmyca15: My vote would be in the ConfigureKeyVault method. The earlier the better.

tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
avanigupta: The `Build()` method wraps `InvalidOperationException` as `ArgumentException` here: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/96abfe3af854bde6d92860e74ab59866c1df03ca/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs#L55-L61

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -34,5 +37,23 @@ public AzureAppConfigurationKeyVaultOptions Register(SecretClient secretClient)
             SecretClients.Add(secretClient);
             return this;
         }
+
+        /// <summary>
+        /// Sets the callback used to resolve key vault references that have no registered <see cref="SecretClient"/>.
+        /// </summary>
+        /// <param name="secretResolver">A callback which returns a value for the given Key Vault reference.</param>
+        /// <remarks>
+        /// The input to the callback will be a <see cref="Uri"/> in key vault reference format: "https://{vault-name}.vault.azure.net/{type}/{name}/{version}".
```
abhilasharora, line 46: ><see cref="Uri"/> in key vault reference format [](start = 48, length = 47)

How about "<see cref="Uri"/> to the secret"?

abhilasharora, line 46: >{type} [](start = 135, length = 6)

The type would always be "secrets" since this method is specific to secrets.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -40,16 +42,25 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
 
-            if (client == null)
+            if (client != null)
+            {
+                KeyVaultSecret secret = await client.GetSecretAsync(secretName, secretVersion, cancellationToken).ConfigureAwait(false);
+                secretValue = secret?.Value;
+            }
+            else if (_secretResolver != null)
+            {
+                secretValue = await _secretResolver(secretUri);
```
abhilasharora, line 56: >await _secretResolver(secretUri) [](start = 30, length = 32)

This await statement is missing a `ConfigureAwait(false)`.

tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
abhilasharora: Do we understand why? It seems odd if the user receives an `ArgumentException` when both default credentials and secret resolver are specified.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
avanigupta: Not necessarily - the `GetSecretAsync` API can resolve certificates too. And we allow users to create references to certificates in App Config.
For example, both of these highlighted references resolve to the same certificate:

image

abhilasharora: `GetSecretAsync` API only resolves secrets. Each certificate in key vault consists of certificate metadata, key and secret, and only the secret part is retrieved. The certificate URL works since we don't have validation on the URI. Even if you change "/certificates/" to "/dummy/", the configuration provider would still fetch the secret.

tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
avanigupta: `ConfigureKeyVault` (or any viable code path where we can throw this exception) is called inside the `ConfigurationBuilder.Build()` method. But I am not sure why we specifically catch `InvalidOperationException` here and throw `ArgumentException` instead.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -34,5 +37,23 @@ public AzureAppConfigurationKeyVaultOptions Register(SecretClient secretClient)
             SecretClients.Add(secretClient);
             return this;
         }
+
+        /// <summary>
+        /// Sets the callback used to resolve key vault references that have no registered <see cref="SecretClient"/>.
+        /// </summary>
+        /// <param name="secretResolver">A callback which returns a value for the given Key Vault reference.</param>
```
abhilasharora, line 44: >A callback which returns a value for the given Key Vault reference. [](start = 41, length = 67)

We can also include details on what the return value `ValueTask<string>` for the callback means.
Example : `A callback that maps the <see cref="Uri"/> of the secret with its value`

tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
abhilasharora: I think we should re-assess if we can let the `InvalidOperationException` bubble up. Since this would be a breaking change, we can skip it for now, but its something we can re-visit in the future.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -296,8 +296,13 @@ public AzureAppConfigurationOptions ConfigureKeyVault(Action<AzureAppConfigurati
             var keyVaultOptions = new AzureAppConfigurationKeyVaultOptions();
             configure?.Invoke(keyVaultOptions);
 
+            if (keyVaultOptions.Credential != null && keyVaultOptions.SecretResolver != null)
+            {
+                throw new InvalidOperationException($"Cannot configure both default credentials and secret resolver for Key Vault references. Please call only one of {nameof(keyVaultOptions.SetCredential)}() or {nameof(keyVaultOptions.SetSecretResolver)}() methods.");
```
abhilasharora, line 301: >Please call only one of {nameof(keyVaultOptions.SetCredential)}() or {nameof(keyVaultOptions.SetSecretResolver)}() methods. [](start = 142, length = 123)

We could remove the opening and closing parentheses, similar to the error message in the following link.
[https://github.com/Azure/AppConfiguration-DotnetProvider/blob/1d2a0cb24f0e193907211be4e6d8df0d0a47606e/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs#L50](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/1d2a0cb24f0e193907211be4e6d8df0d0a47606e/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs#L50)

Suggestion : "Please call either {nameof(keyVaultOptions.SetCredential)} or {nameof(keyVaultOptions.SetSecretResolver)} method."

abhilasharora (approved):
:shipit:

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -296,8 +296,13 @@ public AzureAppConfigurationOptions ConfigureKeyVault(Action<AzureAppConfigurati
             var keyVaultOptions = new AzureAppConfigurationKeyVaultOptions();
             configure?.Invoke(keyVaultOptions);
 
+            if (keyVaultOptions.Credential != null && keyVaultOptions.SecretResolver != null)
+            {
+                throw new InvalidOperationException($"Cannot configure both default credentials and secret resolver for Key Vault references. Please call either {nameof(keyVaultOptions.SetCredential)} or {nameof(keyVaultOptions.SetSecretResolver)} method, not both.");
```
jimmyca15, line 301: Why are we throwing this here instead of in the options?

jimmyca15: Oh apparently I suggested this in a comment. Sorry, I should have suggested that it be in `AzureAppConfigurationKeyVaultOptions`

jimmyca15: Now that I look at this again, having it here is even better than what I was originally thinking. Resolving this comment.

avanigupta: In `AzureAppConfigurationKeyVaultOptions`, we'd need to add this check in 2 places - `SetCredential` and `SetSecretResolver` (as opposed to just one check in `ConfigureKeyVault`).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -34,5 +37,23 @@ public AzureAppConfigurationKeyVaultOptions Register(SecretClient secretClient)
             SecretClients.Add(secretClient);
             return this;
         }
+
+        /// <summary>
+        /// Sets the callback used to resolve key vault references that have no registered <see cref="SecretClient"/>.
+        /// </summary>
+        /// <param name="secretResolver">A callback that maps the <see cref="Uri"/> of the secret to its value.</param>
+        /// <remarks>
+        /// The input to the callback will be a key vault secret <see cref="Uri"/>, for example "https://{vault-name}.vault.azure.net/secrets/{name}/{version}"
```
jimmyca15, line 46: Can we omit the URL example. We don't own the format and it looks like we are documenting exactly what the URL will look like. Also, in different clouds, that url format will change.

avanigupta: In that case, we don't need this remark at all. The param summary should be sufficient.

jimmyca15: Oh, yeah I suppose you're right.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
jimmyca15: That's true, but that's not the reasoning that sold me on this. I like how with the way you have it now that `AzureAppConfigurationKeyVaultOptions` can be a POCO and `ConfigureKeyVault` can be smart and contain the business logic. Having simple objects is something I find valuable.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
abhilasharora: I think the example Uri is useful to understand what the input to the callback would look like. Even though we don't own the format, our current implementation depends on the Uri to be in this format. We could change `{vault-name}.vault.azure.net` to `vaultBaseUri`.

jimmyca15: How does our current implementation rely on that format? If it does, we may have a problem.

abhilasharora: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/1d2a0cb24f0e193907211be4e6d8df0d0a47606e/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs#L41-L42

jimmyca15: I see we should start using [KeyVaultSecretIdentifier.Parse](https://docs.microsoft.com/en-us/dotnet/api/azure.security.keyvault.secrets.keyvaultsecretidentifier.parse?view=azure-dotnet-preview). Since this next release will be adding a new API, and bumping a minor version. It will probably be a good candidate release for us to bump to [Azure.Security.KeyVault.Secrets 4.1.0](https://www.nuget.org/packages/Azure.Security.KeyVault.Secrets/4.1.0) to use this API.

We have no need to document the URI format. We shouldn't be aware of it.

avanigupta: Sure, I can create a new PR for these changes. 
I'll remove the URL example from summary here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -34,5 +37,20 @@ public AzureAppConfigurationKeyVaultOptions Register(SecretClient secretClient)
             SecretClients.Add(secretClient);
             return this;
         }
+
+        /// <summary>
+        /// Sets the callback used to resolve key vault references that have no registered <see cref="SecretClient"/>.
+        /// </summary>
+        /// <param name="secretResolver">A callback that maps the <see cref="Uri"/> of the key vault secret to its value.</param>
+        public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTask<string>> secretResolver)
```
jimmyca15, line 45: `ValueTask` indicates that we expect implementations to typically return in an a synchronous manner. Is this on purpose? I would expect secret resolution to be mostly an async operation.

avanigupta: We should allow both sync and async operations because we don't want to assume how users resolve their secrets.

jimmyca15: Resolved this with Drago. `ValueTask` gives the chance for developers to be efficient when implementing an interface via synchronous completion. Whereas, if we went with `Task`, there wouldn't be any benefit over `ValueTask`. We do not rely on `Task` anywere (no `ContinueWith`) so `ValueTask` is a good option for us here.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
jimmyca15: Leaving Stephen Toub's comment regarding `ValueTask` here as well for future reference.
https://github.com/dotnet/runtime/issues/25182#issuecomment-370609431

Issue #203, opened 2020-09-21
Allow remaining configuration to load when a key-vault reference fails to resolve


abhilasharora:
_[Creating this issue on behalf of @b10-dslappendel as reported in issue [#136](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/136)]_

I see that the latest version 4.0 ignores exceptions thrown from keyvault. However, I think I may have misunderstood the use case here. From what I can see, right now the whole configuration is seen as optional, meaning that if at least one keyvault exception is thrown, no configuration values are loaded.

I had hoped that this would cover our use case, which is ignoring only configuration values from keyvault that are inaccessible (e.g. due to no access permissions), instead of ignoring all configuration values if one can't be loaded from keyvault.

Basically, we have hooked up multiple keyvaults in one app configuration, but not all apps have access to all keyvaults that are referenced there. So right now we can't load any configuration if we don't allow access to all keyvaults from all our apps.

In preview version 2, I could still create a custom keyvault delegating handler that ignores 401, 403 and 404 errors and return an empty string, but I also can't seem to find this option anymore in version 4.

Is there any way to support this use case?

abhilasharora:
@b10-dslappendel Thanks for your feedback. I'm looking into ways to support the use case you shared. The recommended approach is to allow each application to only `Select` configuration settings that are needed for that application. The configuration provider will only attempt to load settings defined by the criteria defined by the `Select` clause(s).

In your scenario, is there a reason the application is selecting key-vault references that correspond to key-vaults that it isn't expected to have access to?

b10-dslappendel:
@abhilasharora Thanks for moving it in a new issue.
In our scenario, we have a bunch of apps that make use of the same app configuration instance. We have multiple key vaults, but for some higher privileged key vaults, most apps and users shouldn't have access to those unless it's really necessary to pass audits. However, since they are still referenced in app configuration, and they are still selected (because we currently use `KeyFilter.All`), they fail to be retrieved.

I'm currently discussing with my team to see if we can improve this on our end. Currently we notice that our startup times are getting longer and longer the more settings we add to app configuration and/or key vault.
Using `KeyFilter.All` isn't the best way to load configurations if 80% of those settings isn't used in one particular app. Currently we aren't using namespaces because we might have been too enthusiastic during preview last year. Namespaces do allow us to select them in a more granular way, but the transition to this is going to be very painful because some settings are used across apps without anything that binds them together in a category.
[...]

PR #211, opened 2020-11-06
Change description of optional parameter for AddAzureAppConfiguration method

avanigupta:
Modifying the `optional` parameter description in `AddAzureAppConfiguration` method to avoid confusion as reported in #203

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Extensions.Configuration
         /// </summary>
         /// <param name="configurationBuilder">The configuration builder to add key-values to.</param>
         /// <param name="connectionString">The connection string used to connect to the configuration store.</param>
-        /// <param name="optional">If true, and if an exception occurs while loading key-values from App Configuration store, this method will neither throw the exception nor load any key-values from App Configuration store.</param>
+        /// <param name="optional">Determines the behavior of the App Configuration provider when an exception occurs. If false, the exception is thrown. If true, the exception is suppressed and no settings are populated from Azure App Configuration.</param>
         /// <returns>The provided configuration builder.</returns>
         public static IConfigurationBuilder AddAzureAppConfiguration(
             this IConfigurationBuilder configurationBuilder,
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
```diff
@@ -17,7 +17,7 @@ public static class AzureAppConfigurationExtensions
         /// </summary>
         /// <param name="configurationBuilder">The configuration builder to add key-values to.</param>
         /// <param name="connectionString">The connection string used to connect to the configuration store.</param>
-        /// <param name="optional">If true, this method will not throw an exception if the configuration store cannot be accessed.</param>
+        /// <param name="optional">If true, and if an exception occurs while loading key-values from App Configuration store, this method will neither throw the exception nor load any key-values from App Configuration store.</param>
```
abhilasharora, line 20: If a configuration store cannot be accessed, isn't it implicit that the key-values will not be loaded?

avanigupta: Sure, but we ignore key-values from App Config on other exceptions too (like KeyVaultReferenceException, RequestFailedException or OperationCanceledException) which can occur even if the store is accessible. So even though the exception may have occurred while processing a single key-value, we ignore all key-values from App Config.

jimmyca15: It seems to cover a bit too much to me.

I would go with "If true, any exceptions that occur while building the Azure App Configuration configuration provider will be suppressed and cause the provider to be empty. "

Let me know what you think.

abhilasharora: Here's another suggestion to keep it simple and avoid the customer's initial confusion.
`If true, any exception while loading key-values would be ignored.`

avanigupta: Thanks for the suggestions. I like Jimmy's suggestion because its important to explicitly state that no key-values will be loaded for this provider. @abhilasharora, do you mind if we use that instead?

abhilasharora: Sure! I would change "Azure App Configuration configuration provider" to "Azure App Configuration provider". Also, it's not clear what provider being empty means. Is there another way to phrase it?

avanigupta: How about this:
"If true, any exceptions that occur while building the App Configuration provider will be suppressed and no key-values will be loaded from App Configuration store."

jimmyca15: I don't think the sentence behaves the same way with that modification. Also, I would avoid the use of "key-values" and "App Configuration store" in favor of "settings" and "App Configuration instance".

This is very different from what I had before but, changing that previous sentence is hard.
"Determines the behavior of the App Configuration provider when an exception occurs. If false, the exception is thrown. If true, the exception is suppressed and no settings are populated from Azure App Configuration."

abhilasharora (approved):

PR #215, opened 2020-11-18
Fix unit test failures

abhilasharora:
This pull request fixes issues with the following unit tests.

1. Tests.TestTurnOffRequestTracing
    This test currently fails when run alongside other tests.

2. KeyVaultReferenceTests.ServerRequestIsMadeWhenDefaultCredentialIsSet
    This test currently returns a different inner exception based on the user identity that invoked the test.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
@@ -12,15 +12,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     internal static class TracingUtils
     {
-        static private HostType? _hostType = null;
-
         public static HostType GetHostType()
         {
-            if (_hostType.HasValue)
-            {
-                return _hostType.Value;
-            }
-
             HostType hostType = HostType.Unidentified;
 
             try
@@ -57,8 +50,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 }
             }
             catch (SecurityException) { }
-
-            _hostType = hostType;
 
             return hostType;
         }
--- a/tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
+++ b/tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
@@ -296,7 +296,6 @@ namespace Tests.AzureAppConfiguration
             });
 
             Assert.NotNull(ex.InnerException);
-            Assert.IsType<AuthenticationFailedException>(ex.InnerException);
         }
 
         [Fact]


tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
```diff
@@ -296,7 +296,6 @@ public void ServerRequestIsMadeWhenDefaultCredentialIsSet()
             });
 
             Assert.NotNull(ex.InnerException);
-            Assert.IsType<AuthenticationFailedException>(ex.InnerException);
```
jimmyca15, line 299: Didn't we have this assertion to verify this line https://github.com/Azure/AppConfiguration-DotnetProvider/blob/c3ed5a3f58096db2bf781724fd07ec15ac3fd9c9/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs#L55 ?

jimmyca15: I do see that we added it in the same PR that we added that line: https://github.com/Azure/AppConfiguration-DotnetProvider/commit/fbde2ecd1a0262675a1c677f322a1a0fec26c1da#diff-1a142c991b5e50180b6e1403412391179ffb37afdccedaaba3567696eeb82efa

jimmyca15: I guess all we can really ensure is that `KeyVaultReferenceException` is the top level exception, so should be fine. Just wanted to make sure that we didn't need this check since I remember we had some discussion about this line specifically in the past.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -12,15 +12,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     internal static class TracingUtils
     {
-        static private HostType? _hostType = null;
-
         public static HostType GetHostType()
         {
-            if (_hostType.HasValue)
```
jimmyca15, line 19: There is some value in having the `_hostType` field cache the result. We don't swallow exceptions on every call from lines 54-59. Can we go with a solution where we tie the _hostType evaluation to the lifetime of the configuration provider that is using it?

tests/Tests.AzureAppConfiguration/KeyVaultReferenceTests.cs
abhilasharora: As long as the actual exception is wrapped in `KeyVaultReferenceException`, we should be good.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
abhilasharora: Yes, that's the current behavior. We retrieve the host type information once and then cache it at the provider level.

jimmyca15: Okay, sounds good.

jimmyca15 (approved):

Issue #163, opened 2020-05-23
Allow selection and sentinel keys on feature flags


RemyBoyer:
Hi everyone,

We are currently building a feature flag infrastructure for all our applications. We want to use AppConfiguration, AppConfiguration-DotnetProvider and FeatureManagement-Dotnet to store and check the feature flags on the applications. We'll build our own feature flag management app though, to provide advanced logic (custom filters, audit, authentication/authorization, more...) to our internal users.

Since we are in a micro-services environment, we would like to reuse a single Azure App Configuration to decrease the costs and, hopefully, the infrastructure complexity.

We would ideally like to use and watch feature flags the same way it's done with "classic" app configuration usage.

For example, with the SDK we can currently do this:

```csharp
                options.Select("App:MyApp:*");

                options.Connect(new Uri(endpoint), credential).ConfigureRefresh(refreshOptions =>
                        {
                            refreshOptions.Register("App:MyAppSentinelKey", true);
                        });
```

Bu this can't be done use feature flags.

The main reasons we would like to do this would be:

- to reduce the number of feature flags that are scanned every 30s (or whatever the cache expiration is set to)
 
Indeed, putting aside the overload that it surely represents on the AppConfiguration instance, it would also require multiple calls for each multiple of 100 feature flags, because of paging. Since our instance will be used by many apps, we expect it to contains many feature flags. If we imagine the app containing 1000 feature flags, this would mean that each app has to make 10 calls every 30s just to check if something has changed. We would break the 20k max requests per hour with just 17 apps.
Of course, this could mitigated by increasing the cache expiration, but we'd loose reactivity, with doesn't seems to be a good tradeoff.

- to avoid downloading feature flags updated data that aren't used by the application.

Indeed, with the currently implemented feature flag configuration, the app watches all the feature flags. This means that when a feature flag used by app A is updated, it will be watched & downloaded by app B, even if it doesn't use it.

For these reasons, we'd like to be able to use explicit `.Select` and `.Register`, which will eliminate those two drawbacks by providing the ability to select only the feature flags that are useful to the app, and by using sentinel keys that will be automatically updated by our Feature Flags management app.

We actually managed to enable this by copying and adapting this method 
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/d19492dd64f379240194eba945468e52f9ec37ff/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs#L138
[...]

PR #221, opened 2020-11-27
Add sentinel keys and filter keys support on feature flags

RemyBoyer:
Hello !

This pull request aims to implement both sentinel and filter keys for feature flags.

The purpose has been explained in this issue: #163 

Developers should be able to opt-in independently to filter keys and refresh keys by calling respectively `Select` and `RegisterRefreshKey`. When enabled, it disables the "global" selection and "global" refresh.

Thank you for your consideration.


RemyBoyer:
@abhilasharora @avanigupta Sorry for tagging you here, but do you think this has a chance to get merged at some point?
Nor the issue nor this PR has had any feedback from the maintainers. This is disappointing as we don't even know if we can rely on this feature being available in the future.

abhilasharora:
> @abhilasharora @avanigupta Sorry for tagging you here, but do you think this has a chance to get merged at some point?
> Nor the issue nor this PR has had any feedback from the maintainers. This is disappointing as we don't even know if we can rely on this feature being available in the future.

Apologies for letting this slip through the cracks. We would love to have users be able to retrieve and refresh a subset of feature flags. I have responded to the original issue referenced in the description.

PR #227, opened 2021-01-06
Maintain nuget version for AspNetCore and base package separately

avanigupta:
Instead of having a common nuget package version for both `Microsoft.Extensions.Configuration.AzureAppConfiguration` and `Microsoft.Azure.AppConfiguration.AspNetCore` packages, we can maintain the version separately for both projects. This would prevent accidentally publishing a new package of AspNetCore without changing the base package version reference.

### DIFF
--- a/src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
+++ b/src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
@@ -27,16 +27,16 @@
 
   <!-- Nuget Package Version Settings -->
 
-  <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'==''">
-    <Version>4.9.9999</Version>
+  <PropertyGroup>
+    <OfficialVersion>4.1.0</OfficialVersion>
   </PropertyGroup>
-
+  
   <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'=='Official'">
-    <Version>4.1.0</Version>
+    <Version>$(OfficialVersion)</Version>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'!='Official'">
-    <Version>4.1.0-$(CDP_PATCH_NUMBER)-$(Revision)</Version>
+    <Version>$(OfficialVersion)-$(CDP_PATCH_NUMBER)-$(Revision)</Version>
   </PropertyGroup>
   
 </Project>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
```diff
@@ -24,4 +24,29 @@
...
+  <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'==''">
+    <Version>4.9.9999</Version>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'=='Official'">
+    <Version>4.1.0</Version>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'!='Official'">
+    <!-- Note that CDP_PACKAGE_VERSION_NUMERIC uses format Major.Minor.MMddyyrrrr, which causes compile error CS7034 because .NET Core doesn't allow version numbers higher than 65534. -->
```
jimmyca15, line 39: Can we keep the <Revision> calculation in the common NugetProperties to avoid defining it in both csproj?

avanigupta: Done

src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
```diff
@@ -24,4 +24,19 @@
   <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
   </ItemGroup>
+
+  <!-- Nuget Package Version Settings -->
+
+  <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'==''">
+    <Version>4.9.9999</Version>
+  </PropertyGroup>
```
abhilasharora, line 32: Since this snippet is common across both files, it could also be shared in `NugetProperties.props` file.

src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
```diff
@@ -24,4 +24,19 @@
...
+
+  <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'=='Official'">
+    <Version>4.1.0</Version>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'!='Official'">
+    <Version>4.1.0-$(CDP_PATCH_NUMBER)-$(Revision)</Version>
+  </PropertyGroup>
```
abhilasharora, line 40: Since part of the version `4.1.0` in this case is expected to be the same for both official and non-official builds, separating it out as a separate property will allow us to have a single source of truth.

jimmyca15 (approved):

abhilasharora (approved):

PR #234, opened 2021-03-08
Support prefix filtering and prefix trimming from feature flags

avanigupta:
In response to customer issue #163, this PR introduces two new APIs for using feature flags.

```cs 
1. FeatureFlagOptions Select(string featureFlagFilter, string labelFilter) 
```
Call this method to selectively load feature flags based on key and label filters. This method can be called multiple times within a `UseFeatureFlags` call. 

Usage example:

```
var configuration = new ConfigurationBuilder()
    {
        options.Connect(connectionString);

        options.UseFeatureFlags(featureFlagOptions =>
        {
            featureFlagOptions.Select("MyFeature", "MyLabel");  // requests a single feature flag whose key is MyFeature and label is MyLabel 
            featureFlagOptions.Select("App1/*", "Label1");  // requests all feature flags whose key starts with App1/ and label is Label1 
            featureFlagOptions.CacheExpirationInterval = cacheExpirationTimeSpan;
        });
    })
    .Build();
```

```cs
2. FeatureFlagOptions TrimFeatureFlagPrefix(string prefix)
```

Trims the provided prefix from the keys of all feature flags retrieved from Azure App Configuration. Follows the same pattern as our existing [TrimKeyPrefix](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.azureappconfiguration.azureappconfigurationoptions.trimkeyprefix?view=azure-dotnet-preview) API.

Usage example:

```
var configuration = new ConfigurationBuilder()
    {
        options.Connect(connectionString);

        // Load feature flags with a specific prefix and label 
        options.UseFeatureFlags(featureFlagOptions =>
        {
            featureFlagOptions.Select("App1/*", "Label1");  // requests all feature flags whose key starts with App1/ and label is Label1 
            featureFlagOptions.Select("*");   // requests all feature flags whose label is null

            featureFlagOptions.TrimFeatureFlagPrefix("App1/");   // Trim "App1/" prefix from all feature flags
            featureFlagOptions.TrimFeatureFlagPrefix("Common/");  // Trim "Common/" prefix from all feature flags
        });
    })
    .Build();
```

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -4,6 +4,7 @@
 using Azure.Core;
 using Azure.Data.AppConfiguration;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Models;
 using System;
@@ -34,6 +35,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private IConfigurationRefresher _refresher = new AzureAppConfigurationRefresher();
 
         private SortedSet<string> _keyPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.InvariantCultureIgnoreCase)));
+        private SortedSet<string> _featureFlagPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.InvariantCultureIgnoreCase)));
 
         /// <summary>
         /// The connection string to use to connect to Azure App Configuration.
@@ -154,24 +156,33 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     string.Format(ErrorMessages.CacheExpirationTimeTooShort, MinimumFeatureFlagsCacheExpirationInterval.TotalMilliseconds));
             }
 
-            if (!(_kvSelectors.Any(selector => selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker) && selector.LabelFilter.Equals(options.Label))))
+            string fullPrefix = FeatureManagementConstants.FeatureFlagMarker + options.Prefix;
+
+            if (!(_kvSelectors.Any(selector => selector.KeyFilter.StartsWith(fullPrefix) && selector.LabelFilter.NormalizeNull() == options.Label.NormalizeNull())))
             {
-                Select(FeatureManagementConstants.FeatureFlagMarker + "*", options.Label);
+                Select(fullPrefix + "*", options.Label);
             }
 
             if (!_adapters.Any(a => a is FeatureManagementKeyValueAdapter))
             {
-                _adapters.Add(new FeatureManagementKeyValueAdapter());
+                _adapters.Add(new FeatureManagementKeyValueAdapter(_featureFlagPrefixes));
             }
 
-            if (!_multiKeyWatchers.Any(kw => kw.Key.Equals(FeatureManagementConstants.FeatureFlagMarker)))
+            var multiKeyWatcher = _multiKeyWatchers.FirstOrDefault(kw => kw.Key.Equals(fullPrefix) && kw.Label.NormalizeNull() == options.Label.NormalizeNull());
+            
+            if (multiKeyWatcher == null)
             {
                 _multiKeyWatchers.Add(new KeyValueWatcher
                 {
-                    Key = FeatureManagementConstants.FeatureFlagMarker,
+                    Key = fullPrefix,
                     Label = options.Label,
                     CacheExpirationInterval = options.CacheExpirationInterval
                 });
+            }
+            else
+            {
+                // If UseFeatureFlags is called multiple times for the same prefix+label, last cache expiration time wins
+                multiKeyWatcher.CacheExpirationInterval = options.CacheExpirationInterval;
             }
 
             return this;
@@ -242,6 +253,21 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             }
 
             _keyPrefixes.Add(prefix);
+            return this;
+        }
+
+        /// <summary>
+        /// Trims the provided prefix from the keys of all feature flags retrieved from Azure App Configuration.
+        /// </summary>
+        /// <param name="prefix">The prefix to be trimmed from all feature flags retrieved from Azure App Configuration.</param>
+        public AzureAppConfigurationOptions TrimFeatureFlagPrefix(string prefix)
+        {
+            if (string.IsNullOrEmpty(prefix))
+            {
+                throw new ArgumentNullException(nameof(prefix));
+            }
+
+            _featureFlagPrefixes.Add(prefix);
             return this;
         }
 
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -34,8 +34,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private List<KeyValueSelector> _kvSelectors = new List<KeyValueSelector>();
         private IConfigurationRefresher _refresher = new AzureAppConfigurationRefresher();
 
-        private SortedSet<string> _keyPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.InvariantCultureIgnoreCase)));
-        private SortedSet<string> _featureFlagPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.InvariantCultureIgnoreCase)));
+        private SortedSet<string> _keyPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.OrdinalIgnoreCase)));
+        private List<string> _featureFlagPrefixes = new List<string>();
 
         /// <summary>
         /// The connection string to use to connect to Azure App Configuration.
@@ -156,33 +156,63 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     string.Format(ErrorMessages.CacheExpirationTimeTooShort, MinimumFeatureFlagsCacheExpirationInterval.TotalMilliseconds));
             }
 
-            string fullPrefix = FeatureManagementConstants.FeatureFlagMarker + options.Prefix;
-
-            if (!(_kvSelectors.Any(selector => selector.KeyFilter.StartsWith(fullPrefix) && selector.LabelFilter.NormalizeNull() == options.Label.NormalizeNull())))
+            if (options.FeatureFlagSelectors.Count() != 0 && options.Label != null)
             {
-                Select(fullPrefix + "*", options.Label);
+                throw new InvalidOperationException($"Please select feature flags by either the {nameof(options.Select)} method or by setting the {nameof(options.Label)}, not both.");
             }
+            
+            if (options.FeatureFlagSelectors.Count() == 0 && options.Label == null)
+            {
+                // No Select clause and the Label property is not set
+                options.FeatureFlagSelectors.Add(new KeyValueSelector
+                {
+                    KeyFilter = FeatureManagementConstants.FeatureFlagMarker + "*",
+                    LabelFilter = LabelFilter.Null
+                });  
+            }
+            else if (options.FeatureFlagSelectors.Count() == 0 && options.Label != null)
+            {
+                // No Select clause but the Label property is set
+                options.FeatureFlagSelectors.Add(new KeyValueSelector
+                {
+                    KeyFilter = FeatureManagementConstants.FeatureFlagMarker + "*",
+                    LabelFilter = options.Label
+                });
+            }
+
+            _featureFlagPrefixes.AddRange(options.FeatureFlagPrefixes);
 
             if (!_adapters.Any(a => a is FeatureManagementKeyValueAdapter))
             {
                 _adapters.Add(new FeatureManagementKeyValueAdapter(_featureFlagPrefixes));
             }
 
-            var multiKeyWatcher = _multiKeyWatchers.FirstOrDefault(kw => kw.Key.Equals(fullPrefix) && kw.Label.NormalizeNull() == options.Label.NormalizeNull());
-            
-            if (multiKeyWatcher == null)
+            foreach (var featureFlagSelector in options.FeatureFlagSelectors)
             {
-                _multiKeyWatchers.Add(new KeyValueWatcher
+                var featureFlagFilter = featureFlagSelector.KeyFilter;
+                var labelFilter = featureFlagSelector.LabelFilter;
+
+                if (!_kvSelectors.Any(selector => selector.KeyFilter == featureFlagFilter && selector.LabelFilter == labelFilter))
                 {
-                    Key = fullPrefix,
-                    Label = options.Label,
-                    CacheExpirationInterval = options.CacheExpirationInterval
-                });
-            }
-            else
-            {
-                // If UseFeatureFlags is called multiple times for the same prefix+label, last cache expiration time wins
-                multiKeyWatcher.CacheExpirationInterval = options.CacheExpirationInterval;
+                    Select(featureFlagFilter, labelFilter);
+                }
+
+                var multiKeyWatcher = _multiKeyWatchers.FirstOrDefault(kw => kw.Key.Equals(featureFlagFilter) && kw.Label.NormalizeNull() == labelFilter.NormalizeNull());
+
+                if (multiKeyWatcher == null)
+                {
+                    _multiKeyWatchers.Add(new KeyValueWatcher
+                    {
+                        Key = featureFlagFilter,
+                        Label = labelFilter,
+                        CacheExpirationInterval = options.CacheExpirationInterval
+                    });
+                }
+                else
+                {
+                    // If UseFeatureFlags is called multiple times for the same key and label filters, last cache expiration time wins
+                    multiKeyWatcher.CacheExpirationInterval = options.CacheExpirationInterval;
+                }
             }
 
             return this;
@@ -253,21 +283,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             }
 
             _keyPrefixes.Add(prefix);
-            return this;
-        }
-
-        /// <summary>
-        /// Trims the provided prefix from the keys of all feature flags retrieved from Azure App Configuration.
-        /// </summary>
-        /// <param name="prefix">The prefix to be trimmed from all feature flags retrieved from Azure App Configuration.</param>
-        public AzureAppConfigurationOptions TrimFeatureFlagPrefix(string prefix)
-        {
-            if (string.IsNullOrEmpty(prefix))
-            {
-                throw new ArgumentNullException(nameof(prefix));
-            }
-
-            _featureFlagPrefixes.Add(prefix);
             return this;
         }
 
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -34,8 +34,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private List<KeyValueSelector> _kvSelectors = new List<KeyValueSelector>();
         private IConfigurationRefresher _refresher = new AzureAppConfigurationRefresher();
 
+        // The following sets are sorted in descending order.
+        // Since multiple prefixes could start with the same characters, we need to trim the longest prefix first.
         private SortedSet<string> _keyPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.OrdinalIgnoreCase)));
-        private List<string> _featureFlagPrefixes = new List<string>();
+        private SortedSet<string> _featureFlagPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.OrdinalIgnoreCase)));
 
         /// <summary>
         /// The connection string to use to connect to Azure App Configuration.
@@ -158,33 +160,17 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             if (options.FeatureFlagSelectors.Count() != 0 && options.Label != null)
             {
-                throw new InvalidOperationException($"Please select feature flags by either the {nameof(options.Select)} method or by setting the {nameof(options.Label)}, not both.");
+                throw new InvalidOperationException($"Please select feature flags by either the {nameof(options.Select)} method or by setting the {nameof(options.Label)} property, not both.");
             }
             
-            if (options.FeatureFlagSelectors.Count() == 0 && options.Label == null)
+            if (options.FeatureFlagSelectors.Count() == 0)
             {
-                // No Select clause and the Label property is not set
+                // Select clause is not present
                 options.FeatureFlagSelectors.Add(new KeyValueSelector
                 {
                     KeyFilter = FeatureManagementConstants.FeatureFlagMarker + "*",
-                    LabelFilter = LabelFilter.Null
+                    LabelFilter = options.Label == null ? LabelFilter.Null : options.Label
                 });  
-            }
-            else if (options.FeatureFlagSelectors.Count() == 0 && options.Label != null)
-            {
-                // No Select clause but the Label property is set
-                options.FeatureFlagSelectors.Add(new KeyValueSelector
-                {
-                    KeyFilter = FeatureManagementConstants.FeatureFlagMarker + "*",
-                    LabelFilter = options.Label
-                });
-            }
-
-            _featureFlagPrefixes.AddRange(options.FeatureFlagPrefixes);
-
-            if (!_adapters.Any(a => a is FeatureManagementKeyValueAdapter))
-            {
-                _adapters.Add(new FeatureManagementKeyValueAdapter(_featureFlagPrefixes));
             }
 
             foreach (var featureFlagSelector in options.FeatureFlagSelectors)
@@ -213,6 +199,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     // If UseFeatureFlags is called multiple times for the same key and label filters, last cache expiration time wins
                     multiKeyWatcher.CacheExpirationInterval = options.CacheExpirationInterval;
                 }
+            }
+
+            options.FeatureFlagPrefixes.ForEach(prefix => _featureFlagPrefixes.Add(prefix));
+
+            if (!_adapters.Any(a => a is FeatureManagementKeyValueAdapter))
+            {
+                _adapters.Add(new FeatureManagementKeyValueAdapter(_featureFlagPrefixes));
             }
 
             return this;
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -14,6 +14,7 @@ using System.Linq;
 using System.Net;
 using System.Security;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -462,10 +463,27 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
                 try
                 {
-                    IEnumerable<ConfigurationSetting> currentKeyValues = _applicationSettings.Values.Where(kv =>
+                    IEnumerable<ConfigurationSetting> currentKeyValues;
+
+                    // Check if the user-provided feature flag filter ends with an unescaped * character:
+                    // (?<!\\)    Matches if the preceding character is not a backslash
+                    // (?:\\\\)*  Matches any number of occurrences of two backslashes
+                    // \*$        Matches if * is the last character
+                    if (Regex.IsMatch(changeWatcher.Key, @"(?<!\\)(?:\\\\)*\*$"))
                     {
-                        return kv.Key.StartsWith(changeWatcher.Key) && kv.Label == changeWatcher.Label.NormalizeNull();
-                    });
+                        // Get current application settings starting with changeWatcher.Key, excluding the last * character
+                        currentKeyValues = _applicationSettings.Values.Where(kv =>
+                        {
+                            return kv.Key.StartsWith(changeWatcher.Key.Substring(0, changeWatcher.Key.Length - 1)) && kv.Label == changeWatcher.Label.NormalizeNull();
+                        });
+                    }
+                    else
+                    {
+                        currentKeyValues = _applicationSettings.Values.Where(kv =>
+                        {
+                            return kv.Key.Equals(changeWatcher.Key) && kv.Label == changeWatcher.Label.NormalizeNull();
+                        });
+                    }
 
                     IEnumerable<KeyValueChange> keyValueChanges = await _client.GetKeyValueChangeCollection(currentKeyValues, new GetKeyValueChangeCollectionOptions
                     {
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -14,7 +14,6 @@ using System.Linq;
 using System.Net;
 using System.Security;
 using System.Text.Json;
-using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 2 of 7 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -245,6 +256,21 @@ public AzureAppConfigurationOptions TrimKeyPrefix(string prefix)
             return this;
         }
 
+        /// <summary>
+        /// Trims the provided prefix from the keys of all feature flags retrieved from Azure App Configuration.
+        /// </summary>
+        /// <param name="prefix">The prefix to be trimmed from all feature flags retrieved from Azure App Configuration.</param>
```
abhilasharora, line 262: >The prefix to be trimmed from all feature flags retrieved from Azure App Configuration. [](start = 33, length = 87)

This could be shortened to `The prefix to be trimmed.` to avoid repeating the details in the summary comment.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -245,6 +256,21 @@ public AzureAppConfigurationOptions TrimKeyPrefix(string prefix)
...
+        public AzureAppConfigurationOptions TrimFeatureFlagPrefix(string prefix)
```
abhilasharora, line 263: >TrimFeatureFlagPrefix [](start = 44, length = 21)

What happens when a user configures both `TrimKeyPrefix` and `TrimFeatureFlagPrefix`? It looks like the `TrimKeyPrefix` would apply to the trimmed feature flags too.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,25 +156,34 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
                     string.Format(ErrorMessages.CacheExpirationTimeTooShort, MinimumFeatureFlagsCacheExpirationInterval.TotalMilliseconds));
             }
 
-            if (!(_kvSelectors.Any(selector => selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker) && selector.LabelFilter.Equals(options.Label))))
+            string fullPrefix = FeatureManagementConstants.FeatureFlagMarker + options.Prefix;
+
+            if (!(_kvSelectors.Any(selector => selector.KeyFilter.StartsWith(fullPrefix) && selector.LabelFilter.NormalizeNull() == options.Label.NormalizeNull())))
```
abhilasharora, line 161: >NormalizeNull [](start = 113, length = 13)

Do you see a scenario where `selector.LabelFilter` could be `null`?

avanigupta: `selector.LabelFilter` should always be `LabelFilter.Null` but `options.Label` could be either `null` or `LabelFilter.Null`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -19,5 +19,11 @@ public class FeatureFlagOptions
         /// The time after which the cached values of the feature flags expire.  Must be greater than or equal to 1 second.
         /// </summary>
         public TimeSpan CacheExpirationInterval { get; set; } = AzureAppConfigurationOptions.DefaultFeatureFlagsCacheExpirationInterval;
+
+        /// <summary>
+        /// Selectively load only those feature flags which begin with this prefix.
+        /// The characters asterisk (*), comma (,) and backslash (\) are reserved and must be escaped using a backslash (\).
+        /// </summary>
+        public string Prefix { get; set; } = string.Empty;
```
abhilasharora, line 27: >Prefix [](start = 22, length = 6)

If my feature flag shows up as `.appconfig.featureflag/UI/EnableFeature` in the Azure Portal, it is not clear from the comment if the prefix expected here to get the key name as `EnableFeature`  is `.appconfig.featureflag/UI/` or `UI/`. It might help to clarify that it is the latter by including an example in the summary comments.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
abhilasharora: Right! Could we get rid of the `NormalizeNull` call on `selector.LabelFilter`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -19,5 +19,11 @@ public class FeatureFlagOptions
         /// The time after which the cached values of the feature flags expire.  Must be greater than or equal to 1 second.
         /// </summary>
         public TimeSpan CacheExpirationInterval { get; set; } = AzureAppConfigurationOptions.DefaultFeatureFlagsCacheExpirationInterval;
+
+        /// <summary>
+        /// Selectively load only those feature flags which begin with this prefix.
+        /// The characters asterisk (*), comma (,) and backslash (\) are reserved and must be escaped using a backslash (\).
```
abhilasharora, line 25: >The characters asterisk (*), comma (,) and backslash (\) are reserved and must be escaped using a backslash (\). [](start = 12, length = 112)

There isn't anything preventing the user from using these reserved characters unescaped in the prefix. We could add validation for this case and return an error message.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -34,6 +35,7 @@ public class AzureAppConfigurationOptions
         private IConfigurationRefresher _refresher = new AzureAppConfigurationRefresher();
 
         private SortedSet<string> _keyPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.InvariantCultureIgnoreCase)));
+        private SortedSet<string> _featureFlagPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.InvariantCultureIgnoreCase)));
```
abhilasharora, line 38: We could use `OrdinalIgnoreCase` for string comparison in both these statements to maintain consistency with the rest of the code base.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
abhilasharora, line 263: >TrimFeatureFlagPrefix [](start = 44, length = 21)

Do we need this method assuming that the existing `TrimKeyPrefix` method works for feature flags too?

avanigupta: `TrimKeyPrefix` comes into play after all adapters have been processed on a key-value, whereas `TrimFeatureFlagPrefix` would try to trim the prefix from feature flag IDs while feature management adapter is processing the key-value.
For example, if I have these key-values and feature flags:
- test_keyvalue1
- .appconfig.featureflag/test_feature1 

`TrimKeyPrefix("test_")` would trim the prefix from the key-value, not the feature flag, because the ***processed*** feature flag key doesn't start with the prefix. Your config would have two keys: `keyvalue1` and `FeatureManagement:test_feature1`.

On the other hand, `TrimFeatureFlagPrefix("test_")` would trim the prefix from feature flag ID. That's how you could get `FeatureManagement:feature1` in your config.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: I thought that we would end up with duplicate selectors if we don't `NormalizeNull` here. But after looking into `Select` logic more closely, I think we can remove the `NormalizeNull` call for both `selector.LabelFilter` and `options.Label`. `Select` method will take care of `null` and `LabelFilter.Null` appropriately.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
avanigupta: This validation could be done using regex to to error out if unescaped reserved characters are present in the prefix. However, if these characters are present in key name, we get 400 bad request from server. So we already have server side validation in place. I am not sure if we should introduce the same logic in config provider considering these reserved characters could change in future. We have similar requirement for the key filter input of `Select`, but we don't do any client side validation there either.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: Same as [this comment](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/234#discussion_r589874260)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,24 +156,63 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
                     string.Format(ErrorMessages.CacheExpirationTimeTooShort, MinimumFeatureFlagsCacheExpirationInterval.TotalMilliseconds));
             }
 
-            if (!(_kvSelectors.Any(selector => selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker) && selector.LabelFilter.Equals(options.Label))))
+            if (options.FeatureFlagSelectors.Count() != 0 && options.Label != null)
             {
-                Select(FeatureManagementConstants.FeatureFlagMarker + "*", options.Label);
+                throw new InvalidOperationException($"Please select feature flags by either the {nameof(options.Select)} method or by setting the {nameof(options.Label)}, not both.");
```
jimmyca15, line 161: >the {nameof(options.Label)} [](start = 142, length = 27)

Probably should say "the {nameof(options.Label)} **property**"

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,24 +156,63 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
+            {
+                // No Select clause but the Label property is set
+                options.FeatureFlagSelectors.Add(new KeyValueSelector
+                {
+                    KeyFilter = FeatureManagementConstants.FeatureFlagMarker + "*",
+                    LabelFilter = options.Label
+                });
             }
 
+            _featureFlagPrefixes.AddRange(options.FeatureFlagPrefixes);
```
jimmyca15, line 183: >_featureFlagPrefixes.AddRange(options.FeatureFlagPrefixes); [](start = 12, length = 59)

Are we preventing duplicates?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,24 +156,63 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
+            }
+            else if (options.FeatureFlagSelectors.Count() == 0 && options.Label != null)
+            {
+                // No Select clause but the Label property is set
+                options.FeatureFlagSelectors.Add(new KeyValueSelector
+                {
+                    KeyFilter = FeatureManagementConstants.FeatureFlagMarker + "*",
+                    LabelFilter = options.Label
+                });
             }
```
jimmyca15, line 181: Looks like

```
            if (options.FeatureFlagSelectors.Count() == 0)
            {
                // No Select clause and the Label property is not set
                options.FeatureFlagSelectors.Add(new KeyValueSelector
                {
                    KeyFilter = FeatureManagementConstants.FeatureFlagMarker + "*",
                    LabelFilter = options.Label ==  null ? LabelFilter.Null : options.Label
                });  
            }
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,24 +156,63 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
 
+            _featureFlagPrefixes.AddRange(options.FeatureFlagPrefixes);
+
             if (!_adapters.Any(a => a is FeatureManagementKeyValueAdapter))
             {
-                _adapters.Add(new FeatureManagementKeyValueAdapter());
+                _adapters.Add(new FeatureManagementKeyValueAdapter(_featureFlagPrefixes));
             }
```
jimmyca15, line 189: Right now this adapter logic is between the two code blocks dealing with selectors, can we move it either before or after.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,24 +156,63 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
+                    Select(featureFlagFilter, labelFilter);
+                }
+
+                var multiKeyWatcher = _multiKeyWatchers.FirstOrDefault(kw => kw.Key.Equals(featureFlagFilter) && kw.Label.NormalizeNull() == labelFilter.NormalizeNull());
+
+                if (multiKeyWatcher == null)
+                {
+                    _multiKeyWatchers.Add(new KeyValueWatcher
+                    {
+                        Key = featureFlagFilter,
```
jimmyca15, line 206: >featureFlagFilter [](start = 30, length = 17)

Is it problematic that this can end with * now?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -462,10 +463,27 @@ private async Task RefreshKeyValueCollections()
...
+                    if (Regex.IsMatch(changeWatcher.Key, @"(?<!\\)(?:\\\\)*\*$"))
                     {
-                        return kv.Key.StartsWith(changeWatcher.Key) && kv.Label == changeWatcher.Label.NormalizeNull();
-                    });
+                        // Get current application settings starting with changeWatcher.Key, excluding the last * character
+                        currentKeyValues = _applicationSettings.Values.Where(kv =>
+                        {
+                            return kv.Key.StartsWith(changeWatcher.Key.Substring(0, changeWatcher.Key.Length - 1)) && kv.Label == changeWatcher.Label.NormalizeNull();
+                        });
+                    }
```
jimmyca15, line 479: Escaped `*` is a complex problem to solve. We would need to implement the same search logic and unescape logic as the backend. I'm thinking we may want to throw an argument exception if someone passes a filter that ends in `\*`.

jimmyca15: We won't support that scenario, but that scenario means somebody would want to select a single feature flag that ends with '*'. Seems rare.

---
In reply to: [597895109](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/234#discussion_r597895109) [](ancestors = 597895109)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: Yes, the underlying data structure is a SortedSet.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,24 +156,63 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
                     string.Format(ErrorMessages.CacheExpirationTimeTooShort, MinimumFeatureFlagsCacheExpirationInterval.TotalMilliseconds));
             }
 
-            if (!(_kvSelectors.Any(selector => selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker) && selector.LabelFilter.Equals(options.Label))))
+            if (options.FeatureFlagSelectors.Count() != 0 && options.Label != null)
             {
```
jimmyca15, line 160: The previous approach doesn't pull any additional feature flags if the user selected feature flags already. Is it intentional to shift away from that approach?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
jimmyca15: I see this line
`        private List<string> _featureFlagPrefixes = new List<string>();`
Where's the sorted set?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: We officially don't support * in feature flag names, but it is still possible to create a feature flag named MyApp* using our REST API. So if someone wants to register this feature flag, their only option is to provide "MyApp\*" as the key filter.

jimmyca15: I get that, but it seems so rare I don't think it's worth complicating our client to support it.

jimmyca15: If someone asks for it we can always add it. But at this moment, no one is even asking for singular select of a feature flag.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: The logic to process user input and add to a list of feature flag prefixes is in `FeatureFlagOptions` [here](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/234/files#diff-6481f817620bb03ea907e90b07f2bdea0493301347d8adb2c6b3af4f6ac45c4dR16)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,24 +156,63 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
                 {
-                    Key = FeatureManagementConstants.FeatureFlagMarker,
-                    Label = options.Label,
-                    CacheExpirationInterval = options.CacheExpirationInterval
-                });
+                    Select(featureFlagFilter, labelFilter);
+                }
+
+                var multiKeyWatcher = _multiKeyWatchers.FirstOrDefault(kw => kw.Key.Equals(featureFlagFilter) && kw.Label.NormalizeNull() == labelFilter.NormalizeNull());
+
```
jimmyca15, line 201: We can optimize so that if the flags are already watched we don't add an additional watcher.

For example,

if .appconfig/featureflag/* is watched, we won't add .appconfig/featureflag/abc*

The interesting part is the cache expiration interval could be different.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Consider "MyApp\\*" key filter. In this case, the `\` character is escaped. So the user is requesting all feature flags that begin with "MyApp\". Should this be a valid request?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: We will still check that before registering new kv selectors [here](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/234/files/e4b6e329af81c52ec6f0f9195f39a6ddbad3dbcf#diff-9fe8bf4d21db7a9a52895366df1e58edce9f4d7dca80514f7b457311f3218cecR195). That behavior has not changed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: It's a valid request. I don't think we should support it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
```diff
@@ -105,7 +95,7 @@ public static async Task<IEnumerable<KeyValueChange>> GetKeyValueChangeCollectio
             var hasKeyValueCollectionChanged = false;
             var selector = new SettingSelector
             {
-                KeyFilter = options.Prefix + "*",
+                KeyFilter = options.Prefix,
```
jimmyca15, line 98: >Prefix [](start = 36, length = 6)

I guess this needs to be renamed since it's not a prefix anymore.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
jimmyca15: I found it after moving files. But it looks like the duplication is still possible through multiple usefeatureflags invocations.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -10,14 +13,90 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManage
     /// </summary>
     public class FeatureFlagOptions
     {
+        private SortedSet<string> _featureFlagPrefixes = new SortedSet<string>(Comparer<string>.Create((k1, k2) => -string.Compare(k1, k2, StringComparison.OrdinalIgnoreCase)));
```
jimmyca15, line 16: >- [](start = 115, length = 1)

What's the `-` for here? and is it the reason that [https://docs.microsoft.com/en-us/dotnet/api/system.stringcomparer.ordinalignorecase?view=net-5.0](https://docs.microsoft.com/en-us/dotnet/api/system.stringcomparer.ordinalignorecase?view=net-5.0) isn't being used?

avanigupta: `-` is for sorting in descending order. I'm not sure what you mean by the second question.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: You're right. `AzureAppConfigurationOptions` should have the SortedSet in that case.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
jimmyca15: Why do we want to sort in descending order again?

`StringComparer.OrdinalIgnoreCase` is a built in string comparer so we don't have to have our own lambda.

avanigupta: Consider two prefixes registered for trimming: "ABC" and "ABCD"
There might be some feature called "ABCDEF" and according to the registered prefixes, users would expects "EF" in their config. If the set is sorted in ascending order, we would just trim "ABC" and add "DEF" key to their config.

This means that we should try to match and trim the longest prefix first.

jimmyca15: I see. That's tricky, can you put a comment.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
jimmyca15: Probably will have to get on a call about this one. I may need some help :)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
zhenlan: Wow, 😄

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: As discussed offline, this optimization is not worth the effort right now. It's a rare situation that should be optimized on customer side by not registering both `.appconfig/featureflag/*` and `.appconfig/featureflag/abc*` filters.

drago-draganov:
Trim for feature flag is a bit confusing. Feature flags already have system required prefix.
Also Trim was introduced when the KV object model wasn't publicly available. Now we plan to add Map function and let customer code transform that way they see it fit. I don't think we should add any fixed transformations going forward.

avanigupta:
> 
> 
> Trim for feature flag is a bit confusing. Feature flags already have system required prefix.
> Also Trim was introduced when the KV object model wasn't publicly available. Now we plan to add Map function and let customer code transform that way they see it fit. I don't think we should add any fixed transformations going forward.

I think both solutions try to solve different problems. The `Map` API would let users transform their `ConfigurationSetting` object and can definitely be used to trim feature flag prefixes. But we also want to make sure that a `ConfigurationSetting` is processed by only one adapter - either our internal adapter, or user defined adapter in `Map` method. So if someone wants to trim a prefix from feature flags using `Map` API, they will have to let go of the built-in flattening of feature flags that we provide with our internal Feature Flag adapter.

For example, if I have this feature flag in AppConfig:

Key = `.appconfig.featureflag/test_feature1`
Value = ` {"id":"test_feature1","description":"","enabled":true,"conditions":{"client_filters":[{"name":"Microsoft.Targeting","parameters":{"Audience":{"Users":[],"Groups":[],"DefaultRolloutPercentage":50}}}]}}`

`TrimFeatureFlagPrefix("test_")` would trim the prefix from the feature flag name while preserving the rest of the structure of feature flags:
```
FeatureManagement:feature1 =
FeatureManagement:feature1:EnabledFor =
FeatureManagement:feature1:EnabledFor:0 =
FeatureManagement:feature1:EnabledFor:0:Parameters =
FeatureManagement:feature1:EnabledFor:0:Parameters:Audience =
FeatureManagement:feature1:EnabledFor:0:Parameters:Audience:DefaultRolloutPercentage = 50
FeatureManagement:feature1:EnabledFor:0:Name = Microsoft.Targeting
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -14,6 +14,7 @@
 using System.Net;
 using System.Security;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 using System.Threading;
```
jimmyca15, line 18: Not needed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -462,14 +463,27 @@ private async Task RefreshKeyValueCollections()
...
+                    IEnumerable<ConfigurationSetting> currentKeyValues;
+
+                    if (changeWatcher.Key.EndsWith("*"))
                     {
-                        return kv.Key.StartsWith(changeWatcher.Key) && kv.Label == changeWatcher.Label.NormalizeNull();
-                    });
+                        // Get current application settings starting with changeWatcher.Key, excluding the last * character
+                        currentKeyValues = _applicationSettings.Values.Where(kv =>
+                        {
+                            return kv.Key.StartsWith(changeWatcher.Key.Substring(0, changeWatcher.Key.Length - 1)) && kv.Label == changeWatcher.Label.NormalizeNull();
```
jimmyca15, line 473: >changeWatcher.Key.Substring(0, changeWatcher.Key.Length - 1) [](start = 53, length = 60)

This substring is being calculated (string allocation) for every item in `_applicationSettings`. Can you put it in a variable and calculate it once.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,24 +158,54 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
                     string.Format(ErrorMessages.CacheExpirationTimeTooShort, MinimumFeatureFlagsCacheExpirationInterval.TotalMilliseconds));
             }
 
-            if (!(_kvSelectors.Any(selector => selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker) && selector.LabelFilter.Equals(options.Label))))
+            if (options.FeatureFlagSelectors.Count() != 0 && options.Label != null)
             {
-                Select(FeatureManagementConstants.FeatureFlagMarker + "*", options.Label);
+                throw new InvalidOperationException($"Please select feature flags by either the {nameof(options.Select)} method or by setting the {nameof(options.Label)} property, not both.");
```
jimmyca15, line 163: >InvalidOperationException [](start = 26, length = 25)

Can you do an ArgumentException instead? It's validation of the argument that's causing the exception.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -4,6 +4,7 @@
 using Azure.Core;
 using Azure.Data.AppConfiguration;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement;
```
jimmyca15, line 8: What is this using added for?

avanigupta: It's for using `NormalizeNull()` to compare labels [here](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/234/files/49994711cbad21f0ddfa36c5e44256655f4dfec4#diff-9fe8bf4d21db7a9a52895366df1e58edce9f4d7dca80514f7b457311f3218cecR186).

jimmyca15: @avanigupta I see, thanks!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -25,8 +32,17 @@ internal class FeatureManagementKeyValueAdapter : IKeyValueAdapter
                 throw new FormatException(setting.Key, e);
             }
 
+            foreach (string prefix in _featureFlagPrefixes)
+            {
+                if (featureFlag.Id.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    featureFlag.Id = featureFlag.Id.Substring(prefix.Length);
```
jimmyca15, line 39: If featureflag.Id gets set to empty string, that doesn't break anything does it?

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
avanigupta: Not for feature flags, since we append the `featureflag.Id` to `"FeatureManagement:"` string. 
But this made notice that the same logic for trimming prefix from key-values. If the key is equal to the prefix, users' config will contain a key that is an empty string.

jimmyca15: Does it cause a problem?

avanigupta: No, it silently adds the empty key and its corresponding value to `ConfigurationProvider`. Since the key is actually `string.Empty` and not `null`, we don't get any unhandled exception.

jimmyca15: Okay, thanks! Should be fine then.

avanigupta:
> 
> 
> > Trim for feature flag is a bit confusing. Feature flags already have system required prefix.
> > Also Trim was introduced when the KV object model wasn't publicly available. Now we plan to add Map function and let customer code transform that way they see it fit. I don't think we should add any fixed transformations going forward.
> 


Discussed offline with @drago-draganov .
The `TrimFeatureFlagPrefix` API can be introduced for convenience and symmetry with `TrimKeyPrefix` API. In future, we can internally translate it to behave like a special case of the upcoming `Map` API.

Issue #142, opened 2020-03-19
Allow configurable pull refresh cache expiration separate for Keyvault references (secrets)


drago-draganov:
Configuration data and secrets often evolve with different velocities. Shorter cache expiration makes sense for feature management, sentinel keys monitoring, etc.. Unfortunately, that may put stress on Keyvault, considering the fact that each secret must be fetched one at a time. That is also amplified when multiple provider instances are active.

AzureAppConfigurationKeyVaultOptions is perhaps a good place to setup cache expiration. If not specified, default to the existing cache expiration.

abhilasharora:
`AzureAppConfigurationKeyVaultOptions` is meant for properties that used to configure the key vault client. It is decoupled from the key-vault reference settings that the cache expiration will be applied to. It might be more logical to set the key vault reference cache expiration on `AzureAppConfigurationRefreshOptions` since it is a refresh-related property and the settings that it applies to are part of the same object.

drago-draganov:
I don't necessary agree with:
> AzureAppConfigurationKeyVaultOptions is meant for properties that are used to configure the key vault client

As the name suggests it's meant to provide configuration for Keyvault integration with AppConfiguration. That certainly includes Keyvault client setup, but it doesn't lock or prohibit additional options.

AzureAppConfigurationRefreshOptions knows noting anything about Keyvault.

abhilasharora:
Sure, the refresh cache expiration time for key vault reference seems related to both `AzureAppConfigurationKeyVaultOptions` and `AzureAppConfigurationRefreshOptions`.

From the developer perspective, the code would get really confusing. Here's an example.

```` csharp
options.Connect(configuration["connection_string"])
       .ConfigureRefresh(refresh =>
       {
           refresh.Register("MyKeyVaultSecret")
                   .SetCacheExpiration(TimeSpan.FromSeconds(10));
       })
       .ConfigureKeyVault(kv => kv.SetCacheExpiration(TimeSpan.FromMinutes(5)));
````

zhenlan:
Setting expiration time using `ConfigureKeyVault` API follows the same pattern of `UseFeatureFlags`, although, under the cover, all refresh logic is handled by `AzureAppConfigurationRefreshOptions`.

drago-draganov:
I don't find it confusing at all. KeyVault related stuff is properly encapsulated and configured under ConfigureKetVault API. It's easy do discover.

abhilasharora:
[...]

PR #249, opened 2021-04-01
Auto-Reload secrets from Key Vault

### LINKED ISSUE #248: Detect change in KeyVault references
I have an Azure Function, where we use Azure App Configuration to pull a certificate from our KeyVault. The AppConfiguration to KeyVault reference has been setup. The AppConfiguration registration looks something like this -
```
aadConfigBuilder.AddAzureAppConfiguration(options =>
            {
                options.Connect(Environment.GetEnvironmentVariable("AppConfiguration:ConnectionString"))
                    .ConfigureKeyVault(kv =>
                       {
                           kv.SetCredential(new DefaultAzureCredential());
                       })
                    .Select("ExternalSystemAuthentication:AadConfig:*")
                       .ConfigureRefresh(refreshOptions =>
                            refreshOptions
                            /* .Register("ExternalSystemAuthentication:AadConfig:AuthCert", refreshAll: true) */ // want to detect when this changes
                            .Register("ExternalSystemAuthentication:AadConfig:Sentinel", refreshAll: true) // this works as expected
                            .SetCacheExpiration(TimeSpan.FromMinutes(2)));
                configurationRefresher = options.GetRefresher();
            });
```
Everything works as expected when I use a value based 'Sentinel' key to detect a change in config. However, in my case I have auto-rotation setup on the KeyVault certificate. When a new version is available, I want the latest version to be pulled into my Function. However, since the KeyVault reference is just a URL its value does not change in the AppConfiguration, and this key cannot be used as the observable key. So I am forced to manually update the 'Sentinel' so that my function can refresh config. Is there a way App Configuration can detect a change in the KV secret value? Or is there a way I can force refresh of the App Configuration periodically without checking for a change in any key?
Thanks.

avanigupta:
## Overview

This PR introduces two new APIs which allow users to opt-in for periodically reloading secrets and certificates from Key Vault:

### Set refresh interval for individual keys of Key Vault references in App Config:

```cs
AzureAppConfigurationKeyVaultOptions SetSecretRefreshInterval(string secretReferenceKey, TimeSpan refreshInterval)
```

This method allows users to set a refresh interval per key of Key Vault references. The API can be called multiple times to register multiple keys of Key Vault references for refresh. 

### Set refresh interval for all Key Vault references in App Config:

```cs
AzureAppConfigurationKeyVaultOptions SetSecretRefreshInterval(TimeSpan refreshInterval)
```

This method allows users to set a refresh interval for all Key Vault references which do not have individual refresh intervals.

## Usage example:

If your App Configuration has the following Key Vault references:

Key | Value
--- |---
KeyVaultRef1 |  {"uri":"https://abc.vault.azure.net/secrets/Secret1"}
KeyVaultRef2 |  {"uri":"https://xyz.vault.azure.net/secrets/Secret2"}
KeyVaultRef3 |  {"uri":"https://abc.vault.azure.net/secrets/Secret3"}
KeyVaultRef4 |  {"uri":"https://abc.vault.azure.net/secrets/Secret4"}

You can set different refresh intervals for any/all of your secret references:

```cs
var builder = new ConfigurationBuilder();
{
  options.Connect(connectionString)
         .Select("KeyVaultRef*")
         .ConfigureRefresh(refreshOptions =>
          {
            refreshOptions.Register("Sentinel", true);
            refreshOptions.SetCacheExpiration(TimeSpan.FromMinutes(1));
          })
         .ConfigureKeyVault(kvOptions =>
		{
		    kvOptions.SetSecretRefreshInterval("KeyVaultRef1", TimeSpan.FromHours(24)); // Secret1 will be reloaded from Key Vault every 24 hours

			kvOptions.SetSecretRefreshInterval("KeyVaultRef2", TimeSpan.FromHours(12)); // Secret2 will be reloaded from Key Vault every 12 hours

			kvOptions.SetSecretRefreshInterval(TimeSpan.FromDays(2)); // Secret3 and Secret4 will be reloaded from Key Vault every 2 days
		});
});

```

## Notes:

- `SetSecretRefreshInterval` does not monitor the value of a key in App Config. For monitoring the value in AppConfig, use the `ConfigureRefresh` API to register keys for refresh.
- The frequency of reloading secrets from Key Vault should be chosen appropriately based on your needs. If the refresh interval is too low, there is a risk of being throttled by Key Vault.
- If `SetSecretRefreshInterval` is not invoked, secrets will not be automatically reloaded from Key Vault. In this case, secrets would be reloaded only if any of the following occur:
	- App restarts;
	- Sentinel key registered for refresh changes in App Config and triggeres a refresh of entire configuration; or 
	- A Key Vault reference is registered for refresh using `ConfigureRefresh` API, and the reference value changes in App Config.
- Any refresh operation triggered using `IConfigurationRefresher` will not update the value for a Key Vault secret until the cached value for that secret has expired.

Fix #248 and #142.


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -31,15 +38,19 @@ public AzureKeyVaultSecretProvider(TokenCredential credential = null, IEnumerabl
...
                 }
             }
-        }
 
-        public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancellationToken)
-        {
-            if (secretUri == null)
+            if (certificateClients != null)
             {
-                throw new ArgumentNullException(nameof(secretUri));
```
avanigupta, line 40: **Note:** I removed the `secretUri` null check here because this exception will never be thrown since `secretUri` has already been validated before calling `GetSecretValue`.
Additionally, this class should only throw exceptions which can be caught and converted to `KeyVaultReferenceException` by `AzureKeyVaultKeyValueAdapter`.
`ArgumentNullException` will bubble up to the application even if `optional` is set to true in `AddAzureAppConfiguration` method.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -52,5 +56,24 @@ public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTas
...
+        /// Sets the refresh interval for periodically reloading a secret from Key Vault. Refresh interval must be greater than 1 hour. Default refresh interval is 12 hours.
+        /// Any refresh operation triggered using <see cref="IConfigurationRefresher"/> will not update the value for a Key Vault secret until the cached value for that secret has expired.
+        /// </summary>
+        /// <param name="key">Key of the Key Vault reference in Azure App Configuration.</param>
+        /// <param name="refreshInterval">Minimum time that must elapse before the secret is reloaded from Key Vault.</param>
+        public AzureAppConfigurationKeyVaultOptions SetSecretRefreshInterval(string key, TimeSpan? refreshInterval = null)
+        {
+            if (refreshInterval != null && refreshInterval < MinimumRefreshInterval)
+            {
+                throw new ArgumentOutOfRangeException(nameof(refreshInterval), refreshInterval?.TotalHours,
```
jimmyca15, line 70: We already checked refreshInterval for null.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -52,5 +56,24 @@ public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTas
...
+
+        /// <summary>
+        /// Sets the refresh interval for periodically reloading a secret from Key Vault. Refresh interval must be greater than 1 hour. Default refresh interval is 12 hours.
+        /// Any refresh operation triggered using <see cref="IConfigurationRefresher"/> will not update the value for a Key Vault secret until the cached value for that secret has expired.
+        /// </summary>
+        /// <param name="key">Key of the Key Vault reference in Azure App Configuration.</param>
+        /// <param name="refreshInterval">Minimum time that must elapse before the secret is reloaded from Key Vault.</param>
+        public AzureAppConfigurationKeyVaultOptions SetSecretRefreshInterval(string key, TimeSpan? refreshInterval = null)
+        {
+            if (refreshInterval != null && refreshInterval < MinimumRefreshInterval)
```
jimmyca15, line 68: Do you mind using `refreshInterval.HasValue` and `refreshInterval.Value`. Seeing a nullable type used in this type of comparison always throws me off.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -450,6 +468,17 @@ private async Task RefreshIndividualKeyValues()
             }
         }
 
+        private async Task RefreshKeyVaultSecrets()
```
jimmyca15, line 471: >RefreshKeyVaultSecrets [](start = 27, length = 22)

The method isn't specific to secrets.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -450,6 +468,17 @@ private async Task RefreshIndividualKeyValues()
             }
         }
 
+        private async Task RefreshKeyVaultSecrets()
+        {
+            foreach (IKeyValueAdapter adapter in _options.Adapters)
+            {
+                if (adapter.NeedsRefresh())
+                {
+                    await SetData(_applicationSettings).ConfigureAwait(false);
```
jimmyca15, line 477: >SetData [](start = 26, length = 7)

Should we break after that?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -63,6 +82,38 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
             return secretValue;
         }
 
+        internal bool AnyExpiredSecrets()
```
jimmyca15, line 85: >internal [](start = 8, length = 8)

It's an internal class. If the methods are meant to be called outside of the class it's easier to understand if they're public.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -63,6 +82,38 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
+            secretsToBeRemovedFromCache.ForEach(secret => RemoveExpiredSecretFromCache(secret.Key));
+            return shouldRefreshKeyVaultSecrets;
+        }
+
+        internal void RemoveAllSecretsFromCache()
+        {
+            _cachedKeyVaultSecrets.Clear();
+        }
+
+        internal void RemoveExpiredSecretFromCache(string key)
```
jimmyca15, line 112: >RemoveExpiredSecretFromCache [](start = 22, length = 28)

That is a bit specific of a name. Although we are using it for expired secrets it's really just removing something from the cache. Can be `SetDirty(string key)`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -63,6 +82,38 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
+                // Remove the cached Key Vault secret for this key
+                secretsToBeRemovedFromCache.Add(new CachedKeyVaultSecret(cachedSecret.Key));
+                shouldRefreshKeyVaultSecrets = true;
+            }
+
+            secretsToBeRemovedFromCache.ForEach(secret => RemoveExpiredSecretFromCache(secret.Key));
+            return shouldRefreshKeyVaultSecrets;
+        }
+
+        internal void RemoveAllSecretsFromCache()
```
jimmyca15, line 107: >RemoveAllSecretsFromCache [](start = 22, length = 25)

How about `ClearCache` or `SetDirty` with no param?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -16,12 +16,15 @@ internal class AzureKeyVaultSecretProvider
         private readonly IDictionary<string, SecretClient> _secretClients;
         private readonly TokenCredential _credential;
         private readonly Func<Uri, ValueTask<string>> _secretResolver;
+        private readonly Dictionary<string, TimeSpan> _secretRefreshIntervals = new Dictionary<string, TimeSpan>();
+        private HashSet<CachedKeyVaultSecret> _cachedKeyVaultSecrets = new HashSet<CachedKeyVaultSecret>();
```
jimmyca15, line 20: >_cachedKeyVaultSecrets [](start = 46, length = 22)

Is this class used in any multi-threading?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -33,13 +36,8 @@ public AzureKeyVaultSecretProvider(TokenCredential credential = null, IEnumerabl
             }
         }
 
-        public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancellationToken)
+        public async Task<string> GetSecretValue(Uri secretUri, string key, CancellationToken cancellationToken)
```
jimmyca15, line 39: >string key [](start = 64, length = 10)

So if we rotate the underlying secret for a given key, other keys that use the same underlying secret won't get updated if they don't specify `SetSecretRefreshInterval`? It might make sense, but I expected the secret to get rotated for any keys that used it. It may be worth seeing what others expected.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -63,6 +82,38 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
+                {
+                    continue;
+                }
+
+                // Remove the cached Key Vault secret for this key
+                secretsToBeRemovedFromCache.Add(new CachedKeyVaultSecret(cachedSecret.Key));
+                shouldRefreshKeyVaultSecrets = true;
+            }
+
+            secretsToBeRemovedFromCache.ForEach(secret => RemoveExpiredSecretFromCache(secret.Key));
```
jimmyca15, line 103: >RemoveExpiredSecretFromCache [](start = 58, length = 28)

It's a bit hard to follow a query method that also modifies state. I'd suggest separating it somehow. I don't think we really need to remove expired secrets from the cache until we query for a secret again.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -52,5 +56,24 @@ public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTas
...
             return this;
         }
+
+        /// <summary>
+        /// Sets the refresh interval for periodically reloading a secret from Key Vault. Refresh interval must be greater than 1 hour. Default refresh interval is 12 hours.
+        /// Any refresh operation triggered using <see cref="IConfigurationRefresher"/> will not update the value for a Key Vault secret until the cached value for that secret has expired.
+        /// </summary>
+        /// <param name="key">Key of the Key Vault reference in Azure App Configuration.</param>
+        /// <param name="refreshInterval">Minimum time that must elapse before the secret is reloaded from Key Vault.</param>
+        public AzureAppConfigurationKeyVaultOptions SetSecretRefreshInterval(string key, TimeSpan? refreshInterval = null)
```
drago-draganov, line 66: Why don't we require a valid refreshInternal instead? What does null provides?
It's named SetSecretRefreshInterval, while the actual interval doesn't need to be specified....it's odd.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
drago-draganov, line 68: Why is there a minimum refresh interval? AppConfiguration doesn't own the Keyvault

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -14,9 +14,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     /// </summary>
     public class AzureAppConfigurationKeyVaultOptions
     {
+        private static readonly TimeSpan DefaultRefreshInterval = TimeSpan.FromHours(12);
+        private static readonly TimeSpan MinimumRefreshInterval = TimeSpan.FromHours(1);
```
zhenlan, line 18: I would suggest we don't introduce a default or minimum interval for secret refresh. This value can be very different for each user depending on their policy requirements. Leave it to users.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
zhenlan, line 66: It would be odd to see a Set*Interval() API does not need an interval to be provided. Suggest not to accept default for interval.

It will however, be useful if we can provide `SetSecretRefreshInterval(TimeSpan refreshInterval)` which refreshes all the rest of secrets.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -48,8 +46,29 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
+
+                    if (secret != null)
+                    {
+                        UpdateCachedKeyVaultSecrets(key, secretValue);
+                    }
+                    else
+                    {
+                        // Secret may have been deleted from KeyVault.
+                        // Delete the secret from cache too.
+                        RemoveExpiredSecretFromCache(key);
```
zhenlan, line 69: NIT: just for readability, the secret deleted from keyvault may or may not expired, so I would avoid using "ExpiredSecret" in the function name.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -63,6 +82,38 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
 
+        internal bool AnyExpiredSecrets()
+        {
+            bool shouldRefreshKeyVaultSecrets = false;
+            List<CachedKeyVaultSecret> secretsToBeRemovedFromCache = new List<CachedKeyVaultSecret>();
+
+            foreach (var cachedSecret in _cachedKeyVaultSecrets)
+            {
+                // Skip the refresh for this key vault secret if it has no expiration time or if it hasn't expired yet
+                if (cachedSecret.ExpiresOn == null || DateTimeOffset.UtcNow < cachedSecret.ExpiresOn)
```
zhenlan, line 93: I might miss something. I thought we didn't make decisions based on expiration anymore.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
zhenlan: +1 for `ClearCache`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
zhenlan: +1, maybe `RemoveSecretFromCache`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
zhenlan: Now I see - it's the cache expiration time, not the secret :)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
zhenlan: Not sure why someone will reference the same secret with two different keys, but yes, that behavior will be as currently designed. Hard to say which behavior will be expected since the user specifically tells us to refresh one but not the other. If we have a scenario in mind, maybe it's easier to decide.

QQ, if people do set refresh interval on both keys, will we refresh the same secret twice?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -429,6 +441,12 @@ private async Task RefreshIndividualKeyValues()
                             // Add the key-value if it is not loaded, or update it if it was loaded with a different label
                             _applicationSettings[watchedKey] = watchedKv;
                             _watchedSettings[watchedKeyLabel] = watchedKv;
+
+                            // Invalidate the cached Key Vault secret (if any) for this ConfigurationSetting
+                            foreach (IKeyValueAdapter adapter in _options.Adapters)
+                            {
+                                adapter.InvalidateCache(watchedKv);
```
jimmyca15, line 448: >adapter.InvalidateCache(watchedKv); [](http://example.com/codeflow?start=32&length=35)

Shouldn't we also invalidate for changed key?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
drago-draganov, line 66: Perhaps the name 'key' doesn't explain what is for. Can we have it a bit more specific? i.e. 'secretReferenceKey'

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Yes, we will refresh the same secret twice if both keys point to the same Key Vault reference. If a refresh interval is set, we will honor it and refresh the secret for every unique key. We don't check what the Key Vault reference is while refreshing secrets.

jimmyca15: I think that's fine. I just wanted to make sure we are on the same page.

zhenlan: The problem can be solved if we introduce references to other keys in appconfig in the future.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Updated to `RefreshKeyValueAdapters`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Updated to `RemoveSecretFromCache`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Updated to `ClearCache`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Switched to using a thread safe collection for cached key vault secrets.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Moved the deletion logic out of this query method as discussed offline. Now `GetSecretValue` will take care of updating/removing stale secrets.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
avanigupta: Made the refresh interval mandatory.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
avanigupta: Removed `MinimumRefreshInterval`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
avanigupta: Removed `MinimumRefreshInterval`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
avanigupta: Added new API to refresh rest of the secrets.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Renamed to `RefreshAt` so that it's less confusing.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
avanigupta: Done.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -48,12 +42,33 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
+
+                    if (secret != null)
+                    {
+                        UpdateCachedKeyVaultSecrets(key, secretValue);
+                    }
+                    else
+                    {
+                        // Secret may have been deleted from KeyVault.
+                        // Delete the secret from cache too.
+                        RemoveSecretFromCache(key);
```
jimmyca15, line 65: >RemoveSecretFromCache [](http://example.com/codeflow?start=24&length=21)

Don't we want to cache null in that case?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -63,6 +78,22 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
             return secretValue;
         }
 
+        public bool ShouldRefreshKeyVaultSecrets()
+        {
+            // return true if the RefreshAt time of any cached secret has already elapsed.
+            return _cachedKeyVaultSecrets.Any(cachedSecret => cachedSecret.Value.RefreshAt.HasValue && cachedSecret.Value.RefreshAt.Value < DateTimeOffset.UtcNow);
```
jimmyca15, line 84: >_cachedKeyVaultSecrets [](http://example.com/codeflow?start=19&length=22)

Iterating over concurrent dictionary creates a copy each time. There's an optimization to be had in this provider around caching the next known expiration time.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -450,6 +468,18 @@ private async Task RefreshIndividualKeyValues()
             }
         }
 
+        private async Task RefreshKeyValueAdapters()
+        {
+            foreach (IKeyValueAdapter adapter in _options.Adapters)
+            {
+                if (adapter.NeedsRefresh())
```
jimmyca15, line 475: >adapter.NeedsRefresh() [](http://example.com/codeflow?start=20&length=22)

If `RefreshAsync` is called many times rapidly we don't want to call SetData each time if the initial refresh operation has started. This is a real scenario for high throughput web apps using our middleware. It looks like we need a semaphore that is tied to the adapter.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -48,12 +42,33 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
+                if (secretValue == null)
+                {
+                    // We dont have a cached secret value or the cached value has expired.
+                    // Get the secret from Key Vault and update the cache.
+                    secret = await client.GetSecretAsync(secretName, secretVersion, cancellationToken).ConfigureAwait(false);
+                    secretValue = secret?.Value;
+
+                    if (secret != null)
+                    {
+                        UpdateCachedKeyVaultSecrets(key, secretValue);
```
jimmyca15, line 59: >UpdateCachedKeyVaultSecrets [](http://example.com/codeflow?start=24&length=27)

I would update that name to `Cache` to make that a bit easier to follow. From the name I thought there was a bit more going on at this step.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ErrorMessages.cs
```diff
@@ -6,5 +6,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     internal class ErrorMessages
     {
         public const string CacheExpirationTimeTooShort = "The cache expiration time cannot be less than {0} milliseconds.";
+        public const string SecretRefreshIntervalTooShort = "The secret refresh interval cannot be less than {0} hour(s).";
```
jimmyca15, line 9: >SecretRefreshIntervalTooShort [](http://example.com/codeflow?start=28&length=29)

Is this used?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -52,5 +54,28 @@ public AzureAppConfigurationKeyVaultOptions SetSecretResolver(Func<Uri, ValueTas
...
             return this;
         }
+
+        /// <summary>
+        /// Sets the refresh interval for periodically reloading a secret from Key Vault. 
+        /// Any refresh operation triggered using <see cref="IConfigurationRefresher"/> will not update the value for a Key Vault secret until the cached value for that secret has expired.
+        /// </summary>
+        /// <param name="secretReferenceKey">Key of the Key Vault reference in Azure App Configuration.</param>
+        /// <param name="refreshInterval">Minimum time that must elapse before the secret is reloaded from Key Vault.</param>
+        public AzureAppConfigurationKeyVaultOptions SetSecretRefreshInterval(string secretReferenceKey, TimeSpan refreshInterval)
```
jimmyca15, line 64: >secretReferenceKey [](http://example.com/codeflow?start=84&length=18)

Validate string for null or empty.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/CachedKeyVaultSecret.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using System;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault
+{
+    internal struct CachedKeyVaultSecret
```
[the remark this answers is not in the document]
avanigupta: Changed to class

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: updated logic to cache null in this case.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Tracking the next refresh time using a private field instead of iterating over concurrent dictionary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: added a semaphore to fix this issue.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Changed to  `SetSecretInCache`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ErrorMessages.cs
avanigupta: removed

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
avanigupta: updated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -470,14 +471,26 @@ private async Task RefreshIndividualKeyValues()
 
         private async Task RefreshKeyValueAdapters()
         {
-            foreach (IKeyValueAdapter adapter in _options.Adapters)
+            if (!AdapterRefreshSemaphore.Wait(0))
```
jimmyca15, line 474: If we don't need a lock we shouldn't take it.

It looks like we may want
```
if (adapters.Any(a => a.NeedsRefresh() &&
    AdapterRefreshSemaphore.Wait(0))
{
    try
    {
        if (adapters.Any(a => a.NeedsRefresh())
        {
            await SetData(_applicationSettings).ConfigureAwait(false);
         }
    }
    finally
    {
           AdapterRefreshSemaphore.Release();
    ]
}
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
jimmyca15: I misspoke. Only [`Keys`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.keys?view=net-5.0#System_Collections_Concurrent_ConcurrentDictionary_2_Keys) and [`Values`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.values?view=net-5.0#System_Collections_Concurrent_ConcurrentDictionary_2_Values) creates a copy.

[Enumeration ](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getenumerator?view=net-5.0#System_Collections_Concurrent_ConcurrentDictionary_2_GetEnumerator)uses the dictionary in place.

avanigupta: Reverted.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -14,32 +13,28 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault
     internal class AzureKeyVaultSecretProvider
     {
         private readonly IDictionary<string, SecretClient> _secretClients;
-        private readonly TokenCredential _credential;
-        private readonly Func<Uri, ValueTask<string>> _secretResolver;
+        private Dictionary<string, CachedKeyVaultSecret> _cachedKeyVaultSecrets = new Dictionary<string, CachedKeyVaultSecret>();
```
jimmyca15, line 16: >Dictionary [](http://example.com/codeflow?start=16&length=10)

looks like it can be readonly. Same with `_keyVaultOptions`.

jimmyca15, line 16: >new Dictionary [](http://example.com/codeflow?start=82&length=14)

We're initializing _secretClients in the constructor. We probably want to follow the same pattern for both.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -72,14 +102,58 @@ private SecretClient GetSecretClient(Uri secretUri)
...
+            foreach (KeyValuePair<string, CachedKeyVaultSecret> secret in _cachedKeyVaultSecrets)
+            {
+                if (secret.Value.RefreshAt.HasValue && secret.Value.RefreshAt.Value < _nextRefreshTime)
+                {
+                    _nextRefreshTime = secret.Value.RefreshAt;
+                    _nextRefreshKey = secret.Key;
+                }
+            }
+
+            _nextRefreshTime = _nextRefreshTime != DateTimeOffset.MaxValue ? _nextRefreshTime : null;
```
jimmyca15, line 156: >_nextRefreshTime = _nextRefreshTime != DateTimeOffset.MaxValue ? _nextRefreshTime : null; [](http://example.com/codeflow?start=12&length=89)

We're doing an assignment, but really it's more like an if statement. Can we write it as an `if` to be more clear.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -48,12 +43,25 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
...
+                    secret = await client.GetSecretAsync(secretName, secretVersion, cancellationToken).ConfigureAwait(false);
+                    secretValue = secret?.Value;
+                    SetSecretInCache(key, secretValue);
+                }
             }
-            else if (_secretResolver != null)
+            else if (_keyVaultOptions.SecretResolver != null)
             {
-                secretValue = await _secretResolver(secretUri).ConfigureAwait(false);
+                secretValue = await _keyVaultOptions.SecretResolver(secretUri).ConfigureAwait(false);
```
jimmyca15, line 64: >_keyVaultOptions.SecretResolver [](http://example.com/codeflow?start=36&length=31)

How come we're not caching secrets returned by secret resolver?

avanigupta: We don't know what users have implemented in SecretResolver. They may want the callback to run every time instead of caching the value. Also, if users are defining a secret resolver, I think it's safe to assume that they are responsible for managing the custom implementation and optimizing the callback themselves.

jimmyca15: But in this case we don't give them the option to rotate secrets/certificates

avanigupta: That's correct. This callback is not meant for getting secrets/certs in the first place. It was introduced as a way to avoid resolving secrets from Key Vault by assigning a dummy value to key vault references. I don't think there should be an expectation to provide the same caching behavior for this custom implementation.

jimmyca15: I'd say that's half true. Yes we designed the API because of a scenario where users wanted to skip resolving certain secrets. However, the API we designed is one that allows custom resolution of secrets instead of KeyVaultClient. This means that it is fine to use for custom resolution of secrets. 

Caching + rotation is a value add a layer above the resolution of secrets. I don't think it should depend on what client/resolver we use to get a secret.

avanigupta: That makes sense. The only thing is, users don't have an option to opt out of caching+rotation for the custom resolved secrets (that is if they want to use rotation at all). If they use the `SetSecretRefreshInterval(refreshInterval)` API, everything under the custom callback will also be cached. 
But maybe that's not such a big issue. I can't think of a use case where someone would want this. So I'll update the code to cache custom secrets too.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -48,12 +44,34 @@ public async Task<string> GetSecretValue(Uri secretUri, CancellationToken cancel
 
             if (client != null)
```
jimmyca15, line 45: Can we avoid the duplicated cache calcuation snippet ?

```
if (_cachedKeyVaultSecrets.TryGetValue(key, out CachedKeyVaultSecret cachedSecret) &&
    (!cachedSecret.RefreshAt.HasValue || DateTimeOffset.UtcNow < cachedSecret.RefreshAt.Value))
{
    secretValue = cachedSecret.SecretValue;
}
else if (client != null)
{

}
else if (resolver != null)
{

}

return secretValue;
```

jimmyca15 (approved):

Issue #157, opened 2020-04-08
Add ability to perform custom transforms on App Configuration Settings.


jimmyca15:
Some applications may need to perform custom transforms on the configuration settings they retrieve from Azure App Configuration. An example might be if an application uses a custom defined content type and has to process the value before it is acceptable to be placed in .NET Core's `IConfiguration`. This is the kind of thing that is done for Key Vault references today.

This can be accomplished by exposing a new `Map` method to `AzureAppConfigurationOptions`.

### Proposed API

```
/// Performs custom mapping for configuration settings retrieved from Azure App Configuration
public AppConfigurationOptions AppConfigurationOptions.Map(Func<ConfigurationSetting, ConfigurationSetting> mapper);
```

### Usage

```
config.AddAzureAppConfiguration(options =>
    {
        options.Connect(settings["connection_string"])
            .Map((ConfigurationSetting setting) =>
            {
                if (setting.ContentType == "application/myContentType")
                {  
                    return new ConfigurationSetting(
                        key: setting.Key,
                        value: PerformSomeTransform(setting.Value),
                        label: setting.Label);
                }

                return setting;
        });
    });
```

The `Value` of the configuration setting is what is exposed from `IConfiguration`.

CC @zhenlan @drago-draganov @abhilasharora

abhilasharora:
This does seem like a viable approach to expose a `ConfigurationSetting` instance to the users and allow them to apply a transform on its value.

It might be a good idea to brainstorm the posssible scenarios where users might want to use transforms. That should help decide if we want to apply these after or before the feature management and key vault adapters are applied.

Also, if we expose the type directly, it is possible for a user to update the `Key` and `Label` fields in the `ConfigurationSetting` object, which could make it harder to understand if this method would act on those or just ignore them. We might want to expose an interface for the `ConfigurationSetting` class that provides setter for only the subset of properties that users are allowed to edit.

jimmyca15:
I see two solutions to the ordering problem of before/after the built-in adapters.

An enum parameter like

```
enum MapPosition
{
  BeforeBuiltinMapping,
  AfterBuiltinMapping
}
```
```
Map(() => {}, MapPosition.BeforeBuiltinMapping);
```


Or pass a way to perform our value resolution on demand, which I think is even more flexible.

```
[...]

PR #252, opened 2021-04-15
Support replacing key separators with colons (:)

marleypowell:
Key separators can be specified as hierarchical delimiters but when used in the "Configuration API" they are not converted to colon (:) seperators. 
This is a pattern that's already follow for environment variables and [Azure Key Vault configuration provider](https://docs.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-5.0) as specified here: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0#configuration-keys-and-values. 

> Within the Configuration API, a colon separator (:) works on all platforms.

The aim of this PR is to be able to convert the supported separators to colons (:) when configuring Azure App Configuration.

Usage example:

```
var configuration = new ConfigurationBuilder()
    {
        options.Connect(connectionString);

        options.ReplaceKeySeparator(KeySeparator.Period)
               .ReplaceKeySeparator(KeySeparator.Comma)
               .ReplaceKeySeparator(KeySeparator.Semicolon)
               .ReplaceKeySeparator(KeySeparator.ForwardSlash)
               .ReplaceKeySeparator(KeySeparator.Dash)
               .ReplaceKeySeparator(KeySeparator.Underscore)
               .ReplaceKeySeparator(KeySeparator.DoubleUnderscore);
    })
    .Build();
```


marleypowell:
@avanigupta Do I need to do anything to get reviewers on this PR? I've not contributed before so I'm not sure...

avanigupta:
Hi @marleypowell, could you help me understand your specific scenario for which you need this functionality?

marleypowell:
> Hi @marleypowell, could you help me understand your specific scenario for which you need this functionality?

In `AzureAppConfiguration` I used a double underscore as the key separator:
image

This is fine when I export from the azure portal because I have the option to specify the separators to be replaced correctly for the export format (json, yaml, properties):
image

When using `AppConfiguration-DotnetProvider` I would like the format to be `{[parent:child, value]}` instead of `{[parent_child, value]}`. This would allow me to use `Configuration.GetSection("parent")` to bind the whole object including it's child properties to a class. Currently when the delimiter isn't a `colon (:)` you can't bind the parent and all children to an object.
image

This is something that the `AzureKeyVault-DotnetProvider` already does: 
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0#configuration-keys-and-values
image

avanigupta:
@marleypowell thanks for describing your setup!

> 
> In `AzureAppConfiguration` I used a double underscore as the key separator:

Would it be possible for you to use the `:` separator in App Configuration? I'm trying to understand if there might be a better way to manage key-values in App Configuration.

>
> This is something that the `AzureKeyVault-DotnetProvider` already does:

Key Vault has a limitation that it only supports `--` as separator. That's why the Key Vault provider _has to_ support this conversion by default. There's no such limitation in App Configuration.

marleypowell:
> Would it be possible for you to use the `:` separator in App Configuration? I'm trying to understand if there might be a better way to manage key-values in App Configuration.

@avanigupta we have a decided to use a general naming convention which is double underscore. Seeing as the portal supports parsing these variables I don't see why the Dotnet provider shouldn't be able to.

avanigupta:
> 
> 
> > Would it be possible for you to use the `:` separator in App Configuration? I'm trying to understand if there might be a better way to manage key-values in App Configuration.
> 
> @avanigupta we have a decided to use a general naming convention which is double underscore. Seeing as the portal supports parsing these variables I don't see why the Dotnet provider shouldn't be able to.

You're right, the portal supports double underscore as a separator along with many other key separators. I'm trying to understand your motivation behind choosing double underscore over any other separator like dot, hyphen, etc.
For example, is it because you use the double underscore separator in environment variables and your App Config key-values were copied from there?

marleypowell:
> > > Would it be possible for you to use the `:` separator in App Configuration? I'm trying to understand if there might be a better way to manage key-values in App Configuration.
> > 
> > 
> > @avanigupta we have a decided to use a general naming convention which is double underscore. Seeing as the portal supports parsing these variables I don't see why the Dotnet provider shouldn't be able to.
> 
> You're right, the portal supports double underscore as a separator along with many other key separators. I'm trying to understand your motivation behind choosing double underscore over any other separator like dot, hyphen, etc.
> For example, is it because you use the double underscore separator in environment variables and your App Config key-values were copied from there?

Yeah we use the double underscore seperator in environment variables which are currently all split accross azure devops variable groups. We're looking to eventually migrate to an azure app config solution but it's not ideal to have to renamed all of them. Also some variables are used in other places which won't use the DotnetProvider.

avanigupta:
@marleypowell, we are working on a new API which would give users full control over their key-values being read from App Config. This API would allow users to modify the keys and replace any separators with colon before we add the key-value to `IConfiguration`. Please see #157 for more details. Would this resolve your current issue?

PR #254, opened 2021-04-29
Block concurrent network operations

avanigupta:
This PR implements thread synchronization to ensure that multiple threads do not make network calls simultaneously. 
We have two [critical sections](https://en.wikipedia.org/wiki/Critical_section) where network calls are made: 
- App startup
- Refresh
 
Access to the critical sections will be controlled using [System.Threading.Interlocked.Exchange](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.exchange?view=net-5.0) atomic operation.

As a result, we no longer need to maintain semaphores in `AzureAppConfigurationProvider` or `KeyValueWatcher`.  
`ConcurrentDictionary` can also be replaced by `Dictionary`.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -33,13 +33,15 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private ConcurrentDictionary<KeyValueIdentifier, ConfigurationSetting> _watchedSettings = new ConcurrentDictionary<KeyValueIdentifier, ConfigurationSetting>();
 
         private readonly TimeSpan MinCacheExpirationInterval;
-        private readonly SemaphoreSlim InitializationSemaphore = new SemaphoreSlim(1);
 
         // The most-recent time when the refresh operation attempted to load the initial configuration
         private DateTimeOffset InitializationCacheExpires = default;
 
         private static readonly TimeSpan MinDelayForUnhandledFailure = TimeSpan.FromSeconds(5);
         private static readonly TimeSpan DefaultMaxSetDirtyDelay = TimeSpan.FromSeconds(30);
+
+        // To avoid concurrent network operations, this flag is used to achieve synchronization between multiple threads.
+        private static int _networkOperationsInProgress = 0;
 
         public Uri AppConfigurationEndpoint
         {
@@ -105,11 +107,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         public override void Load()
         {
             var watch = Stopwatch.StartNew();
-            var refresher = (AzureAppConfigurationRefresher)_options.GetRefresher();
-            refresher.SetProvider(this);
 
             try
             {
+                // Load() is invoked only once during application startup. We don't need to check for concurrent network
+                // operations here because there can't be any other startup or refresh operation in progress at this time.
                 LoadAll(_optional).ConfigureAwait(false).GetAwaiter().GetResult();
             }
             catch (ArgumentException)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -33,14 +33,16 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
...
-        private readonly SemaphoreSlim InitializationSemaphore = new SemaphoreSlim(1);
 
         // The most-recent time when the refresh operation attempted to load the initial configuration
         private DateTimeOffset InitializationCacheExpires = default;
 
         private static readonly TimeSpan MinDelayForUnhandledFailure = TimeSpan.FromSeconds(5);
         private static readonly TimeSpan DefaultMaxSetDirtyDelay = TimeSpan.FromSeconds(30);
 
+        // To avoid concurrent network operations, this flag is used to achieve synchronization between multiple threads.
+        private static int _networkOperationsInProgress = 0;
```
jimmyca15, line 44: >static [](http://example.com/codeflow?start=16&length=6)

Why is it static? Two different providers should be able to operate independently.

avanigupta: You're right. It shouldn't be static.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -33,14 +33,16 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
...
         private readonly TimeSpan MinCacheExpirationInterval;
-        private readonly SemaphoreSlim InitializationSemaphore = new SemaphoreSlim(1);
 
         // The most-recent time when the refresh operation attempted to load the initial configuration
         private DateTimeOffset InitializationCacheExpires = default;
 
         private static readonly TimeSpan MinDelayForUnhandledFailure = TimeSpan.FromSeconds(5);
         private static readonly TimeSpan DefaultMaxSetDirtyDelay = TimeSpan.FromSeconds(30);
 
+        // To avoid concurrent network operations, this flag is used to achieve synchronization between multiple threads.
```
jimmyca15, line 43: >// To avoid concurrent [](http://example.com/codeflow?start=8&length=22)

I would say watch out for comments that might be implied by the variable name alone. Otherwise we may end up with a lot of comment bloat.

jimmyca15 (changes requested):
just the static.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: I agree. In this case, since the use of this variable is more complicated, I wanted to explain the concept too for future readers.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -294,21 +314,11 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
...
-                return;
-            }
-
-            InitializationCacheExpires = DateTimeOffset.UtcNow.Add(MinCacheExpirationInterval);
-
-            try
+            if (InitializationCacheExpires < DateTimeOffset.UtcNow)
             {
+                InitializationCacheExpires = DateTimeOffset.UtcNow.Add(MinCacheExpirationInterval);
                 await LoadAll(ignoreFailures: false).ConfigureAwait(false);
```
drago-draganov, line 320: Don't need to await anymore.

avanigupta: Updated.

jimmyca15 (approved):

PR #262, opened 2021-06-18
Initial OfflineFileCache Changes

avanigupta:
To make it easier to review the new OfflineFileCache implementation, this PR only covers the following basic changes:

- Updates to `OfflineFileCacheOptions`:
    - Added DataProtector (optional)
    - Added FileCacheExpiration (mandatory)
    - Updated Path to be mandatory
    - Removed Key, IV and SignKey
- Removed our old implementation of encryption/decryption.
- Added time limited encryption/decryption using DataProtection APIs.
- Updated the fields in cache file:
    - Encrypted data (d)
    - Plain text scope token (s)
    - Plain text expiration time (e)
- Updated `Microsoft.Extensions.DependencyInjection.Abstractions` from 2.1.1 to 3.1.16 so that we can use more recent 3.xx DataProtection packages.
- Added new tests and removed the dependency on `cache.json` offline cache test file. Since new cache files have an expiration time, we cannot use persisted test files anymore. 

Other behavioral changes will be made in future PRs.


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -308,5 +248,22 @@ internal static void ValidateCachePath(string path)
...
+        internal static void ValidateFileCacheExpiration(TimeSpan expiration)
+        {
+            if (expiration <= TimeSpan.Zero)
+            {
+                throw new ArgumentException($"Please provide a valid {nameof(OfflineFileCacheOptions)}.{nameof(OfflineFileCacheOptions.FileCacheExpiration)}.");
+            }
+
+            try
+            {
+                _ = DateTimeOffset.UtcNow + expiration;
```
jimmyca15, line 261: Can this be rewritten without exception ? `DateTimeOffset.MaxValue - expiration > DatetimeOffset.UtcNow

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -308,5 +248,22 @@ internal static void ValidateCachePath(string path)
                 throw new ArgumentException($"The directory with path {directoryPath} does not exist.");
             }
         }
+
+        internal static void ValidateFileCacheExpiration(TimeSpan expiration)
+        {
+            if (expiration <= TimeSpan.Zero)
+            {
+                throw new ArgumentException($"Please provide a valid {nameof(OfflineFileCacheOptions)}.{nameof(OfflineFileCacheOptions.FileCacheExpiration)}.");
```
jimmyca15, line 256: Looks like it's also out of range.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -285,12 +222,15 @@ private void EnsureOptions(AzureAppConfigurationOptions azconfigOptions)
...
         internal static void ValidateCachePath(string path)
         {
+            if (path == null)
+            {
+                throw new ArgumentNullException($"{nameof(OfflineFileCacheOptions)}.{nameof(OfflineFileCacheOptions.Path)}", "Please provide the path for storing offline file cache.");
+            }
+
             if (!Path.IsPathRooted(path) || !string.Equals(Path.GetFullPath(path), path) || string.IsNullOrWhiteSpace(Path.GetFileName(path)))
             {
                 throw new ArgumentException($"The path {path} is not a full file path.");
```
jimmyca15, line 236: Can we update this to use the second parameter, argument name.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
jimmyca15: With ARgumentOutOfRange just need param name in constructor.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -308,5 +248,22 @@ internal static void ValidateCachePath(string path)
...
+                throw new ArgumentException($"Please provide a valid {nameof(OfflineFileCacheOptions)}.{nameof(OfflineFileCacheOptions.FileCacheExpiration)}.");
+            }
+
+            try
+            {
+                _ = DateTimeOffset.UtcNow + expiration;
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                throw new ArgumentOutOfRangeException($"{nameof(OfflineFileCacheOptions)}.{nameof(OfflineFileCacheOptions.FileCacheExpiration)}", "Please provide a shorter file cache expiration. The expiration time added to the current UTC time results in an un-representable DateTime.");
```
jimmyca15, line 265: > The expiration time added to the current UTC time results in an un-representable DateTime.

That's going pretty deep into the details. would probably be fine to just call out it's too big.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCacheOptions.cs
```diff
@@ -10,22 +13,27 @@ public class OfflineFileCacheOptions
     {
         /// <summary>
         /// The file path to use for persisting cached data.
+        /// If your application is running in Azure App Service, storing the cache file in %HOME%
```
jimmyca15, line 16: 👍

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: I tried that too. If `expiration` is `TimeSpan.MaxValue`, `DateTimeOffset.MaxValue - expiration` also threw `ArgumentOutOfRangeException`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -237,43 +194,23 @@ private async Task DoUpdate(string tempFile, int retry = 0)
             }
         }
 
-        private void EnsureOptions(AzureAppConfigurationOptions azconfigOptions)
+        private void CreateScopeToken(AzureAppConfigurationOptions azconfigOptions)
         {
             if (azconfigOptions == null)
```
jimmyca15, line 199: Can you move this to public API entry point and use debug.assert here. Also, param name should match public API.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -237,43 +194,23 @@ private async Task DoUpdate(string tempFile, int retry = 0)
...
-                //
                 // The default scope token is the configuration store endpoint combined with all of the key-value filters
-
-                string endpoint = ConnectionStringParser.Parse(azconfigOptions.ConnectionString, "Endpoint");
-
+                string endpoint = (azconfigOptions.Endpoint != null)
+                                  ? azconfigOptions.Endpoint.ToString() 
+                                  : ConnectionStringParser.Parse(azconfigOptions.ConnectionString, "Endpoint");
+                
                 if (string.IsNullOrWhiteSpace(endpoint))
```
jimmyca15, line 211: How can this happen?

jimmyca15: That would be if someone passed a cooked AppConfigOptions that doesn't have correct endpoint or connection string? In that case wouldn't it be arg exception?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -144,23 +103,25 @@ public string Import(AzureAppConfigurationOptions options)
...
-                            }
-                        }
+                        return _timeLimitedDataProtector.Unprotect(data);
                     }
                 }
                 catch (IOException ex) when (ex.HResult == ERROR_SHARING_VIOLATION)
                 {
                     Task.Delay(new Random().Next(delayRange)).ConfigureAwait(false).GetAwaiter().GetResult();
                 }
+                catch (CryptographicException)
```
jimmyca15, line 115: The expired data case is interesting. I imagine we should treat that as if the file doesn't exist at all rather than throwing.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -130,12 +93,8 @@ public string Import(AzureAppConfigurationOptions options)
                                     data = reader.ReadAsString();
```
jimmyca15, line 93: I'm wondering if we should add a version property representing the schema version. This is one way we can ensure that we don't interact with an old offline cache file.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: I wanted to handle `CryptographicException` in `AzureAppConfigurationProvider` where `Import` is called. Provider might need to decide whether to rethrow this exception with a good error message, or silently fail.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
jimmyca15: This validation is interesting. We use expiration at later points during runtime, at which the addition here may overflow. We're validating something who's validity depends on when it's validated. It may require being more restrictive on upper limit to be reasonable.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -285,12 +222,15 @@ private void EnsureOptions(AzureAppConfigurationOptions azconfigOptions)
 
                 _scopeToken = sb.ToString();
```
jimmyca15, line 223: It looks like scopetoken is dependent on method local variables. I don't think it can be a member.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: If we do read an old offline cache file, we won't be able to decrypt it with the new DataProtector anyway. So `Import` is definitely going to fail. Do you have any scenario in mind where we might need to be more careful?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
jimmyca15: > I wanted to handle CryptographicException in AzureAppConfigurationProvider

CryptographicException is quite generic. Can we distinguish the case where the data has expired? That's the only one I'm thinking about at the moment. The rest will most likely go unhandled.

jimmyca15: I think we should treat expired data case the same as the file not existing.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: I checked that this can never happen. If the connection string is invalid or endpoint is missing, AppConfig SDK will throw an exception during `AzureAppConfigurationSource.Build()` as it tries to create the `ConfigurationClient`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: I checked the HResult for different CryptographicExceptions. They always have the same HResult (-2146233087). Only the message is different. For example:
- Expired data = "The payload expired at 6/23/2021 11:22:04 PM +00:00."
- Tampered data = "The provided payload cannot be decrypted because it was not protected with this protection provider."
- Deleted encryption keys = "The provided payload cannot be decrypted because it was not protected with this protection provider."
- Some old invalid data (not sure how it was invalid) = "The payload was invalid."

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: This exception will be thrown even if someone doesn't set the FileCacheExpiration since we can't really check for null here. I could split the check and handle `expiration==TimeSpan.Zero` and `expiration<TimeSpan.Zero` separately, but even then, we can't pin point if user passed 0 expiration or just forgot to set it. So I just bundled those up into one `ArgumentException`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: I agree that this addition could still overflow at the time when data expiration is actually calculated. But we can't really have an arbitrary upper limit here. DataProtection docs suggest that the expiration should be no longer than a few weeks/one month. But they also don't have any enforcement for the upper limit. They just bubble up `ArgumentOutOfRangeException` if expiration is too big.

The only reason I want to keep this check here is because there might be some users who might try to persist their cache file forever and will therefore pass `TimeSpan.MaxValue` as their file expiration. For such users, this is an error that can be thrown much earlier during application startup instead of failing at the end of startup.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
jimmyca15: Perhaps we should persist the expiration date unencrypted so that we can avoid an exception in the expired data case.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -237,75 +197,76 @@ private async Task DoUpdate(string tempFile, int retry = 0)
...
             }
 
-            _options = options;
+            if (_options.DataProtector == null)
+            {
+                IDataProtectionProvider dataProtectionProvider = DataProtectionProvider.Create($"{_options.Path}-{_appConfigEndpoint}");
+                _options.DataProtector = dataProtectionProvider.CreateProtector($"AppConfigurationOfflineFileCacheProtector-{_appConfigEndpoint}");
+            }
+
+            _timeLimitedDataProtector = _timeLimitedDataProtector ?? _options.DataProtector.ToTimeLimitedDataProtector();
```
jimmyca15, line 228: >_timeLimitedDataProtector = _timeLimitedDataProtector [](http://example.com/codeflow?start=12&length=53)

This kind of looks like an unintended assignment since you're assigning the same variable. If this is just for initialization I'd recommend using an if statment.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -237,75 +197,76 @@ private async Task DoUpdate(string tempFile, int retry = 0)
...
                 }
 
                 _scopeToken = sb.ToString();
             }
 
-            _options = options;
+            if (_options.DataProtector == null)
+            {
+                IDataProtectionProvider dataProtectionProvider = DataProtectionProvider.Create($"{_options.Path}-{_appConfigEndpoint}");
+                _options.DataProtector = dataProtectionProvider.CreateProtector($"AppConfigurationOfflineFileCacheProtector-{_appConfigEndpoint}");
```
jimmyca15, line 225: >_options.DataProtector [](http://example.com/codeflow?start=16&length=22)

Do we need to set this? It kind of looks like it's just being used for conditional initialization logic.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: If we persist the expiration date unencrypted, someone can manually edit the expiration time to use the file forever. In that case, we will have to include something like a digital signature for the entire file to make sure that the expiration time hasn't been tampered with. That is an extra step that the time limited data protector saves.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
jimmyca15: That's fine, if we need versioning because we change some formatting then we can worry about that in the future and consider v1 to be the object that doesn't have version.

jimmyca15: What's going to be the expectation for customers who reference the new provider, and configure the offline file cache to point to the same file path?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
jimmyca15: It doesn't change the fact that we use a time limited data protector. If the date gets extended we get a cryptographic exception when we try to decrypt the data. But if the user wants to cause a cryptographic exception all they need to do is modify the encrypted payload and they could do the same thing.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
```diff
@@ -17,9 +17,22 @@
...
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.0.1" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.1.1" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.1.1" />
     <PackageReference Include="System.Text.Json" Version="4.6.0" />
   </ItemGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="5.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.DataProtection.Abstractions" Version="5.0.7" />
+    <PackageReference Include="Microsoft.AspNetCore.DataProtection" Version="5.0.7" />
```
jimmyca15, line 26: I don't believe 5.x packages are expected to be used with netcoreapp3.1. The 3.x train of packages are still maintained side-by-side with 5.x https://www.nuget.org/packages/Microsoft.AspNetCore.DataProtection/6.0.0-preview.5.21301.17

jimmyca15: Also, 2.x is still supported until August 21. https://dotnet.microsoft.com/platform/support/policy/dotnet-core

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
```diff
@@ -17,9 +17,22 @@
     <PackageReference Include="Azure.Data.AppConfiguration" Version="1.0.0" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.0.1" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.1.1" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.1.1" />
     <PackageReference Include="System.Text.Json" Version="4.6.0" />
   </ItemGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
```
jimmyca15, line 23: I'm not sure multi-targeting is needed because of these new references. They are all still maintained as netstandard.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: In that case, its possible to get a false positive for expired data. Consider this:
Someone manually modified the expiration date in the cache file to a time in the past. They also tampered with the encrypted data. Now when we get a CryptographicException, we are going to deduce that this is because the file has expired, when in fact its the data that is corrupt.

I'm not sure why we need to differentiate between expired data vs other cryptographic exceptions. I agree the behavior in case of expired data and cache file not being present is the same. But isn't the end result same for all other cryptographic exceptions too - throw an exception and let users deal with it?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: In the cache-first approach during startup, I think we should swallow (and log in future) any exceptions while reading from cache file and continue loading from server.
The first time data is loaded from server, their cache file will be updated with new data encrypted using data protector.
If they also get a server exception on the first try loading from server, then we can throw the server exception.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
avanigupta: You're right. Not sure why I had to introduce multi targeting during my testing, but now I see that its not really needed. We have two options:
1. Upgrade `Microsoft.Extensions.DependencyInjection.Abstractions` to 5.x so that we can use 5.x DataProtection packages.
2. Add 2.x DataProtection packages if we don't want to upgrade `Microsoft.Extensions.DependencyInjection.Abstractions`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: We do need to set `_options.DataProtector`. This is because every time we call `EnsureOptions`, we need to check if `_options.DataProtector` is null. If it's null, that means users have not provided an instance of DataProtector and we need to create a default instance only once.

jimmyca15: But if we check if _timeLimitedDataProtector is null instead don't we avoid that skip without assigning that on the input options?

avanigupta: _timeLimitedDataProtector will be null on the first import/export call regardless of whether users have provided a DataProtector or not.

We will have to do something like this:
```
if (_timeLimitedDataProtector == null)
{
    if (_options.DataProtector == null)
    {
        IDataProtectionProvider dataProtectionProvider = DataProtectionProvider.Create($"{_options.Path}-{_appConfigEndpoint}");
       _timeLimitedDataProtector = dataProtectionProvider.CreateProtector($"AppConfigurationOfflineFileCacheProtector-{_appConfigEndpoint}").ToTimeLimitedDataProtector();
    }
    else
    {
        _timeLimitedDataProtector  = _options.DataProtector.ToTimeLimitedDataProtector();
    }
}
```

jimmyca15: Right, which makes sense if `_timeLimitedDataProtector` is what we are trying to initialize

avanigupta: Okay, I'll change it to initialize _timeLimitedDataProtector  only.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: Moved this logic to `Export` because we need to calculate the expiration time anyway.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: Storing the expiration time in the file and using that for two purposes:
1. If expiration time has already passed, avoid CryptographicException by not decrypting the data.
2. If cache file doesn't contain expiration time, it could be an old cache file that the new provider will not be able to decrypt. Avoid another CryptographicException by not decrypting files which don't have expiration time.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
avanigupta: Updated `Microsoft.Extensions.DependencyInjection.Abstractions` to 3.x and added 3.x DataProtection packages.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
```diff
@@ -17,8 +17,11 @@
     <PackageReference Include="Azure.Data.AppConfiguration" Version="1.0.0" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.0.1" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.1.1" />
```
jimmyca15, line 19: If we're using 3.x now we probably want to update the Microsoft.Extensions.Configuration reference.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -144,23 +111,29 @@ public string Import(AzureAppConfigurationOptions options)
...
+                catch (CryptographicException)
+                {
+                    // TBD: This should be handled in AzureAppConfigurationProvider class.
+                    // This exception means that one of the following occured:
+                    //  - the encryption key is invalid/inaccessible;
+                    //  - encrypted data has been tampered with;
+                    //  - data has expired;
+                    //  - any other internal error during decryption.
+                    throw;
+                }
```
jimmyca15, line 136: This doesn't look like it's being used for anything. Should be able to remove the catch

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -170,20 +143,35 @@ public string Import(AzureAppConfigurationOptions options)
...
+            string expiryString;
+
+            try
+            {
+                expiryTime = DateTimeOffset.UtcNow + _options.FileCacheExpiration;
+                expiryString = expiryTime.ToString(CultureInfo.InvariantCulture);
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                throw new ArgumentOutOfRangeException($"{nameof(OfflineFileCacheOptions)}.{nameof(OfflineFileCacheOptions.FileCacheExpiration)}", "Please provide a shorter file cache expiration.");
```
jimmyca15, line 165: >ArgumentOutOfRangeException [](http://example.com/codeflow?start=26&length=27)

Is it reasonable to just set the expirty to max datetime here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
```diff
@@ -192,9 +180,9 @@ public void Export(AzureAppConfigurationOptions options, string data)
                             using (var writer = new Utf8JsonWriter(memoryStream))
                             {
                                 writer.WriteStartObject();
-                                writer.WriteString(dataProp, Convert.ToBase64String(encryptedBytes));
-                                writer.WriteString(hashProp, dataHash);
-                                writer.WriteString(scopeProp, scopeHash);
+                                writer.WriteString(dataProp, encryptedData);
+                                writer.WriteString(scopeProp, _scopeToken);
+                                writer.WriteString(expiryProp, expiryString);
```
jimmyca15, line 185: >expiryString [](http://example.com/codeflow?start=63&length=12)

The variable naming is unconventional, usually type isn't a part of the name and should be avoided. But I see it's because there are two "expiry" here. Could be avoided by not having expiryString at all and just doing the tostring here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: Although its possible to set expiration to `DateTimeOffset.MaxValue`, we shouldn't be doing it because its discouraged in Data Protection docs. I think its safer to throw the exception instead of allowing users to create cache files that persist indefinitely.

jimmyca15: And that's even though we already validated the expiration when we validated options on construction? In this case it would have to be that the initial expiry was valid, but now it is not.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: How about expiryToken? 
The same naming problem exists in `Import` too, and there we can't avoid an intermediate string while reading from file.

jimmyca15: I saw it in Import, but didn't comment because I didn't see a way to avoid it. Here though, it looked like only one variable was needed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: I removed that check from options construction because we have to go through the same check here again.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: I wanted to keep it as a to-do here so that we remember to catch it in `AzureAppConfigurationProvider` as required. In the final version, we shouldn't have to catch it here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/OfflineFileCache.cs
avanigupta: Okay, removed the `expiryString` variable from Export.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
avanigupta: Done

PR #265, opened 2021-08-03
Add user telemetry to correlation context header

avanigupta:
This change adds the following keywords to the `correlation-context` header:
1. **DevEnvironment** - added if request is made from a dev machine.
2. **KeyVaultConfigured** - added if application has invoked `ConfigureKeyVault` API.
3. **OfflineCacheConfigured** - added if application has invoked `SetOfflineCache` API.

Sample header values:

- `RequestType=Startup,DevEnvironment,KeyVaultConfigured,OfflineCacheConfigured`
- `RequestType=Startup,HostType=AzureWebApp,KeyVaultConfigured`

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -619,11 +619,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         private void SetRequestTracingOptions()
         {
-            _requestTracingOptions = new RequestTracingOptions();
-            _requestTracingOptions.HostType = TracingUtils.GetHostType();
-            _requestTracingOptions.IsDevEnvironment = TracingUtils.IsDevEnvironment();
-            _requestTracingOptions.IsKeyVaultConfigured = _options.IsKeyVaultConfigured;
-            _requestTracingOptions.IsOfflineCacheConfigured = _options.IsOfflineCacheConfigured;
+            _requestTracingOptions = new RequestTracingOptions
+            {
+                HostType = TracingUtils.GetHostType(),
+                IsDevEnvironment = TracingUtils.IsDevEnvironment(),
+                IsKeyVaultConfigured = _options.IsKeyVaultConfigured,
+                IsOfflineCacheConfigured = _options.IsOfflineCacheConfigured
+            };
         }
 
         private DateTimeOffset AddRandomDelay(DateTimeOffset dt, TimeSpan maxDelay)
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -25,11 +25,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private bool _isInitialLoadComplete = false;
         private readonly bool _requestTracingEnabled;
 
-        private readonly HostType _hostType;
         private readonly ConfigurationClient _client;
         private AzureAppConfigurationOptions _options;
         private Dictionary<string, ConfigurationSetting> _applicationSettings;
         private Dictionary<KeyValueIdentifier, ConfigurationSetting> _watchedSettings = new Dictionary<KeyValueIdentifier, ConfigurationSetting>();
+        private RequestTracingOptions _requestTracingOptions;
 
         private readonly TimeSpan MinCacheExpirationInterval;
 
@@ -96,7 +96,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             if (_requestTracingEnabled)
             {
-                _hostType = TracingUtils.GetHostType();
+                SetRequestTracingOptions();
             }
         }
 
@@ -382,7 +382,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 if (_watchedSettings.TryGetValue(watchedKeyLabel, out ConfigurationSetting watchedKv))
                 {
                     KeyValueChange keyValueChange = default;
-                    await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, RequestType.Watch, _hostType,
+                    await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, RequestType.Watch, _requestTracingOptions,
                         async () => keyValueChange = await _client.GetKeyValueChange(watchedKv, CancellationToken.None).ConfigureAwait(false)).ConfigureAwait(false);
 
                     changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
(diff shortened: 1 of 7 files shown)


jimmyca15:
Why DevEnvironment instead of Environment=Dev?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -54,37 +54,48 @@ public static HostType GetHostType()
...
+        {
+            await CallWithRequestTracing(tracingEnabled, requestType, requestTracingOptions, () =>
             {
                 clientCall();
                 return Task.CompletedTask;
             }).ConfigureAwait(false);
         }
 
-        public static async Task CallWithRequestTracing(bool tracingEnabled, RequestType requestType, HostType hostType, Func<Task> clientCall)
+        public static async Task CallWithRequestTracing(bool tracingEnabled, RequestType requestType, RequestTracingOptions requestTracingOptions, Func<Task> clientCall)
```
jimmyca15, line 82: What if RequestTracingOptions is null.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -17,6 +21,9 @@ internal class RequestTracingConstants
 
         public const string RequestTypeKey = "RequestType";
         public const string HostTypeKey = "Host";
+        public const string DevEnvironmentTag = "DevEnvironment";
+        public const string KvrConfiguredTag = "KeyVaultConfigured";
```
jimmyca15, line 25: Maybe just `KvConfigured` to match the constant.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -95,6 +106,59 @@ public static async Task CallWithRequestTracing(bool tracingEnabled, RequestType
             }
         }
 
+        private static string CreateCorrelationContextHeader(RequestType requestType, RequestTracingOptions requestTracingOptions)
+        {
+            IList<KeyValuePair<string, string>> correlationContextKeyValues = new List<KeyValuePair<string, string>>();
+            IList<string> correlationContextTags = new List<string>();
+            
+            AddRequestType(correlationContextKeyValues, requestType);
```
jimmyca15, line 114: Can we inline this?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -95,6 +106,59 @@ public static async Task CallWithRequestTracing(bool tracingEnabled, RequestType
...
+
+            if (requestTracingOptions.HostType != HostType.Unidentified)
+            {
+                AddHostType(correlationContextKeyValues, requestTracingOptions.HostType);
```
jimmyca15, line 118: And this one

avanigupta:
> Why DevEnvironment instead of Environment=Dev?

Because we don't plan on adding more environment types.  `Environment=Dev` makes it look like we might have other values for  `Environment`, but we are only interested in one specific environment that is `DevEnvironment`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
avanigupta: In that case, I think `KeyVaultConfigured` is more descriptive and makes more sense in logs than `KvrConfiguredTag`.
How about this:
` public const string KeyVaultConfiguredTag = "KeyVaultConfigured";`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
avanigupta: If request tracing is enabled, we create a new instance of `RequestTracingOptions` in `AzureAppConfigurationProvider` constructor. So `requestTracingOptions` cannot be null.

If request tracing is disabled, we won't hit the code path that is using the `requestTracingOptions` anyway.
But I can still add another null check to be explicit.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -95,16 +106,57 @@ public static async Task CallWithRequestTracing(bool tracingEnabled, RequestType
...
+                headerKeyValues = string.Join(",", correlationContextKeyValues.Select(kvp => $"{kvp.Key}={kvp.Value}"));
+            }
+                
+            if (correlationContextTags.Count > 0)
+            {
+                headerTags = string.Join(",", correlationContextTags);
+            }
+
+            if (!string.IsNullOrWhiteSpace(headerKeyValues) && !string.IsNullOrWhiteSpace(headerTags))
+            {
```
jimmyca15, line 151: The if/else took me a second. We have this pattern elsewhere and an alternative way to do it is like the following.

```
StringBuilder sb

foreach (string s in list)
{
    if (sb.Length > 0)
    {
        sb.Append(",");
    }

    sb.Append(s);
}
```

As far as pro's cons, this approach is more of a list building rather than a branched logic.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -614,7 +614,16 @@ private void ProcessChanges(IEnumerable<KeyValueChange> changes)
...
         {
             var requestType = _isInitialLoadComplete ? RequestType.Watch : RequestType.Startup;
-            await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, requestType, _hostType, clientCall).ConfigureAwait(false);
+            await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, requestType, _requestTracingOptions, clientCall).ConfigureAwait(false);
+        }
+
+        private void SetRequestTracingOptions()
+        {
+            _requestTracingOptions = new RequestTracingOptions();
+            _requestTracingOptions.HostType = TracingUtils.GetHostType();
```
jimmyca15, line 623: Looks like a good candidate for Object Initializer pattern.

```
abc = new def
{
   foo = bar
}
```

jimmyca15 (approved):
:shipit:

jimmyca15:
> > Why DevEnvironment instead of Environment=Dev?
> 
> Because we don't plan on adding more environment types. `Environment=Dev` makes it look like we might have other values for `Environment`, but we are only interested in one specific environment that is `DevEnvironment`.

I'm fine either way, but Environment steers more on the multi-value type field to me rather than flag like KeyVaultConfigured.

avanigupta:
> > > Why DevEnvironment instead of Environment=Dev?
> > 
> > 
> > Because we don't plan on adding more environment types. `Environment=Dev` makes it look like we might have other values for `Environment`, but we are only interested in one specific environment that is `DevEnvironment`.
> 
> I'm fine either way, but Environment steers more on the multi-value type field to me rather than flag like KeyVaultConfigured.

Okay, I can change it to `Environment=Dev`, which also gives us scope to expand this option in future, if needed.

jimmyca15 (approved):

PR #267, opened 2021-08-10
Add telemetry when Key Vault refresh is configured

avanigupta:
Continuation of telemetry collection work done in #265

This PR adds a new telemetry tag to correlation context header if auto-refresh of Key Vault secrets is configured. 
The names of other telemetry options have also been changed:

1. **Env=Dev** - added if request is made from a dev machine.
2. **UsesKeyVault** - added if application has invoked `ConfigureKeyVault` API.
3. **RefreshesKeyVault** - added if application has invoked `SetSecretRefreshInterval` API.
4. **UsesOfflineCache** - added if application has invoked `SetOfflineCache` API.

Sample header values:

- `RequestType=Startup,Env=Dev,UsesKeyVault,UsesOfflineCache`
- `RequestType=Startup,HostType=AzureWebApp,UsesKeyVault,RefreshesKeyVault`
- `RequestType=Watch,HostType=IISExpress,Env=Dev,UsesKeyVault,RefreshesKeyVault,UsesOfflineCache`

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         public const string HostTypeKey = "Host";
         public const string EnvironmentKey = "Environment";
         public const string KeyVaultConfiguredTag = "KeyVaultConfigured";
+        public const string KeyVaultRefreshConfiguredTag = "KeyVaultRefreshConfigured";
         public const string OfflineCacheConfiguredTag = "OfflineCacheConfigured";
 
         public const string DiagnosticHeaderActivityName = "Azure.CustomDiagnosticHeaders";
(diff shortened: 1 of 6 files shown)


jimmyca15 (approved):

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -21,9 +21,11 @@ internal class RequestTracingConstants
 
         public const string RequestTypeKey = "RequestType";
         public const string HostTypeKey = "Host";
-        public const string EnvironmentKey = "Environment";
-        public const string KeyVaultConfiguredTag = "KeyVaultConfigured";
-        public const string OfflineCacheConfiguredTag = "OfflineCacheConfigured";
+        public const string EnvironmentKey = "Env";
+        public const string DevEnvironmentValue = "Dev";
+        public const string KeyVaultConfiguredTag = "HasKeyVault";
```
zhenlan, line 26: Maybe **Use**KeyVault and **Use**OfflineCache?

jimmyca15: In that case, we probably want `UsesKeyVault`?

avanigupta: If we go with "Use", we should call it "**Using**KeyVault" or "**Uses**KeyVault". Both have more letters than "HasKeyVault". But its not a big difference. How about "UsesKeyVault" and "UsesOfflineCache" then?

avanigupta: Ah, just saw Jimmy's comment. I can change it to "Uses..." then.

zhenlan (approved):

Issue #113, opened 2019-12-13
Add IConfigurationRefresher.TryRefresh API


zhenlan:
To make it easier for customers to trigger configuration refresh without worrying about crashing the app due to various transient errors, propose to add the API below in `IConfigurationRefresher`.
```cs
Task<bool> TryRefresh()
```
**Details**
The `TryRefresh` will not throw on Transient or any FailedRequestException, regardless of the status code (even 401, 403)

The existing API `Refresh` will not catch anything.

**More discussion**
Propose to follow the convention and rename APIs to
```cs
Task RefreshAsync();
Task<bool> TryRefreshAsync();
```

abhilasharora:
Closing this issue since the proposed APIs `RefreshAsync` and `TryRefreshAsync` have been released as a part of version `3.0.0` of the package `Microsoft.Extensions.Configuration.AzureAppConfiguration`.

The addition of logging to `TryRefreshAsync` is being tracked at https://github.com/Azure/AppConfiguration-DotnetProvider/issues/134.

PR #273, opened 2021-08-17
Add ILogger support for TryRefreshAsync in .NET Core package

avanigupta:
This PR adds a new public property to the `IConfigurationRefresher` interface to support logging any exceptions that occur during refresh operations:
`ILoggerFactory IConfigurationRefresher.LoggerFactory { get; set; }`

**Using ILoggerFactory with Dependency Injection**

If you want to get your `ILoggerFactory` through dependency injection, the default `ILoggerFactory` will be added automatically when `services.AddAzureAppConfiguration()` is invoked in your `ConfigureServices` method. No code changes are needed.

**Using ILoggerFactory without Dependency Injection**

If you don't use dependency injection, you can set your own instance of `ILoggerFactory` on the `IConfigurationRefresher` instance.

```cs
IConfigurationRefresher refresher = null;
var config = new ConfigurationBuilder()
            {
                {
                    refreshOptions.Register("Sentinel", true)
                        .SetCacheExpiration(TimeSpan.FromSeconds(30));
                })

                refresher = options.GetRefresher();
            })
            .Build();

refresher.LoggerFactory = _loggerFactory;
bool success = await refresher.TryRefreshAsync();

```

**Logging Category**
The category of refresh logs will be: `Microsoft.Extensions.Configuration.AzureAppConfiguration.Refresh`

**Breaking Change**
This is a breaking change because we added a new dependency on `Microsoft.Extensions.Logging` package and the public interface has changed. 

The AspNetCore middleware doesn't need any special changes to support ILogger.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -117,16 +117,16 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         internal bool IsOfflineCacheConfigured { get; private set; } = false;
 
         /// <summary>
-        /// The <see cref="ILoggerFactory"/> for creating a logger to log errors during refresh operation.
+        /// The <see cref="ILoggerFactory"/> for creating a logger to log errors.
         /// </summary>
         internal ILoggerFactory LoggerFactory { get; set; }
 
         /// <summary>
-        /// Register a logger factory to log errors during refresh operations.
+        /// Set a logger factory to log errors.
         /// </summary>
         /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> for creating a logger.</param>
         /// <returns></returns>
-        public AzureAppConfigurationOptions RegisterLoggerFactory(ILoggerFactory loggerFactory)
+        public AzureAppConfigurationOptions SetLoggerFactory(ILoggerFactory loggerFactory)
         {
             LoggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
             return this;
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -117,22 +117,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         internal bool IsOfflineCacheConfigured { get; private set; } = false;
 
         /// <summary>
-        /// The <see cref="ILoggerFactory"/> for creating a logger to log errors.
-        /// </summary>
-        internal ILoggerFactory LoggerFactory { get; set; }
-
-        /// <summary>
-        /// Set a logger factory to log errors.
-        /// </summary>
-        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> for creating a logger.</param>
-        /// <returns></returns>
-        public AzureAppConfigurationOptions SetLoggerFactory(ILoggerFactory loggerFactory)
-        {
-            LoggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
-            return this;
-        }
-
-        /// <summary>
         /// Specify what key-values to include in the configuration provider.
         /// <see cref="Select"/> can be called multiple times to include multiple sets of key-values.
         /// </summary>
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -7,6 +7,7 @@ using Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Models;
+using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -114,6 +115,22 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         /// Flag to indicate whether Offline Cache has been configured.
         /// </summary>
         internal bool IsOfflineCacheConfigured { get; private set; } = false;
+
+        /// <summary>
+        /// The <see cref="ILoggerFactory"/> for creating a logger to log errors during refresh operation.
+        /// </summary>
+        internal ILoggerFactory LoggerFactory { get; set; }
+
+        /// <summary>
+        /// Register a logger factory to log errors during refresh operations.
+        /// </summary>
+        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> for creating a logger.</param>
+        /// <returns></returns>
+        public AzureAppConfigurationOptions RegisterLoggerFactory(ILoggerFactory loggerFactory)
+        {
+            LoggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
+            return this;
+        }
 
         /// <summary>
         /// Specify what key-values to include in the configuration provider.
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -7,7 +7,6 @@ using Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Models;
-using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
 using System.Linq;
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -201,7 +201,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             {
                 if (e.Status == (int)HttpStatusCode.Unauthorized || e.Status == (int)HttpStatusCode.Forbidden)
                 {
-                    _logger?.LogError(e, "Refresh operation failed due to an authentication error.");
+                    _logger?.LogWarning(e, "Refresh operation failed due to an authentication error.");
+                }
+                else
+                {
+                    _logger?.LogWarning(e, "Refresh operation failed due to an error.");
                 }
 
                 return false;
@@ -211,14 +215,18 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 if (e.InnerExceptions.Any(exception => (exception is RequestFailedException ex) 
                                                         && (ex.Status == (int)HttpStatusCode.Unauthorized || ex.Status == (int)HttpStatusCode.Forbidden)))
                 {
-                    _logger?.LogError(e, "Refresh operation failed due to an authentication error.");
+                    _logger?.LogWarning(e, "Refresh operation failed due to an authentication error.");
+                }
+                else
+                {
+                    _logger?.LogWarning(e, "Refresh operation failed due to an error.");
                 }
 
                 return false;
             }
             catch (KeyVaultReferenceException e)
             {
-                _logger?.LogError(e, "Refresh operation failed while resolving a Key Vault reference.");
+                _logger?.LogWarning(e, "Refresh operation failed while resolving a Key Vault reference.");
                 return false;
             }
             catch (OperationCanceledException)
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -231,6 +231,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             }
             catch (OperationCanceledException)
             {
+                _logger?.LogWarning("Refresh operation was canceled.");
                 return false;
             }
 
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -197,8 +197,9 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             {
                 await RefreshAsync().ConfigureAwait(false);
             }
-            catch (Exception ex) when (ex is RequestFailedException ||
-                                              ((ex as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false))
+            catch (Exception ex) when (
+                ex is RequestFailedException ||
+                ((ex as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false))
             {
                 if (IsAuthenticationError(ex))
                 {
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -19,7 +19,6 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
-
     internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigurationRefresher
     {
         private bool _optional;
@@ -42,6 +41,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
        
         // To avoid concurrent network operations, this flag is used to achieve synchronization between multiple threads.
         private int _networkOperationsInProgress = 0;
+        private ILogger _logger;
+        private ILoggerFactory _loggerFactory;
 
         public Uri AppConfigurationEndpoint
         {
@@ -69,11 +70,26 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             }
         }
 
+        public ILoggerFactory LoggerFactory
+        {
+            get
+            {
+                return _loggerFactory;
+            }
+            set
+            {
+                _loggerFactory = value;
+                _logger = _loggerFactory?.CreateLogger(LoggingConstants.AppConfigLogCategory);
+            }
+        }
+
         public AzureAppConfigurationProvider(ConfigurationClient client, AzureAppConfigurationOptions options, bool optional)
         {
             _client = client ?? throw new ArgumentNullException(nameof(client));
             _options = options ?? throw new ArgumentNullException(nameof(options));
             _optional = optional;
+            _loggerFactory = options.LoggerFactory;
+            _logger = _loggerFactory?.CreateLogger(LoggingConstants.AppConfigLogCategory);
 
             IEnumerable<KeyValueWatcher> watchers = options.ChangeWatchers.Union(options.MultiKeyWatchers);
 
@@ -177,7 +193,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             }
         }
 
-        public async Task<bool> TryRefreshAsync(ILogger logger)
+        public async Task<bool> TryRefreshAsync()
         {
             try
             {
@@ -187,7 +203,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             {
                 if (e.Status == (int)HttpStatusCode.Unauthorized || e.Status == (int)HttpStatusCode.Forbidden)
                 {
-                    logger?.LogError(e, "Refresh operation failed due to an authentication error");
+                    _logger?.LogError(e, "Refresh operation failed due to an authentication error.");
                 }
 
                 return false;
@@ -197,14 +213,14 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 if (e.InnerExceptions.Any(exception => (exception is RequestFailedException ex) 
                                                         && (ex.Status == (int)HttpStatusCode.Unauthorized || ex.Status == (int)HttpStatusCode.Forbidden)))
                 {
-                    logger?.LogError(e, "Refresh operation failed due to an authentication error.");
+                    _logger?.LogError(e, "Refresh operation failed due to an authentication error.");
                 }
 
                 return false;
             }
             catch (KeyVaultReferenceException e)
             {
-                logger?.LogError(e, "Refresh operation failed while resolving a Key Vault reference.");
+                _logger?.LogError(e, "Refresh operation failed while resolving a Key Vault reference.");
                 return false;
             }
             catch (OperationCanceledException)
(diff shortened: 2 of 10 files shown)


jimmyca15:
Instead of having ILogger as an argument of TryRefresh, can we have ILogger registered on our configuration provider?

avanigupta:
> Instead of having ILogger as an argument of TryRefresh, can we have ILogger registered on our configuration provider?

Our configuration provider is created during app startup but dependency injection takes place after startup has completed. So `AzureAppConfigurationProvider` will not be able to get `ILogger` from DI. Or maybe I'm missing something? What is the benefit of registering `ILogger` with config provider instead of as an argument of `TryRefreshAsync`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -176,17 +177,37 @@ public async Task RefreshAsync()
             }
         }
 
-        public async Task<bool> TryRefreshAsync()
+        public async Task<bool> TryRefreshAsync(ILogger logger)
```
jimmyca15, line 180: Since we're breaking the interface it's probably a good idea to add cancellation token.

jimmyca15:
It is a bit of a chicken and egg problem in the context of asp.net core dependency injection. Passing ILogger to each invocation of TryRefreshAsync just seems a bit odd to me. I expect ILogger to be more of a state of the refresher object than an argument to operations on the refresher.

Having a settable logger property on IConfigurationRefresher would seem more reasonable to me, but having it on the provider itself sounds even better.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Will add cancellation token in a separate PR.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -115,6 +116,22 @@ internal IEnumerable<IKeyValueAdapter> Adapters
         /// </summary>
         internal bool IsOfflineCacheConfigured { get; private set; } = false;
 
+        /// <summary>
+        /// The <see cref="ILoggerFactory"/> for creating a logger to log errors during refresh operation.
+        /// </summary>
+        internal ILoggerFactory LoggerFactory { get; set; }
+
+        /// <summary>
+        /// Register a logger factory to log errors during refresh operations.
```
jimmyca15, line 125: I would opt for generalizing the description a bit more. We may log other things in the future, like startup failures.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationRefresher.cs
```diff
@@ -17,6 +18,11 @@ public interface IConfigurationRefresher
         /// </summary>
         Uri AppConfigurationEndpoint { get; }
 
+        /// <summary>
+        /// An <see cref="ILoggerFactory"/> for creating logger to log errors during refresh operations.
```
jimmyca15, line 22: Perhaps "creating **a** logger" ?

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -1103,15 +1104,15 @@ static void optionsInitializer(AzureAppConfigurationOptions options)
                 .AddAzureAppConfiguration(optionsInitializer, optional: true)
                 .Build();
 
-            IConfigurationRefresherProvider refresherProvider = new AzureAppConfigurationRefresherProvider(configuration);
+            IConfigurationRefresherProvider refresherProvider = new AzureAppConfigurationRefresherProvider(configuration, NullLoggerFactory.Instance);
```
jimmyca15, line 1107: If doing multi-line params it's usually better to put all params on new line otherwise the first param ends up as an odd one out.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal class LoggingConstants
+    {
+        public const string AppConfigLogCategory = "Microsoft.Extensions.Configuration.AzureAppConfiguration";
```
jimmyca15, line 8: I'm thinking we may want `Microsoft.Extensions.Configuration.AzureAppConfiguration.Refresh` as the refresh category.

jimmyca15: It allows us to add more logs and for the customer to selectively configure refresh logs vs our others.

jimmyca15:
Your PR description is great! The one thing I would add is to the part about dependency injection to call out that the logger factory will be wired up automatically as long as services.AddAzureAppConfiguration is called.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -115,6 +116,22 @@ internal IEnumerable<IKeyValueAdapter> Adapters
...
+        /// </summary>
+        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> for creating a logger.</param>
+        /// <returns></returns>
+        public AzureAppConfigurationOptions RegisterLoggerFactory(ILoggerFactory loggerFactory)
```
jimmyca15, line 129: `RegisterLoggerFactory`. Where does the naming come from? I know we have existing set-like APIs with `SetSecretResolver`.

avanigupta: I didn't use `SetLoggerFactory` because we discussed that that might be confusing as `Set<property>` is the setter format in Java. But I do like it more than `RegisterLoggerFactory` because we have other `Set` APIs like your pointed out.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
avanigupta: In that case, we might need to create two different ILogger instances during startup - one for startup logs and one for refresh logs. I think it might be better to indicate refresh vs startup logs via the error message so that we don't have to manage two different loggers. What do you think?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
jimmyca15: Yeah, I think `SetSecretResolver` has that same problem, but now that I'm seeing it again, it looks like consistency is the way to go.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
jimmyca15: Loggers are configured by their category name. Error message won't work.

Example: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-5.0#configure-logging 

In that example they configure the hosting lifetime category individually.

avanigupta: Updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: Updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -182,11 +199,31 @@ public async Task<bool> TryRefreshAsync()
...
-            catch (Exception e) when (
-                e is KeyVaultReferenceException ||
-                e is RequestFailedException ||
-                ((e as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
-                e is OperationCanceledException)
+            catch (RequestFailedException e)
+            {
+                if (e.Status == (int)HttpStatusCode.Unauthorized || e.Status == (int)HttpStatusCode.Forbidden)
+                {
+                    _logger?.LogError(e, "Refresh operation failed due to an authentication error.");
```
zhenlan, line 206: I am wondering if these should be logged as Warning. Single occurrence is not the end of day.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -182,11 +199,31 @@ public async Task<bool> TryRefreshAsync()
...
+                }
```
zhenlan, line 207: Should we log something in the `else` or we are missing a failure case here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -182,11 +199,31 @@ public async Task<bool> TryRefreshAsync()
...
+
+                return false;
+            }
+            catch (AggregateException e) when (e?.InnerExceptions?.All(e => e is RequestFailedException) ?? false)
+            {
+                if (e.InnerExceptions.Any(exception => (exception is RequestFailedException ex) 
+                                                        && (ex.Status == (int)HttpStatusCode.Unauthorized || ex.Status == (int)HttpStatusCode.Forbidden)))
+                {
+                    _logger?.LogError(e, "Refresh operation failed due to an authentication error.");
+                }
```
zhenlan, line 217: Log something in `else` condition?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: But it's still an error that led to the failure of the current refresh operation. Warning might suggest that refresh succeeded with some transient errors.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: I think the [original plan](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/113#issuecomment-582167408) was to log only authentication/Key Vault errors for which users might need to take some action. Everything else is transient and might be resolved with the next refresh operation.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -68,11 +70,26 @@ public Uri AppConfigurationEndpoint
             }
         }
 
+        public ILoggerFactory LoggerFactory
+        {
+            get
+            {
+                return _loggerFactory;
+            }
+            set
```
zhenlan, line 79: I don't see anyone calling the setter. Is it needed?

zhenlan: Never mind. It's used by the refresher.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -115,6 +116,22 @@ internal IEnumerable<IKeyValueAdapter> Adapters
...
+        /// The <see cref="ILoggerFactory"/> for creating a logger to log errors.
+        /// </summary>
+        internal ILoggerFactory LoggerFactory { get; set; }
+
+        /// <summary>
+        /// Set a logger factory to log errors.
+        /// </summary>
+        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> for creating a logger.</param>
+        /// <returns></returns>
+        public AzureAppConfigurationOptions SetLoggerFactory(ILoggerFactory loggerFactory)
```
zhenlan, line 129: We don't do any logging for startup. I'm inclined to not introduce this API at all - only support obtaining logger from DI. We can introduce this API in the future if there is a need.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresher.cs
```diff
@@ -9,18 +10,36 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     internal class AzureAppConfigurationRefresher : IConfigurationRefresher
     {
         private AzureAppConfigurationProvider _provider = null;
+        private ILoggerFactory _loggerFactory;
```
zhenlan, line 13: Can we always get/set the loggerFactory from/to the provider instead of keeping a local copy? I'm losing track of which one is the source of truth.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationRefresher.cs
```diff
@@ -17,6 +18,11 @@ public interface IConfigurationRefresher
         /// </summary>
         Uri AppConfigurationEndpoint { get; }
 
+        /// <summary>
+        /// An <see cref="ILoggerFactory"/> for creating a logger to log errors.
+        /// </summary>
+        ILoggerFactory LoggerFactory { get; set; }
```
zhenlan, line 24: To avoid introducing this new property (a breaking change), we can cast the `IConfigurationRefresher` to `AzureAppConfigurationRefresher` in the refresherProvider.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresher.cs
avanigupta: Good idea, I'll update that

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: Okay, I can remove it. And if someone is not using DI, they can always call GetRefresher() and set the LoggerFactory on that refresher instance manually.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationRefresher.cs
avanigupta: Do you mean cast `IConfigurationRefresher` to `AzureAppConfigurationProvider`? `AzureAppConfigurationRefresherProvider`  loops over all `IConfigurationProvider`s . So we cannot cast it to `AzureAppConfigurationRefresher`.

avanigupta: This is possible, but I'm not sure if its the best thing to do. 
Are we saying that `AzureAppConfigurationRefresherProvider` needs to know about a concrete type (`AzureAppConfigurationProvider`) in addition to an interface (`IConfigurationRefresher`)?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Models;
+using Microsoft.Extensions.Logging;
```
zhenlan, line 10: No need any more.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: Yes, we agreed everything could be transient including auth issues. Customers want to know when "transient" errors happened so they can react if necessary. We log details when we know better, but we should still log when we don't have details.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: As discussed in the other comment, we consider all these failures transient. I'm putting myself in the shoes of the customer. If we use the config provider in our service, I will not want to see a level 4 error when this occurs just once. I may set up a monitor when this occurs repeatedly. So warning sounds more appropriate to me.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationRefresher.cs
zhenlan: It's not a problem for one concrete type to know the other concrete type.

However, you brought up a good point in the other comment. This can be useful for scenarios DI is not used. So I'm good to keep this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Added a warning for non-auth issues too.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -182,11 +197,39 @@ public async Task<bool> TryRefreshAsync()
...
+                return false;
+            }
+            catch (KeyVaultReferenceException e)
+            {
+                _logger?.LogWarning(e, "Refresh operation failed while resolving a Key Vault reference.");
+                return false;
+            }
+            catch (OperationCanceledException)
             {
                 return false;
```
zhenlan, line 234: Missing a log?

avanigupta: Since we don't support cancellation token in `TryRefreshAsync` flow, we shouldn't be handling `OperationCanceledException` in the first place. I was on the fence about adding a log here or adding the log later when we work on cancellation token support. But I guess it doesn't hurt to add it now.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -182,12 +197,41 @@ public async Task<bool> TryRefreshAsync()
...
-            catch (Exception e) when (
-                e is KeyVaultReferenceException ||
-                e is RequestFailedException ||
-                ((e as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
-                e is OperationCanceledException)
+            catch (RequestFailedException e)
+            {
+                if (e.Status == (int)HttpStatusCode.Unauthorized || e.Status == (int)HttpStatusCode.Forbidden)
+                {
+                    _logger?.LogWarning(e, "Refresh operation failed due to an authentication error.");
```
jimmyca15, line 204: I'd suggest "A refresh operation failed ..." for a complete sentence.

jimmyca15: Also, looks like we have duplicated strings. Probably want to centralize those.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -182,12 +197,41 @@ public async Task<bool> TryRefreshAsync()
...
                 await RefreshAsync().ConfigureAwait(false);
             }
-            catch (Exception e) when (
-                e is KeyVaultReferenceException ||
-                e is RequestFailedException ||
-                ((e as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
-                e is OperationCanceledException)
+            catch (RequestFailedException e)
+            {
+                if (e.Status == (int)HttpStatusCode.Unauthorized || e.Status == (int)HttpStatusCode.Forbidden)
```
jimmyca15, line 202: We may want to have `bool IsAuthenticationError(Exception e)` for `e.Status == (int)HttpStatusCode.Unauthorized || e.Status == (int)HttpStatusCode.Forbidden` since we use it in Aggregate case and RFE case.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -678,5 +677,10 @@ private DateTimeOffset AddRandomDelay(DateTimeOffset dt, TimeSpan maxDelay)
             long randomTicks = (long)(maxDelay.Ticks * RandomGenerator.NextDouble());
             return dt.AddTicks(randomTicks);
         }
+
+        private bool IsAuthenticationError(RequestFailedException ex)
```
jimmyca15, line 681: What about a pattern that allows the catch blocks to be identical.
```
bool IsAuthenticationError(Exception e)
{
  if (e is RequestFailedException rfe)
  {
    return ex.Status == (int)HttpStatusCode.Unauthorized || ex.Status == (int)HttpStatusCode.Forbidden;
  }

  if (e is AggregateException ae)
  {
    return ae.InnerExceptions?.Any(inner => IsAuthenticationError(inner));
  }

  return false;
}
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    internal class LoggingConstants
+    {
+        // Categories
+        public const string AppConfigRefreshLogCategory = "Microsoft.Extensions.Configuration.AzureAppConfiguration.Refresh";
+
+        // Error messages
+        public const string RefreshFailedDueToAuthenticationError = "A refresh operation failed due to an authentication error.";
+        public const string RefreshFailedDueToKeyVaultError = "A refresh operation failed while resolving a Key Vault reference.";
+        public const string RefreshFailedError = "A refresh operation failed due to an error.";
```
jimmyca15, line 14: In a failure scenario, the presence of an error is implicit. I usually see a message like "A refresh operation failed." and omit the "due to an error part".

jimmyca15 (approved):

zhenlan (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Updated, thanks!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -182,12 +197,29 @@ public async Task<bool> TryRefreshAsync()
             {
                 await RefreshAsync().ConfigureAwait(false);
             }
-            catch (Exception e) when (
-                e is KeyVaultReferenceException ||
-                e is RequestFailedException ||
-                ((e as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
-                e is OperationCanceledException)
+            catch (Exception ex) when (
```
jimmyca15, line 200: I'd like to keep the change you had where we aren't catching base `Exception`. It's generally a bad practice to catch base exception and do anything with it, even a `when` check. For example stackoverflowexception would be valid here, but Ideally we'd like that to bubble up asap.

The `when` clause is good if the type of the exception alone is not good enough to determine if you want to enter the catch block, but I'd consider using when to affirm the type a bit of an abuse.

jimmyca15: In this case it will result in two identical catch blocks, but there's not much of a choice with how ConfigurationClient propagates errors.

avanigupta: Updated.

jimmyca15 (approved):

PR #278, opened 2021-10-12
Ensure users get the most up to date configurations from App Configuration with sync tokens

MBSolomon:
### Problem
Replication lag can occur when users request key-values from their configuration stores. This can occur after a key-value is modified and before the replication of the modification to other regions is completed. During this time users can experience an issue where the requested key-value has not been updated yet in the requested region – Which would result in outdated key-values being sent to a user’s applications.  

### Proposed Solution
Through the integration of a `SyncToken` into the read call, the user can be certain they are receiving the most up to date configurations in their application. With the new integration, the user will also send a `SyncToken` to the Azure App Configuration SDK which then gets passed along to cosmos DB to ensure the most recent configurations are returned. For this we implement a new `PushNotification` class add two new API's:

**PushNotification**
```cs
public class PushNotification() {
     public Uri ResourceUri;
     public string SyncToken;
     public string EventType;
}
```

**TryParseJson**
This is a method in the `EventGridEventParser` class which tries to parse message data from EventGrid and populate a `PushNotification` object.
```cs
EventGridEventParser.TryParseJson(string message, out PushNotification pushNotification)
```

**ProcessPushNotification**
This method is called by the refresher and provider to process the `PushNotification`. `ProcessPushNotification()` will invoke `UpdateSyncToken()` and then invoke `SetDirty()` to ensure the most recent configuration is provided to the user.
```cs
ProcessPushNotification(PushNotification pushNotification, TimeSpan? maxDelay)
```

### Usage Example
First you must get the Body of the response message from EventGrid and invoke `TryParseJson()` to populate the `PushNotification`. Then you invoke `ProcessPushNotification()` to process the `PushNotification` and update the `SyncToken`.
```cs
serviceBusClient.RegisterMessageHandler(handler: (message, cancellationToken) => 
{ 
		string messageText = Encoding.UTF8.GetString(message.Body); 

  EventGridEventParser.TryParseJson(messageText, out PushNotification pushNotification); 
	     _refresher.ProcessPushNotification(pushNotification); 
}); 
```


examples/ConsoleApplication/Program.cs
```diff
@@ -98,4 +98,4 @@ private static async Task Run(CancellationToken token)
             }
         }
     }
-}
+}
```
avanigupta, line 101: Same as above, undo files where only whitespace changed.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,335 @@
...
+
+			string badEventType = "{\"id\":\"id-value1\",\"topic\":\"/subscriptions/subscription-value1/resourceGroups/resourceGroup1/providers/provider1/configurationstores/store1\",\"subject\":\"https://store1.resource.io/kv/searchQuery1\",\"data\":{\"key\":\"searchQuery1\",\"etag\":\"etagValue1\",\"syncToken\":\"syncToken1;sn=001\"},\"eventType\":\"eventType.KeyValue\",\"dataVersion\":\"2\",\"metadataVersion\":\"1\",\"eventTime\":\"2021-10-06T20:08:07.2536025Z\"}";
+			string noEventTypeMsg   = "{\"id\":\"id-value1\",\"topic\":\"/subscriptions/subscription-value1/resourceGroups/resourceGroup1/providers/provider1/configurationstores/store1\",\"subject\":\"https://store1.resource.io/kv/searchQuery1\",\"data\":{\"key\":\"searchQuery1\",\"etag\":\"etagValue1\",\"syncToken\":\"syncToken1;sn=001\"},\"dataVersion\":\"2\",\"metadataVersion\":\"1\",\"eventTime\":\"2021-10-06T20:08:07.2536025Z\"}";
+
+			PushNotification pushNotification = new PushNotification();
```
avanigupta, line 235: No need to initialize a new object for out parameter, you can change the first TryParseJson call to:
`Assert.False(EventGridEventParser.TryParseJson(emptyMessage, out PushNotification pushNotification));`

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,335 @@
...
+			{
+				refresher.ProcessPushNotification(pushNotification, TimeSpan.FromSeconds(0));
+				refresher.RefreshAsync().Wait();
+			}
+
+			mockClient.Verify(c => c.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()), Times.Exactly(8));
+			mockClient.Verify(c => c.UpdateSyncToken(It.IsAny<string>()), Times.Exactly(8));
+		}
+
+		private bool IsPushNotificationNull(PushNotification pn)
```
avanigupta, line 293: This method is not just checking if pushnotification is null. Its also checking if any of its properties are null. So I suggest renaming it to something like `IsPushNotificationInvalid`.

avanigupta: On second thought, `IsPushNotificationInvalid` has negative semantics since it returns 'true' when something is 'invalid'. The better approach would be to use an affirmative phrase like `IsPushNotificationValid` and change the method logic accordingly.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/EventGridEventParser.cs
```diff
@@ -0,0 +1,51 @@
...
+						SyncToken = jsonMessage.GetProperty("data").GetProperty("syncToken").GetString(),
+						EventType = jsonMessage.GetProperty("eventType").GetString(),
+						ResourceUri =  new Uri(jsonMessage.GetProperty("subject").GetString())};
+
+				return true;
+			}
+			catch (JsonException) { }
+			catch (KeyNotFoundException) { }
+			catch (ArgumentException) { }
+			catch (InvalidOperationException) { }
```
avanigupta, line 45: We should catch `UriFormatException` too ([reference](https://docs.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=netstandard-2.0#System_Uri__ctor_System_String_)).

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,316 @@
...
+					{
+						refreshOptions.Register("TestKey1", "label")
+							.SetCacheExpiration(TimeSpan.FromDays(30));
+					});
+					refresher = options.GetRefresher();
+				})
+				.Build();
+
+			//Create list of PushNotifications with null Parameters
+			List<PushNotification> invalidPushNotifications = invalidPushNotificationList;
```
avanigupta, line 181: You can also use the `invalidPushNotificationList` directly.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,316 @@
...
+				refresher.RefreshAsync().Wait();
+			}
+
+			mockClient.Verify(c => c.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()), Times.Exactly(8));
+			mockClient.Verify(c => c.UpdateSyncToken(It.IsAny<string>()), Times.Exactly(8));
+		}
+
+		private bool IsPushNotificationValid(PushNotification pn)
+        {
+			return ((pn == null) || (pn.SyncToken == null || pn.EventType == null || pn.ResourceUri == null)) ? false : true;
```
avanigupta, line 276: How about this:
```
return pn != null && pn.SyncToken != null && pn.EventType != null && pn.ResourceUri != null;
```

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,316 @@
...
+                {
+                    refresher.ProcessPushNotification(invalidPushNotification);
+                    refresher.RefreshAsync().Wait();
+
+                }
+				//Should be caught in this block and continue to the next invalidNotification
+                catch (ArgumentNullException) { Assert.True(true); continue; }
+                catch (ArgumentException) { Assert.True(true); continue; }
+
+				//ProcessPushNotification did not throw any errors
```
avanigupta, line 195: We shouldn't try/catch in unit tests - use `Assert.Throws` to verify which exception is thrown.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,302 @@
...
+			foreach (PushNotification pushNotification in pushNotifications)
+			{
+				refresher.ProcessPushNotification(pushNotification, TimeSpan.FromSeconds(0));
+				refresher.RefreshAsync().Wait();
+			}
+
+			mockClient.Verify(c => c.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()), Times.Exactly(8));
+			mockClient.Verify(c => c.UpdateSyncToken(It.IsAny<string>()), Times.Exactly(8));
+		}
+
```
avanigupta, line 259: Can you a test for validating that push refresh flow actually refreshes configuration? Set a high cache expiration time, process push notification, verify that next RefreshAsync call updates config.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/EventGridEventParser.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+	/// <summary>
+	/// EventGridEventParser contains the method to parse a Json EventGrid Message Notification
```
avanigupta, line 8: How about this:
"This class provides a helper method to parse an Event Grid event."

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
```diff
@@ -0,0 +1,59 @@
+using Microsoft.Azure.EventGrid.Models;
```
avanigupta, line 1: Please add license header like other files.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Text.Json;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+	/// <summary>
+	/// 
+	/// </summary>
+	public static class EventGridEventExtensions
+    {
+		private static string dataSyncToken = "syncToken";
```
avanigupta, line 15: We can define `private const string SyncTokenPropertyName = "syncToken";`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+
+			if (eventGridEvent == null || eventGridEvent.Data == null || eventGridEvent.EventType == null || eventGridEvent.Subject == null)
+			{
+				return false;
+			}
+
+			try
+			{
+				string eventGridData = eventGridEvent.Data.ToString();
+				JsonElement jsonMessage = JsonDocument.Parse(eventGridData).RootElement;
```
avanigupta, line 35: These two statements can be combined into:
`JsonElement jsonMessage = JsonDocument.Parse(eventGridEvent.Data.ToString()).RootElement;`

avanigupta: Actually, we can combine all 3:

`string syncToken = JsonDocument.Parse(eventGridEvent.Data.ToString()).RootElement.GetProperty(dataSyncToken).GetString();`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+			}
+			catch (JsonException) { }
+			catch (ArgumentNullException) { }
+			catch (DecoderFallbackException) { }
+			catch (ArgumentException) { }
+			catch (AmbiguousMatchException) { }
+			catch (NullReferenceException) { }
+			catch (InvalidOperationException) { }
+			catch (KeyNotFoundException) { }
+			catch (UriFormatException) { }
```
avanigupta, line 57: Why are we catching `DecoderFallbackException`, `AmbiguousMatchException`, `NullReferenceException`?
The only possible exceptions come from the following calls:
1. eventGridEvent.Data.ToString() - [BinaryData.ToString()](https://docs.microsoft.com/en-us/dotnet/api/system.binarydata.tostring?view=dotnet-plat-ext-5.0) does not throw any exception.
2. [JsonDocument.Parse(string)](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsondocument.parse?view=net-5.0#System_Text_Json_JsonDocument_Parse_System_String_System_Text_Json_JsonDocumentOptions_) only throws `JsonException` and `ArgumentException`. We should catch both.
3. [JsonElement.GetProperty(string)](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.getproperty?view=net-5.0#System_Text_Json_JsonElement_GetProperty_System_String_)  only throws `InvalidOperationException` and `KeyNotFoundException` in addition to the two in my previous point. We should catch both. No need to catch `ArgumentNullException` and `ObjectDisposedException` because we already caught their base exceptions.
4. [JsonElement.GetString()](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.getstring?view=net-5.0) - no new exceptions other than the ones already discussed above.
5. [Uri(String)](https://docs.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=net-5.0#System_Uri__ctor_System_String_) throws `UriFormatException`

So these are the only 5 exceptions that we should catch here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+				JsonElement jsonMessage = JsonDocument.Parse(eventGridData).RootElement;
+				string syncToken = jsonMessage.GetProperty(dataSyncToken).GetString();
+
+				pushNotification = new PushNotification()
+				{
+					SyncToken = syncToken,
+					EventType = eventGridEvent.EventType,
+					ResourceUri = new Uri(eventGridEvent.Subject)
+				};
+				return true;
```
avanigupta, line 47: Nit: can you leave a line after `pushNotification` initialization block?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+using System.Text.Json;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+	/// <summary>
+	///
```
avanigupta, line 15: Missing summary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
MBSolomon: Shouldn't we catch the most specific exception for each case? So including the `ObjectDisposedException` and `ArgumentNullException`?
The first three you mentioned I checked again and have removed them

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+
+using Microsoft.Azure.EventGrid.Models;
```
avanigupta, line 5: This is a legacy package. We should instead use the [Azure.Messaging.EventGrid](https://www.nuget.org/packages/Azure.Messaging.EventGrid/4.7.0) package. Also, please add the package reference in the .csproj file.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
avanigupta: Yes, we should catch the most specific exception. But in this case, ObjectDisposedException and ArgumentNullException will always be caught by the catch blocks of InvalidOperationException and ArgumentException respectively. So adding (or not adding) separate catch blocks for ObjectDisposedException and ArgumentNullException will have no effect.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -252,6 +252,24 @@ public void SetDirty(TimeSpan? maxDelay)
...
         }
 
+        public void ProcessPushNotification(PushNotification notification, TimeSpan? maxDelay)
+        {
+            
+            if (notification == null) 
+            { 
+                throw new ArgumentNullException(nameof(notification)); 
+            }
+            if (notification.SyncToken == null || notification.EventType == null || notification.ResourceUri == null)
```
avanigupta, line 262: Nit: leave a line after the closing bracket `}`. Can you update this in other places too?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/PushNotification.cs
```diff
@@ -0,0 +1,28 @@
...
+		/// </summary>
+		public Uri ResourceUri { get; set; }
+
+        /// <summary>
+        /// The Synchronization Token to be added to the next request to the App Configuration Service.
+        /// </summary>
+        public string SyncToken { get; set; }
+
+        /// <summary>
+        /// Event Type of Event Grid Message
```
avanigupta, line 23: >Event Grid Message [](http://example.com/codeflow?start=26&length=18)

The PushNotification object is decoupled from Event Grid events. We could have a different event source in future, so we shouldn't talk about event grid (or its property names like 'subject') in this class. We should just summarize what this property is in AppConfig terms.
For example:

ResourceUri -> The URI of the key-value that triggered the push notification.

EventType -> The type of event that triggered the push notification.

Class summary -> An object containing the details of a push notification received from App Configuration service.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/PushNotification.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+
```
avanigupta, line 2: Missing license header

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,313 @@
+using Azure;
+using Azure.Core.Testing;
```
avanigupta, line 2: License header

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+	/// </summary>
+	public static class EventGridEventExtensions
+    {
+		private const string SyncTokenPropertyName = "syncToken";
+
+		/// <summary>
+		/// This method uses an EventGridEvent from EventGrid and tries to create a <see cref="PushNotification"/>
+		/// </summary>
+		/// <param name="eventGridEvent"> EventGridEvent from EventGrid</param>
+		/// <param name="pushNotification"> out parameter set up in this method</param>
```
avanigupta, line 25: >/// <param name="pushNotification"> out parameter set up in this method</param> [](http://example.com/codeflow?start=2&length=79)

Suggested parameter summary for pushNotification:
"If this method returns true, the pushNotification object contains details populated from eventGridEvent. If this method returns false, the pushNotification object is null."

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+	/// <summary>
+	/// 
+	/// </summary>
+	public static class EventGridEventExtensions
+    {
+		private const string SyncTokenPropertyName = "syncToken";
+
+		/// <summary>
+		/// This method uses an EventGridEvent from EventGrid and tries to create a <see cref="PushNotification"/>
```
avanigupta, line 22: >/// This method uses an EventGridEvent from EventGrid and tries to create a <see cref="PushNotification"/> [](http://example.com/codeflow?start=2&length=106)

Suggested method summary:
"Tries to create the <see cref="PushNotification"/> object from the details in <see cref="EventGridEvent"/>object. Return value indicates whether the operation succeeded or failed."

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,313 @@
...
+        List<PushNotification> invalidPushNotificationList = new List<PushNotification>
+            {
+              new PushNotification  {
+                                    ResourceUri = null,
+                                    EventType = "eventType.KeyValueModified",
+                                    SyncToken = "SyncToken1;sn=001"
+                                    },
+              new PushNotification  {
+                                    ResourceUri = new Uri("https://store2.resource.io/kv/searchQuery2"),
+
```
avanigupta, line 111: Nit: remove extra line

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,313 @@
...
+                                    ResourceUri = null,
+                                    EventType = null,
+                                    SyncToken = null
+                                    }
+            };
+
+		ConfigurationSetting FirstKeyValue => _kvCollection.First();
+
+        [Fact]
+        public void TryTryCreatePushNotification()
```
avanigupta, line 136: Typo: `Try` is added twice to test name

avanigupta: Also, we now have a public method with the same name so we shouldn't name the test like this. Can you rename this test accordingly?

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,313 @@
...
+				EventType = "eventType.KeyValueModified",
+				Id = "some id",
+				Subject = "https://store1.resource.io/kv/searchQuery1",
+				Topic = "/subscriptions/subscription-value1/resourceGroups/resourceGroup1/providers/provider1/configurationstores/store1\\"
+			};
+
+			eventGridEvent.TryCreatePushNotification(out PushNotification pushNotification);
+
+			Assert.Equal(eventGridEvent.EventType, pushNotification.EventType);
+			Assert.Equal(new Uri(eventGridEvent.Subject), pushNotification.ResourceUri);
```
avanigupta, line 163: >new Uri(eventGridEvent.Subject) [](http://example.com/codeflow?start=16&length=31)

You can use [Uri.OriginalString]( [https://docs.microsoft.com/en-us/dotnet/api/system.uri.originalstring?view=net-5.0](https://docs.microsoft.com/en-us/dotnet/api/system.uri.originalstring?view=net-5.0) ) to comapre the strings, instead of creating a new Uri object.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,313 @@
...
+			Assert.Equal("syncToken1;sn=001", pushNotification.SyncToken);
+
+			//Should Fail assertions since bad EventGridEvent
+			badEventGridEventDataCapitalization.TryCreatePushNotification(out pushNotification);
```
avanigupta, line 167: >badEventGridEventDataCapitalization.TryCreatePushNotification(out pushNotification); [](http://example.com/codeflow?start=3&length=84)

This can also be wrapped in 
`Assert.False(badEventGridEventDataCapitalization.TryCreatePushNotification(out pushNotification);)`

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,313 @@
...
+				.Build();
+
+			foreach (PushNotification pushNotification in pushNotificationList)
+			{
+				refresher.ProcessPushNotification(pushNotification, TimeSpan.FromSeconds(0));
+				refresher.RefreshAsync().Wait();
+			}
+
+			mockClient.Verify(c => c.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()), Times.Exactly(8));
+			mockClient.Verify(c => c.UpdateSyncToken(It.IsAny<string>()), Times.Exactly(8));
```
avanigupta, line 235: Instead of hardcoding 8, we can check if its equal to the length of pushNotificationList. In future, if we add or remove something from the list, this test should still pass.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
MBSolomon: It was intentionally named `TryTryCreatePushNotification`. Since we are trying the `TryCreatePushNotification()` Method. I could rename it to `TryCreatePushNotificationPopulatesNotification()` or `CheckTryCreatePushNotification()` maybe?

examples/ConfigStoreDemo/ConfigStoreDemo.csproj
```diff
@@ -6,6 +6,9 @@
   <ItemGroup>
     <DotNetCliToolReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Tools" Version="2.0.2" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.7.0" />
+  </ItemGroup>
```
avanigupta, line 11: We dont need to add this package in any of the sample projects

src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
```diff
@@ -14,6 +14,11 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.7.0" />
```
avanigupta, line 17: ><PackageReference Include="Azure.Messaging.EventGrid" Version="4.7.0" /> [](http://example.com/codeflow?start=4&length=72)

EventGrid package dependency not required in AspNetCore project

src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
```diff
@@ -14,6 +14,11 @@
...
+    <PackageReference Include="Microsoft.Extensions.Configuration.AzureAppConfiguration" Version="4.5.0" />
```
avanigupta, line 18: ><PackageReference Include="Microsoft.Extensions.Configuration.AzureAppConfiguration" Version="4.5.0" /> [](http://example.com/codeflow?start=4&length=103)

Remove reference to 4.5.0 Microsoft.Extensions.Configuration.AzureAppConfiguration nuget package. We now refer to the project directly (see ProjectReference below this).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1,736 +1,755 @@
...
+                changeWatcher.CacheExpires = cacheExpires;
+            }
+
+            foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
+            {
+                changeWatcher.CacheExpires = cacheExpires;
+            }
+        }
+
+        public void ProcessPushNotification(PushNotification notification, TimeSpan? maxDelay)
```
avanigupta, line 255: >notification [](http://example.com/codeflow?start=61&length=12)

Nit: can you rename the argument to `pushNotification` for consistency between this file, AzureAppConfigurationRefresher.cs and IConfigurationRefresher.cs?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
```diff
@@ -14,8 +14,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Data.AppConfiguration" Version="1.0.0" />
+    <PackageReference Include="Azure.Data.AppConfiguration" Version="1.2.0" />
+    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.7.0" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.0.1" />
+    <PackageReference Include="Microsoft.Azure.EventGrid" Version="3.2.1" />
```
avanigupta, line 20: This is a legacy package and should not be used. Everything we need is available in Azure.Messaging.EventGrid package.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+				{
+					SyncToken = syncToken,
+					EventType = eventGridEvent.EventType,
+					ResourceUri = new Uri(eventGridEvent.Subject)
+				};
+
+				return true;
+			}
+			catch (JsonException) { }
+			catch (ArgumentNullException) { }
```
avanigupta, line 48: >ArgumentNullException [](http://example.com/codeflow?start=10&length=21)

ArgumentNullException and AmbiguousMatchException are not required.

KeyNotFoundException is required but missing here.

tests/Tests.AzureAppConfiguration.AspNetCore/Tests.AzureAppConfiguration.AspNetCore.csproj
```diff
@@ -10,6 +10,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.7.0" />
+    <PackageReference Include="Microsoft.Azure.ServiceBus" Version="5.1.3" />
```
avanigupta, line 14: Both packages not required

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,334 @@
...
+using Azure.Data.AppConfiguration;
+using Azure.Messaging.EventGrid;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
```
avanigupta, line 15: System.Text.Json not being used in this file

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,334 @@
...
+                                    SyncToken = null
+                                    },
+              new PushNotification  {
+                                    ResourceUri = null,
+                                    EventType = null,
+                                    SyncToken = null
+                                    }
+            };
+
+		Dictionary<string, EventGridEvent> eventGridEventList = new Dictionary<string, EventGridEvent>
```
avanigupta, line 134: >eventGridEventList [](http://example.com/codeflow?start=37&length=18)

Its not a list. Also, we should avoid adding data type to the variable names. You can name it somethign like `eventGridEvents`

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,334 @@
...
+		{
+            {
+                "sn;Vxujfidne",
+            new EventGridEvent(
+                "https://store1.resource.io/kv/searchQuery1",
+                "eventType.KeyValueModified", "2",
```
avanigupta, line 140: >eventType.KeyValueModified [](http://example.com/codeflow?start=17&length=26)

Is this the actual EventType from EventGridEvent? According to our documentation, the event type should be:
`Microsoft.AppConfiguration.KeyValueModified` or `Microsoft.AppConfiguration.KeyValueDeleted`.

It doesn't matter what event type you use in this test because we dont even use it, but for accuracy and future-proofing the tests, lets use the real event types.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,334 @@
...
+                BinaryData.FromString("{\"key\":\"searchQuery1\",\"etag\":\"etagValue1\",\"syncToken\":\"sn;CRAle3342\"}")
+                )
+            }
+        };
+
+		ConfigurationSetting FirstKeyValue => _kvCollection.First();
+
+        [Fact]
+        public void CheckTryCreatePushNotification()
```
avanigupta, line 176: You can call it `ValidatePushNotificationCreation`. "CheckTry..." sounds odd.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,334 @@
...
+				})
+				.Build();
+
+			foreach (PushNotification pushNotification in pushNotificationList)
+			{
+				refresher.ProcessPushNotification(pushNotification, TimeSpan.FromSeconds(0));
+				refresher.RefreshAsync().Wait();
+			}
+
+			mockClient.Verify(c => c.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()), Times.Exactly(8));
```
avanigupta, line 255: >Times.Exactly(8) [](http://example.com/codeflow?start=141&length=16)

This is also `pushNotificationList.Count`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1,736 +1,755 @@
...
+        {
+            
+            if (notification == null) 
+            { 
+                throw new ArgumentNullException(nameof(notification)); 
+            }
+
+            if (notification.SyncToken == null || notification.EventType == null || notification.ResourceUri == null)
```
jimmyca15, line 263: The standard way of doing this is

```
if (string.IsNullOrEmpty(propA))
{
    throw new ArgumentNullException(nameof(propA));
}

if (string.IsNullOrEmpty(propB))
{
    throw new ArgumentNullException(nameof(propB));
}

if (string.IsNullOrEmpty(propC))
{
    throw new ArgumentNullException(nameof(propC));
}

//
// Continue for however many properties are required.
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+		/// This method uses an EventGridEvent from EventGrid and tries to create a <see cref="PushNotification"/>
+		/// </summary>
+		/// <param name="eventGridEvent"> EventGridEvent from EventGrid</param>
+		/// <param name="pushNotification"> out parameter set up in this method</param>
+		/// <returns></returns>
+		public static bool TryCreatePushNotification(this EventGridEvent eventGridEvent, out PushNotification pushNotification)
+		{
+			pushNotification = null;
+
+			if (eventGridEvent == null || eventGridEvent.Data == null || eventGridEvent.EventType == null || eventGridEvent.Subject == null)
```
jimmyca15, line 29: The typical expectation for extension methods is not to check for null on the object you are extending.

If we look at this method, the expected usage is
```
eventGridEvent.TryCreatePushNotification(out PushNotification pushNotification);
```

If the method was an instance method instead of an extension method that would throw a null-ref if `eventGridEvent` was null. Since extension methods are used the same way they share the same expectation.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+				{
+					SyncToken = syncToken,
+					EventType = eventGridEvent.EventType,
+					ResourceUri = new Uri(eventGridEvent.Subject)
+				};
+
+				return true;
+			}
+			catch (JsonException) { }
+			catch (ArgumentException) { }
```
jimmyca15, line 49: What throws ArgumentException?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+			catch (InvalidOperationException) { }
+			catch (KeyNotFoundException) { }
```
jimmyca15, line 51: What throws KeyNotFoundException?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+
+			if (eventGridEvent == null || eventGridEvent.Data == null || eventGridEvent.EventType == null || eventGridEvent.Subject == null)
+			{
+				return false;
+			}
+
+			try
+			{
+				string syncToken = (JsonDocument.Parse(eventGridEvent.Data.ToString()).RootElement)
+					.GetProperty(SyncTokenPropertyName).GetString();
```
jimmyca15, line 37: Can this return null?

jimmyca15: We're catching JsonException, so it looks like we're not trusting the JSON payload completely.

jimmyca15: Is it possible for EventGridEvent.Data to be invalid JSON?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
MBSolomon: the `.GetProperty()` 
[Json.GetProperty](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.getproperty?view=net-5.0)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+					ResourceUri = new Uri(eventGridEvent.Subject)
+				};
+
+				return true;
+			}
+			catch (JsonException) { }
+			catch (ArgumentException) { }
+			catch (InvalidOperationException) { }
+			catch (KeyNotFoundException) { }
+			catch (UriFormatException) { }
```
jimmyca15, line 52: Uri has TryCreate

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+					SyncToken = syncToken,
+					EventType = eventGridEvent.EventType,
+					ResourceUri = new Uri(eventGridEvent.Subject)
+				};
+
+				return true;
+			}
+			catch (JsonException) { }
+			catch (ArgumentException) { }
+			catch (InvalidOperationException) { }
```
jimmyca15, line 50: What is throwing InvalidOperationException?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
jimmyca15: Is there a ContainsKey method?

jimmyca15: Looks like there's TryGetProperty and TryGetString.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
MBSolomon: Actually we don't need this. Json.Parse() only throws an ArgumentException if the provided options are unsupported.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+using Azure.Messaging.EventGrid;
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text.Json;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+	/// <summary>
+	/// Extension of the EventGridEvent class for creating <see cref="PushNotification"/> object from an <see cref="EventGridEvent "/> object.
```
jimmyca15, line 13: > for creating <see cref="PushNotification"/> object from an <see cref="EventGridEvent "/> object.

I wouldn't limit the scope of the class like that in the summary. This is a class that offers extensions for EventGridEvents, I don't think we need to say more. The method summary on each method provides the details in the text above.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/PushNotification.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    /// <summary>
+    /// An Object containing the details of a PushNotification received from the Azure App Configuration service.
+    /// </summary>
+    public class PushNotification
+    {
+
```
jimmyca15, line 13: Unecessary new line.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/PushNotification.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// The URI of the key-value which triggered the PushNotification.
```
jimmyca15, line 15: Should be "resource" instead of key-value.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/PushNotification.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        public Uri ResourceUri { get; set; }
+
+        /// <summary>
+        /// The Synchronization Token to be added to the next request to the App Configuration Service.
+        /// </summary>
+        public string SyncToken { get; set; }
+
+        /// <summary>
+        /// The Type of Event which triggered the PushNotification.
```
jimmyca15, line 25: Should either be "push notification" or a <cref= ... >

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
```diff
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.3.0" />
+    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.7.0" />
```
jimmyca15, line 14: We reference the main project so don't we get this reference transitively?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: It makes sense if we check the passed arguments this way, but does this standard also apply to the properties of passed arguments?
Say if `notification.SyncToken` is null, does that mean its an `ArgumentNullException` or an `ArgumentException`?

jimmyca15: ArgumentNullException. ArgumentNullException is just an ArgumentException that's error condition is that the argument is null. So since ArgumentException is being considered, and the issue is that the argument is null, then ArgumentNullException would be best.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
avanigupta: I dont think there's any validation on what's inside `EventGridEvent.Data`. Users may create their own EventGridEvent object and put whatever they want in `Data`: [code](https://github.com/Azure/azure-sdk-for-net/blob/1d8f6fe0207c8f78d10fa9e8e71d823ff751e88f/sdk/eventgrid/Azure.Messaging.EventGrid/src/Customization/EventGridEvent.cs#L52)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: nameof would use the child property name at the least. `throw new ArgumentNullException(nameof(propC.ChildProperty));` Or possibly `throw new ArgumentNullException($"{nameof(propC)}.{nameof(propC.ChildProperty)}");`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
jimmyca15: In that case, need to check if the property exists on the JsonDocument

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Throwing `ArgumentNullException` looks odd to me if the actual argument (`notification` in this case) is not null.
Found [this article](https://social.technet.microsoft.com/wiki/contents/articles/51035.c-proper-use-of-argumentexception-and-argumentnullexception.aspx#:~:text=Basically%2C%20ArgumentException%20is%20deriving%20from%20SystemException%20whereas%20ArgumentNullException,ArgumentException%20is%20more%20generic%2C%20ArgumentNullException%20is%20more%20specific.) as a reference.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
avanigupta: I wasn't able to find TryGetString in the docs? 
Also, its strange that [TryGetProperty]( https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.trygetproperty?view=net-5.0#System_Text_Json_JsonElement_TryGetProperty_System_String_System_Text_Json_JsonElement__) will also throw exceptions.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: It does look like the accepted practice here is to use `ArgumentException` for null properties. I ended up on the wrong side of that at some point I suppose. Let's keep ArgumentException then, but let's still do individual validation rather than grouped.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
jimmyca15: I think I read TryGetSingle too fast.

avanigupta: I think we should be good using the TryGetProperty method. It will not throw an exception in our case because our property name is a const string "syncToken" ([code](https://github.com/dotnet/runtime/blob/b07b26dd18ae64cd4a6b091dd4a2c87aee773eec/src/libraries/System.Text.Json/src/System/Text/Json/Document/JsonElement.cs#L237))

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
MBSolomon: TryGetProperty()

MBSolomon: we may not need this since TryGetProperty() only throws InvalidOperationException if the ValueKind is not an object. Which in our case it is

examples/ConsoleApplication/ConsoleApplication.csproj
```diff
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.ServiceBus" Version="5.2.0" />
```
jimmyca15, line 10: why is this package reference added?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1,736 +1,769 @@
...
-                }
-
-                cacheExpirationTime = changeWatcher.CacheExpirationInterval.CalculateBackoffTime(changeWatcher.RefreshAttempts);
-            }
-
-            changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(cacheExpirationTime);
-        }
-    }
-}
+﻿// Copyright (c) Microsoft Corporation.
```
jimmyca15, line 1: this whole file is showing up as a change.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+
+			if (eventGridEvent.Data == null || eventGridEvent.EventType == null || eventGridEvent.Subject == null)
+			{
+				return false;
+			}
+
+			Uri resourceUri;
+
+			if (!Uri.TryCreate(eventGridEvent.Subject, UriKind.Absolute, out resourceUri))
+            {
```
jimmyca15, line 35: The indentation is off for the brackets of this if statement.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+				return false;
+            }
+
+			try
+			{
+				JsonElement eventGridEventData = JsonDocument.Parse(eventGridEvent.Data.ToString()).RootElement;
+
+				if (eventGridEventData.ValueKind == JsonValueKind.Object &&
+					eventGridEventData.TryGetProperty(SyncTokenPropertyName, out JsonElement syncTokenJson))
```
jimmyca15, line 44: Can you line up `eventGridEventData` on these two lines.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/PushNotification.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using System;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    /// <summary>
+    /// An Object containing the details of a <see cref="PushNotification"/> received from the Azure App Configuration service.
```
jimmyca15, line 9: It's a bit odd that the summary for this type references itself.

avanigupta:
> <Project Sdk="Microsoft.NET.Sdk">

Can you undo changes to this file? It only differs in whitespace.

---
Refers to: src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj:1 in b2e3a93. [](commit_id = b2e3a93e4417aa4f0bb8687bc9c00eea58543a39, deletion_comment = True)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Still need to address this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
avanigupta: This comment still needs to be addressed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridExtensions.cs
avanigupta: This comment still needs to be addressed.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,333 @@
...
+            {
+				string syncToken = eventGridAndSync.Key;
+				EventGridEvent eventGridEvent = eventGridAndSync.Value; 
+
+				eventGridEvent.TryCreatePushNotification(out PushNotification pushNotification);
+
+                Assert.NotNull(pushNotification);
+                Assert.Equal(eventGridEvent.EventType, pushNotification.EventType);
+                Assert.Equal(new Uri(eventGridEvent.Subject), pushNotification.ResourceUri);
+				Assert.Equal(syncToken, pushNotification.SyncToken);
```
avanigupta, line 187: You can use [Uri.OriginalString](https://docs.microsoft.com/en-us/dotnet/api/system.uri.originalstring?view=net-5.0)  to compare the strings, instead of creating a new Uri object.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,333 @@
...
+
+        [Fact]
+        public void ValidatePushNotificationCreation()
+        {
+			foreach (KeyValuePair<string, EventGridEvent> eventGridAndSync in eventGridEvents)
+            {
+				string syncToken = eventGridAndSync.Key;
+				EventGridEvent eventGridEvent = eventGridAndSync.Value; 
+
+				eventGridEvent.TryCreatePushNotification(out PushNotification pushNotification);
```
avanigupta, line 182: >eventGridEvent.TryCreatePushNotification(out PushNotification pushNotification); [](http://example.com/codeflow?start=4&length=80)

This can also be wrapped in 
`Assert.False(badEventGridEventDataCapitalization.TryCreatePushNotification(out pushNotification));`

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
```diff
@@ -0,0 +1,333 @@
...
+
+			ConfigurationModelFactory.ConfigurationSetting(
+				key: "TestKeyWithMultipleLabels",
+				label: "label2",
+				value: "TestValueForLabel2",
+				eTag: new ETag("bb203f2b-c113-44fc-995d-b933c2143339"),
+				contentType: "text")
+		};
+
+        List<PushNotification> pushNotificationList = new List<PushNotification>
```
avanigupta, line 60: >pushNotificationList [](http://example.com/codeflow?start=31&length=20)

We should use [_camelCase](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#camel-case) naming for all private fields.

tests/Tests.AzureAppConfiguration/PushRefreshTests.cs
avanigupta: Still need to address this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: That's strange. CodeFlow shows only the new changes.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            pushNotification = null;
+
+            if (eventGridEvent.Data == null || eventGridEvent.EventType == null || eventGridEvent.Subject == null)
+            {
+                return false;
+            }
+
+            Uri resourceUri;
+
+            if (!Uri.TryCreate(eventGridEvent.Subject, UriKind.Absolute, out resourceUri))
```
avanigupta, line 34: >out resourceUri [](http://example.com/codeflow?start=73&length=15)

Nit:  `out Uri resourceUri`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                return false;
+            }
+
+            try
+            {
+                JsonElement eventGridEventData = JsonDocument.Parse(eventGridEvent.Data.ToString()).RootElement;
+
+                if (eventGridEventData.ValueKind == JsonValueKind.Object &&
+                    eventGridEventData.TryGetProperty(SyncTokenPropertyName, out JsonElement syncTokenJson))
```
avanigupta, line 44: We can add another check here to make sure syncTokenJson.GetString() doesnt throw:

```
if (eventGridEventData.ValueKind == JsonValueKind.Object && 
                    eventGridEventData.TryGetProperty(SyncTokenPropertyName, out JsonElement syncTokenJson) && 
                    syncTokenJson.ValueKind == JsonValueKind.String)
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+                    pushNotification = new PushNotification()
+                    {
+                        SyncToken = syncTokenJson.GetString(),
+                        EventType = eventGridEvent.EventType,
+                        ResourceUri = resourceUri
+                    };
+
+                    return true;
+                }
+            }
```
avanigupta, line 54: Since `JsonDocument.Parse` is the only call that could throw `JsonException`, I'd suggest wrapping only that line in the try block. Everything else can be outside the try block to make it easier to understand in future.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    public static class EventGridEventExtensions
+    {
+        private const string SyncTokenPropertyName = "syncToken";
+
+        /// <summary>
+        /// Tries to create the object from the details in object. Return value indicates whether the operation succeeded or failed.
+        /// </summary>
+        /// <param name="eventGridEvent"> EventGridEvent from EventGrid</param>
+        /// <param name="pushNotification"> If this method returns true, the push notification object contains details populated from eventGridEvent. If this method returns false, the pushNotification object is null. </param>
```
avanigupta, line 21:
```suggestion
        /// <param name="pushNotification"> If this method returns true, the pushNotification  object contains details populated from eventGridEvent. If this method returns false, the pushNotification object is null. </param>
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    /// <summary>
+    /// This class offers extensions for EventGridEvents.
+    /// </summary>
+    public static class EventGridEventExtensions
+    {
+        private const string SyncTokenPropertyName = "syncToken";
+
+        /// <summary>
+        /// Tries to create the object from the details in object. Return value indicates whether the operation succeeded or failed.
```
avanigupta, line 18:
```suggestion
        /// Tries to create the <see cref="PushNotification"/> object from the details in <see cref="EventGridEvent"/> object. Return value indicates whether the operation succeeded or failed.
```

src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
```diff
@@ -12,8 +12,8 @@
     <PackageReleaseNotes>https://aka.ms/MicrosoftAzureAppConfigurationAspNetCoreReleaseNotes</PackageReleaseNotes>
     <PackageTags>Microsoft Azure Configuration AppConfig AppConfiguration AzureAppConfiguration aspnetcore</PackageTags>
   </PropertyGroup>
-
-  <ItemGroup>
+  
+   <ItemGroup>
```
jimmyca15, line 16: Unnecessary whitespace change

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1,736 +1,769 @@
...
+            {
+                throw new ArgumentException(
+                    "Sync token is required.",
+                    $"{nameof(pushNotification)}.{nameof(pushNotification.SyncToken)}");
+            }
+
+            if (string.IsNullOrEmpty(pushNotification.EventType))
+            {
+                throw new ArgumentException(
+                    "Event Type is required.",
```
jimmyca15, line 272: Event type

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1,736 +1,769 @@
...
+                    $"{nameof(pushNotification)}.{nameof(pushNotification.EventType)}");
+            }
+
+            if (pushNotification.ResourceUri == null)
+            {
+                throw new ArgumentException(
+                    "Resource Uri is required.",
```
jimmyca15, line 279: URI

jimmyca15 (approved):

jimmyca15 (approved):
Approving. There are only a few minor comments to address.

avanigupta (approved):

PR #280, opened 2021-10-18
Remove offline cache

avanigupta:
This PR deprecates all offline caching capabilities for the next major release.


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -353,18 +351,6 @@ await foreach (ConfigurationSetting setting in _client.GetConfigurationSettingsA
...
-                    // During startup or refreshAll scenario, we'll try to populate config from offline cache, if available
-                    cachedData = _options.OfflineCache.Import(_options);
-                    
-                    if (cachedData != null)
-                    {
-                        data = JsonSerializer.Deserialize<IDictionary<string, ConfigurationSetting>>(cachedData);
-                    }
-                }
-
-                // If we're unable to load data from offline cache, check if we need to ignore or rethrow the exception
```
jimmyca15, line 367: Looks like `data` will always be null now.

jimmyca15: So could add `ignoreFailures` to the when block.

jimmyca15 (approved):

jimmyca15:
Title is a little misleading. This is removing it completely.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: We still need to catch the specific exceptions regardless of `ignoreFailures`. If `ignoreFailures` is added to the when block filtering, we will end up throwing the exceptions to the caller since no matching catch block is present to handle those exceptions.

jimmyca15 (approved):

PR #281, opened 2021-10-18
Add support for CancellationToken during refresh operations

### LINKED ISSUE #275: Add CancellationToken support in TryRefreshAsync

avanigupta:
This PR adds a cancellation token to `RefreshAsync` and `TryRefreshAsync` methods.
Since the `IConfigurationRefresher` interface has changed, this is a breaking change. However, users will still be able to call  `RefreshAsync()` and `TryRefreshAsync()` without any parameter.

Fix #275

### DIFF
--- a/src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
+++ b/src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.AppConfiguration.AspNetCore
         {
             foreach (var refresher in Refreshers)
             {
-                _ = refresher.TryRefreshAsync(context.RequestAborted);
+                _ = refresher.TryRefreshAsync();
             }
 
             // Call the next delegate/middleware in the pipeline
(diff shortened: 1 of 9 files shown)


src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
```diff
@@ -26,7 +26,7 @@ public async Task InvokeAsync(HttpContext context)
         {
             foreach (var refresher in Refreshers)
             {
-                _ = refresher.TryRefreshAsync();
+                _ = refresher.TryRefreshAsync(context.RequestAborted);
```
jimmyca15, line 29: This is an orphaned task. I don't think we want to require it to complete within the execution window of the request.

avanigupta: Isn't `HttpContext.RequestAborted` equivalent to passing a CancellationToken? Do you mean we should check the `IsCancellationRequested` property before calling `TryRefreshAsync` with the cancellation token?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -657,7 +660,7 @@ private void ProcessChanges(IEnumerable<KeyValueChange> changes)
         private async Task CallWithRequestTracing(Func<Task> clientCall)
```
jimmyca15, line 660:
```
        private Task CallWithRequestTracing(Func<Task> clientCall)
        {
            var requestType = _isInitialLoadComplete ? RequestType.Watch : RequestType.Startup;
            return TracingUtils.CallWithRequestTracing(_requestTracingEnabled, requestType, _requestTracingOptions, clientCall);
        }
```

jimmyca15 (changes requested):

jimmyca15 (approved):

PR #283, opened 2021-10-21
Update cache expiration time if refresh operation failed

avanigupta:
When a refresh operation fails, we should also extend the cache expiration time for all refresh registered settings. This avoids making excessive requests to AppConfig service in case of failed refresh operations.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -737,7 +737,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     changeWatcher.RefreshAttempt++;
                 }
 
-                cacheExpirationTime = CalculateBackoffTime(cacheExpirationTime, changeWatcher.RefreshAttempt);
+                cacheExpirationTime = CalculateBackoffTime(changeWatcher.CacheExpirationInterval, changeWatcher.RefreshAttempt);
             }
             else
             {
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -418,87 +418,81 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
                 bool hasChanged = false;
                 KeyValueIdentifier watchedKeyLabel = new KeyValueIdentifier(watchedKey, watchedLabel);
+                bool success = false;
 
-                if (_watchedSettings.TryGetValue(watchedKeyLabel, out ConfigurationSetting watchedKv))
+                try
                 {
-                    KeyValueChange keyValueChange = default;
-                    bool success = false;
-
-                    try
+                    if (_watchedSettings.TryGetValue(watchedKeyLabel, out ConfigurationSetting watchedKv))
                     {
+                        KeyValueChange keyValueChange = default;
                         await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, RequestType.Watch, _requestTracingOptions,
                             async () => keyValueChange = await _client.GetKeyValueChange(watchedKv, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
 
                         success = true;
+
+                        // Check if a change has been detected in the key-value registered for refresh
+                        if (keyValueChange.ChangeType != KeyValueChangeType.None)
+                        {
+                            if (changeWatcher.RefreshAll)
+                            {
+                                shouldRefreshAll = true;
+                                break;
+                            }
+
+                            if (keyValueChange.ChangeType == KeyValueChangeType.Deleted)
+                            {
+                                _watchedSettings.Remove(watchedKeyLabel);
+                            }
+                            else if (keyValueChange.ChangeType == KeyValueChangeType.Modified)
+                            {
+                                _watchedSettings[watchedKeyLabel] = keyValueChange.Current;
+                            }
+
+                            hasChanged = true;
+                            ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
+                        }
                     }
-                    finally
+                    else
                     {
-                        UpdateCacheExpirationTime(changeWatcher, success);
-                    }
+                        // Load the key-value in case the previous load attempts had failed
+                        var options = new SettingSelector { LabelFilter = watchedLabel };
 
-                    // Check if a change has been detected in the key-value registered for refresh
-                    if (keyValueChange.ChangeType != KeyValueChangeType.None)
-                    {
-                        if (changeWatcher.RefreshAll)
+                        try
                         {
-                            shouldRefreshAll = true;
-                            break;
+                            await CallWithRequestTracing(async () => watchedKv = await _client.GetConfigurationSettingAsync(watchedKey, watchedLabel, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
+                        }
+                        catch (RequestFailedException e) when (e.Status == (int)HttpStatusCode.NotFound)
+                        {
+                            watchedKv = null;
                         }
 
-                        if (keyValueChange.ChangeType == KeyValueChangeType.Deleted)
-                        {
-                            _watchedSettings.Remove(watchedKeyLabel);
-                        }
-                        else if (keyValueChange.ChangeType == KeyValueChangeType.Modified)
-                        {
-                            _watchedSettings[watchedKeyLabel] = keyValueChange.Current;
-                        }
+                        success = true;
 
-                        hasChanged = true;
-                        ProcessChanges(Enumerable.Repeat(keyValueChange, 1));
+                        if (watchedKv != null)
+                        {
+                            if (changeWatcher.RefreshAll)
+                            {
+                                shouldRefreshAll = true;
+                                break;
+                            }
+
+                            hasChanged = true;
+
+                            // Add the key-value if it is not loaded, or update it if it was loaded with a different label
+                            _applicationSettings[watchedKey] = watchedKv;
+                            _watchedSettings[watchedKeyLabel] = watchedKv;
+
+                            // Invalidate the cached Key Vault secret (if any) for this ConfigurationSetting
+                            foreach (IKeyValueAdapter adapter in _options.Adapters)
+                            {
+                                adapter.InvalidateCache(watchedKv);
+                            }
+                        }
                     }
                 }
-                else
+                finally
                 {
-                    // Load the key-value in case the previous load attempts had failed
-                    var options = new SettingSelector { LabelFilter = watchedLabel };
-                    bool success = false;
-
-                    try
-                    {
-                        await CallWithRequestTracing(async () => watchedKv = await _client.GetConfigurationSettingAsync(watchedKey, watchedLabel, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
-                        success = true;
-                    }
-                    catch (RequestFailedException e) when (e.Status == (int)HttpStatusCode.NotFound)
-                    {
-                        watchedKv = null;
-                        success = true;
-                    }
-                    finally
-                    {
-                        UpdateCacheExpirationTime(changeWatcher, success);
-                    }
-
-                    if (watchedKv != null)
-                    {
-                        if (changeWatcher.RefreshAll)
-                        {
-                            shouldRefreshAll = true;
-                            break;
-                        }
-
-                        hasChanged = true;
-
-                        // Add the key-value if it is not loaded, or update it if it was loaded with a different label
-                        _applicationSettings[watchedKey] = watchedKv;
-                        _watchedSettings[watchedKeyLabel] = watchedKv;
-
-                        // Invalidate the cached Key Vault secret (if any) for this ConfigurationSetting
-                        foreach (IKeyValueAdapter adapter in _options.Adapters)
-                        {
-                            adapter.InvalidateCache(watchedKv);
-                        }
-                    }
+                    UpdateCacheExpirationTime(changeWatcher, success);
                 }
                 
                 if (hasChanged)
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -728,7 +728,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         private void UpdateCacheExpirationTime(KeyValueWatcher changeWatcher, DateTimeOffset initialLoadTime, bool operationFailed = false)
         {
-            var cacheExpirationTime = changeWatcher.CacheExpirationInterval;
+            TimeSpan cacheExpirationTime;
 
             if (operationFailed)
             {
@@ -738,6 +738,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 }
 
                 cacheExpirationTime = CalculateBackoffTime(cacheExpirationTime, changeWatcher.RefreshAttempt);
+            }
+            else
+            {
+                changeWatcher.RefreshAttempt = 0;
+                cacheExpirationTime = changeWatcher.CacheExpirationInterval;
             }
 
             changeWatcher.CacheExpires = initialLoadTime.Add(cacheExpirationTime);
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -184,6 +184,26 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     await RefreshKeyValueCollections(cancellationToken).ConfigureAwait(false);
                     await RefreshKeyValueAdapters(cancellationToken).ConfigureAwait(false);
                 }
+                catch (Exception exception) when (exception is RequestFailedException ||
+                                              ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
+                                              exception is OperationCanceledException)
+                {
+
+                    // Update the cache expiration time for all refresh registered settings before rethrowing the exception
+                    var initialLoadTime = DateTimeOffset.UtcNow;
+
+                    foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
+                    {
+                        changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
+                    }
+
+                    foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
+                    {
+                        changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
+                    }
+
+                    throw;
+                }
                 finally
                 {
                     Interlocked.Exchange(ref _networkOperationsInProgress, 0);
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -419,9 +419,9 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     continue;
                 }
 
+                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
                 bool hasChanged = false;
                 KeyValueIdentifier watchedKeyLabel = new KeyValueIdentifier(watchedKey, watchedLabel);
-                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
 
                 if (_watchedSettings.TryGetValue(watchedKeyLabel, out ConfigurationSetting watchedKv))
                 {
@@ -520,6 +520,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     continue;
                 }
 
+                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
                 IEnumerable<ConfigurationSetting> currentKeyValues;
 
                 if (changeWatcher.Key.EndsWith("*"))
@@ -539,7 +540,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     });
                 }
 
-                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
                 IEnumerable<KeyValueChange> keyValueChanges = await _client.GetKeyValueChangeCollection(
                     currentKeyValues, 
                     new GetKeyValueChangeCollectionOptions
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -184,26 +184,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     await RefreshKeyValueCollections(cancellationToken).ConfigureAwait(false);
                     await RefreshKeyValueAdapters(cancellationToken).ConfigureAwait(false);
                 }
-                catch (Exception exception) when (exception is RequestFailedException ||
-                                              ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
-                                              exception is OperationCanceledException)
-                {
-
-                    // Update the cache expiration time for all refresh registered settings before rethrowing the exception
-                    var initialLoadTime = DateTimeOffset.UtcNow;
-
-                    foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
-                    {
-                        changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
-                    }
-
-                    foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
-                    {
-                        changeWatcher.CacheExpires = initialLoadTime.Add(changeWatcher.CacheExpirationInterval);
-                    }
-
-                    throw;
-                }
                 finally
                 {
                     Interlocked.Exchange(ref _networkOperationsInProgress, 0);
@@ -441,14 +421,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
                 bool hasChanged = false;
                 KeyValueIdentifier watchedKeyLabel = new KeyValueIdentifier(watchedKey, watchedLabel);
+                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
 
                 if (_watchedSettings.TryGetValue(watchedKeyLabel, out ConfigurationSetting watchedKv))
                 {
                     KeyValueChange keyValueChange = default;
                     await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, RequestType.Watch, _requestTracingOptions,
                         async () => keyValueChange = await _client.GetKeyValueChange(watchedKv, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
-
-                    changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
 
                     // Check if a change has been detected in the key-value registered for refresh
                     if (keyValueChange.ChangeType != KeyValueChangeType.None)
@@ -486,11 +465,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                         watchedKv = null;
                     }
 
-                    changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
-
                     if (watchedKv != null)
                     {
-
                         if (changeWatcher.RefreshAll)
                         {
                             shouldRefreshAll = true;
@@ -563,6 +539,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     });
                 }
 
+                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
                 IEnumerable<KeyValueChange> keyValueChanges = await _client.GetKeyValueChangeCollection(
                     currentKeyValues, 
                     new GetKeyValueChangeCollectionOptions
@@ -574,7 +551,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     }, 
                     cancellationToken).ConfigureAwait(false);
 
-                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
 
                 if (keyValueChanges?.Any() == true)
                 {
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -38,7 +38,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         private static readonly TimeSpan MinDelayForUnhandledFailure = TimeSpan.FromSeconds(5);
         private static readonly TimeSpan DefaultMaxSetDirtyDelay = TimeSpan.FromSeconds(30);
-       
+        
+        private static readonly TimeSpan DefaultBackoffDuringRefreshErrors = TimeSpan.FromSeconds(1);
+        private static readonly TimeSpan MaxBackoffDuringRefreshErrors = TimeSpan.FromMinutes(10);
+        private const int MaxRefreshAttempts = 20;
+
         // To avoid concurrent network operations, this flag is used to achieve synchronization between multiple threads.
         private int _networkOperationsInProgress = 0;
         private ILogger _logger;
@@ -314,9 +318,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 await Task.Run(() => LoadKeyValuesRegisteredForRefresh(serverData, cancellationToken).ConfigureAwait(false).GetAwaiter().GetResult()).ConfigureAwait(false);
                 data = serverData;
             }
-            catch (Exception exception) when (exception is RequestFailedException ||
-                                              ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
-                                              exception is OperationCanceledException)
+            catch (Exception exception) when (CanHandleException(exception))
             {
                 if (_options.OfflineCache != null)
                 {
@@ -332,6 +334,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 // If we're unable to load data from offline cache, check if we need to ignore or rethrow the exception 
                 if (data == null && !ignoreFailures)
                 {
+                    UpdateCacheExpirationTimeForAllChangeWatchers(DateTimeOffset.UtcNow, operationFailed: true);
                     throw;
                 }
             }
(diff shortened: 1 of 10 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -184,6 +184,26 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                     await RefreshKeyValueCollections(cancellationToken).ConfigureAwait(false);
                     await RefreshKeyValueAdapters(cancellationToken).ConfigureAwait(false);
                 }
+                catch (Exception exception) when (exception is RequestFailedException ||
```
jimmyca15, line 187: I was expecting that we just move the cache expiration update to before we make a server call. Does that not work?

drago-draganov: Network delay shouldn't be an objective for Refresh. It's meant to be precise anyway.

avanigupta: Okay, updating cache expiration time before server call.

> We probably can do it in `finally` instead.

Just realized that if we do it in catch or finally block, we wont be able to control the exact sentinel key for which cache expiration needs to be updated. The new approach solves both issues.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -421,15 +421,14 @@ private async Task RefreshIndividualKeyValues(CancellationToken cancellationToke
...
                 KeyValueIdentifier watchedKeyLabel = new KeyValueIdentifier(watchedKey, watchedLabel);
+                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
 
                 if (_watchedSettings.TryGetValue(watchedKeyLabel, out ConfigurationSetting watchedKv))
                 {
                     KeyValueChange keyValueChange = default;
                     await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, RequestType.Watch, _requestTracingOptions,
                         async () => keyValueChange = await _client.GetKeyValueChange(watchedKv, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
 
-                    changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
```
zhenlan, line 431: Just to make sure I understand, in case of 401, the `CallWithRequestTracing` will throw. That's why this line is skipped?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -421,15 +421,14 @@ private async Task RefreshIndividualKeyValues(CancellationToken cancellationToke
 
                 bool hasChanged = false;
                 KeyValueIdentifier watchedKeyLabel = new KeyValueIdentifier(watchedKey, watchedLabel);
+                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
```
zhenlan, line 424: Just for better readability, can we move this line right after the `if (DateTimeOffset.UtcNow < changeWatcher.CacheExpires)`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -543,6 +539,7 @@ private async Task RefreshKeyValueCollections(CancellationToken cancellationToke
                     });
                 }
 
+                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
```
zhenlan, line 542: Can we move this line right after the `if (DateTimeOffset.UtcNow < changeWatcher.CacheExpires)`, so it's clear the extension is the first thing we do every time.

zhenlan (dismissed):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        /// <summary>
+        /// This method calculates a random exponential backoff which lies between <see cref="MinBackoff"/> and the passed value of maxBackoff.
+        /// </summary>
+        /// <param name="maxBackoff">The maximum backoff time.</param>
+        /// <param name="attempts">The number of attempts made to backoff.</param>
+        /// <returns>The calculated exponential backoff time.</returns>
+        public static TimeSpan CalculateBackoffTime(this TimeSpan maxBackoff, int attempts)
+        {
+            long ticks = MinBackoff.Ticks * new Random().Next(0, (int)Math.Min(Math.Pow(2, attempts - 1), int.MaxValue));
```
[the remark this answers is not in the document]
avanigupta: Done, thanks!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -154,5 +172,37 @@ private void UpdateNextRefreshableSecretFromCache()
                 _nextRefreshTime = null;
             }
         }
+
+        private void UpdateNextRefreshTimeForFailedSecret(string key, CachedKeyVaultSecret cachedSecret)
```
zhenlan, line 176: This is almost duplicated as `SetSecretInCache`. Any chance we can merge the code? For example, `refreshInterval` will be the same as `refreshInterval.CalculateBackoffTime(cachedSecret.RefreshAttempts)` when `RefreshAttempts` is 1.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+    internal static class TimeSpanExtensions
+    {
+        private static readonly TimeSpan MinBackoff = TimeSpan.FromSeconds(1);
```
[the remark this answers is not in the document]
zhenlan: I know they are different conceptually, but they share the same reasoning when we choose the value - that is the minimum refresh interval we want an app ever does. If we do define it separately, can we define this somewhere in a file with all those consts?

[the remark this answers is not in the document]
avanigupta: Sure, we have a lot of consts but they are defined in their own files where they are used. I can move them to their own file of refresh constants.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -425,10 +422,19 @@ private async Task RefreshIndividualKeyValues(CancellationToken cancellationToke
                 if (_watchedSettings.TryGetValue(watchedKeyLabel, out ConfigurationSetting watchedKv))
                 {
                     KeyValueChange keyValueChange = default;
-                    await TracingUtils.CallWithRequestTracing(_requestTracingEnabled, RequestType.Watch, _requestTracingOptions,
-                        async () => keyValueChange = await _client.GetKeyValueChange(watchedKv, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
+                    bool success = false;
+
+                    try
```
zhenlan, line 427: I'm wondering if it makes sense to wrap the whole `if...else..` (line 419- 502) in a big `try...finally`. You will call `UpdateCacheExpirationTime` just once in the end.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -543,20 +556,27 @@ private async Task RefreshKeyValueCollections(CancellationToken cancellationToke
...
-                    currentKeyValues, 
-                    new GetKeyValueChangeCollectionOptions
-                    {
-                        KeyFilter = changeWatcher.Key,
-                        Label = changeWatcher.Label.NormalizeNull(),
-                        RequestTracingEnabled = _requestTracingEnabled,
-                        RequestTracingOptions = _requestTracingOptions
-                    }, 
-                    cancellationToken).ConfigureAwait(false);
+                try
```
zhenlan, line 559: Similarly, can we move the `try` to line 538 to just include everything?

zhenlan:
>                 }

If anything throws in this `if` block, do we still consider the refresh a success? In other words, should we retry upon failures?

---
Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs:584 in e113102. [](commit_id = e11310251b5a7be45c46fa077ec98dc4b7477867, deletion_comment = False)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: I see a bigger value in minimizing the amount of code within the `try` statement. The more statements, the more time it takes to discern where the possible error comes from.

avanigupta:
> > ```
> >             }
> > ```
> 
> If anything throws in this `if` block, do we still consider the refresh a success? In other words, should we retry upon failures?
> 
> Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs:584 in [e113102](https://github.com/Azure/AppConfiguration-DotnetProvider/commit/e11310251b5a7be45c46fa077ec98dc4b7477867). [](commit_id = [e113102](https://github.com/Azure/AppConfiguration-DotnetProvider/commit/e11310251b5a7be45c46fa077ec98dc4b7477867), deletion_comment = False)

We have already updated the cache expiration time before `SetData` is called. The current refresh operation will fail if `SetData` throws, but the next refresh operation will be triggered after the cache expires.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: True, if we handle errors. But the goal here is to extend cache no matter what. The code is cleaner if we use one big try....finally.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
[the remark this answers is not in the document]
avanigupta: On second thoughts, I could move other refresh constants to a separate file, but I think the `MinBackoff` time should still stay in this class. Its not tied to refresh operations in any way. It's solely the property of an extension to the `TimeSpan` struct.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: I did some refactoring to reduce code duplication. However, I don't think we should optimize the calculation of refresh time based on which attempt it is. The code may become unnecessarily complicated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -543,20 +556,27 @@ private async Task RefreshKeyValueCollections(CancellationToken cancellationToke
...
-                changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.CacheExpirationInterval);
+                    success = true;
+                }
+                finally
+                {
+                    UpdateCacheExpirationTime(changeWatcher, success);
+                }
 
-                if (keyValueChanges?.Any() == true)
+                if (keyValueChanges.Any())
```
zhenlan, line 579: > > > ```
> > >             }
> > > ```
> > 
> > 
> > If anything throws in this `if` block, do we still consider the refresh a success? In other words, should we retry upon failures?
> > Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs:584 in [e113102](https://github.com/Azure/AppConfiguration-DotnetProvider/commit/e11310251b5a7be45c46fa077ec98dc4b7477867). [](commit_id = [e113102](https://github.com/Azure/AppConfiguration-DotnetProvider/commit/e11310251b5a7be45c46fa077ec98dc4b7477867), deletion_comment = False)
> 
> We have already updated the cache expiration time before `SetData` is called. The current refresh operation will fail if `SetData` throws, but the next refresh operation will be triggered after the cache expires.

(Codeflow messes up the github comments, so I restart it here)

Yes, we updated the cache expiration time as a successful case, so no quick retry. If an exception thrown from here will cause the app to crash, I guess it's fine. Otherwise, there will be a discrepancy between what's loaded from appconfig and what the app sees. Anyway, it's not a new problem.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Okay, that makes sense. Since we are not handling any exceptions, I like the idea of a single try/finally covering both failure points.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: In this case, we only have one failure point that is covered in the try block. So does it makes sense to include everything?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
zhenlan: No, not meant to depend on the implementation detail of the calculation, but we can have something like 
```cs
refreshSecretAt = DateTimeOffset.UtcNow.Add(
    cachedSecret.RefreshAttempts == 0 ? refreshInterval : refreshInterval.CalculateBackoffTime(cachedSecret.RefreshAttempts)
);
```
This can further simplify your current `GetCacheExpirationTimeForSecret`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -154,5 +167,42 @@ private void UpdateNextRefreshableSecretFromCache()
...
             }
         }
+
+        private DateTimeOffset? GetCacheExpirationTimeForSecret(string key, CachedKeyVaultSecret cachedSecret, bool success)
+        {
+            DateTimeOffset? refreshSecretAt = null;
+
+            if (success)
+            {
+                cachedSecret.RefreshAttempts = 0;
```
zhenlan, line 177: It's not a good idea to make changes inside a `Get` method. Change it outside.

zhenlan: Alternatively, you change the method name to something like `ExtendCacheExpirationTimeForSecret`.

avanigupta: Thanks, updated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Did more refactoring to get closer to the suggested approach.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        public static TimeSpan CalculateBackoffTime(this TimeSpan interval, int attempts)
+        {
+            if (attempts < 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(attempts), attempts, "The number of attempts should not be less than 1.");
+            }
+
+            if (interval < MinBackoff)
+            {
+                throw new ArgumentOutOfRangeException(nameof(interval), interval, $"The interval cannot be less than {nameof(MinBackoff)}.");
```
zhenlan, line 25: The `internal` is the cache expiration time a user sets. Now that you throw if it's smaller than `MinBackOff`. This is exactly what we are already doing in `AzureAppConfigurationRefreshOptions.SetCacheExpiration`. The `MinBackoff` shouldn't be different than `MinimumCacheExpirationInterval`.

zhenlan:
> using System;

Please include the license header.

---
Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs:1 in b531bdb. [](commit_id = b531bdb2dbaf3e5a7b9540f9e5b72ce9c571edd0, deletion_comment = False)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -50,7 +51,23 @@ public async Task<string> GetSecretValue(Uri secretUri, string key, Cancellation
...
+
+                try
+                {
+                    secret = await client.GetSecretAsync(secretName, secretVersion, cancellationToken).ConfigureAwait(false);
+                    success = true;
+                }
+                finally
+                {
+                    // If this is not already a cached secret, we don't have any refresh time to update,
+                    // i.e. there's no way to track next refresh time of new secrets that fail to resolve.
```
zhenlan, line 64: When this actually happens, shouldn't we cache a null secret and extend the expiration anyway (as if it was a successful case)? I know this shouldn't happen, but this makes sure we don't have a code path that doesn't extend the expiration.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -50,7 +51,23 @@ public async Task<string> GetSecretValue(Uri secretUri, string key, Cancellation
...
+                    if (!success && cachedSecret != null)
+                    {
+                        SetSecretInCache(key, secretValue: null, cachedSecret: cachedSecret, success: false);
+                    }
+                }
+
                 secretValue = secret?.Value;
                 SetSecretInCache(key, secretValue);
```
zhenlan, line 72: If you agree that we should extend the expiration in all cases, this whole `else if` block can be simplified to 

```cs
try
{
    KeyVaultSecret secret = await client.GetSecretAsync(secretName, secretVersion, cancellationToken).ConfigureAwait(false);
    secretValue = secret?.Value;
}
finally
{
        SetSecretInCache(key, secretValue, cachedSecret);
}

```

Then in your `SetSecretInCache`, you create a new cached secret only if the passed in `cachedSecret` is null (the success case).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
zhenlan: Now I think about it again - this whole exponential backoff is to prevent stale cache in case a user sets a very long expiration and the last attempt fails. This is not a problem if the user sets a reasonable short expiration time. Therefore, I think our `MinBackoff` should be something higher. It can be our default cache expiration time, i.e., 30s or even higher like 1 min.

avanigupta: If the refresh interval set by users is short (say 5 seconds), are we saying we still won't refresh for another 1 minute?
The idea was to have a backoff time between `[MinimumCacheExpirationInterval, Math.Min(cacheExpirationInterval, DefaultMaxBackoff)]`.

zhenlan: Correct. The underlying SDK retries in a much shorter time automatically already. 

You introduce a second layer of retry to solve the long expiration time issue. In your example, your new backoff retry won't kick in at all. If something keeps failing, we will retry every 5 seconds as the user configured.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: `null` is a valid secret value. If we start caching `null` for failed secrets, we won't be able to track which secret failed to load previously.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
zhenlan: Here is what I think the backoff time should be. Let's say a user sets cacheExpiration

- if (cacheExpiration < minBackoff), we honor cacheExpiration
- if (minBackoff < cacheExpiration < maxBackoff), we pick a value between [minBackoff, cacheExpiration]
- if (cacheExpiration > maxBackoff), we use maxBackoff

avanigupta: Sounds good, let's use 30 seconds min backoff time then. Its consistent with our DefaultCacheExpirationInterval.
Also, one correction to the last point:

- if (cacheExpiration > maxBackoff), we pick a value between [minBackoff, maxBackoff]

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
zhenlan: I don't think we need to differentiate. If the previous attempt failed, we would not try again until the "backoff" time passes again. Isn't that what you want anyway?

avanigupta: That makes sense because as far as the application is concerned, null secret value means that the secret is unusable. So we can say that the secret value can be `null` if:
- The secret was not found in KeyVault.
- The secret was found but its value is `null`.
- The secret failed to load due to an exception.
 
Also, right now if we fail to load a secret from KeyVault due to an exception, we would just never try to load the secret again - until its time to refresh another cached secret, or until RefreshAll is triggered. But if we start caching failed secrets as `null` and tracking their next refresh time, this would no longer be an issue.
Thanks for the great suggestion!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
avanigupta: Updated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -108,30 +125,25 @@ private SecretClient GetSecretClient(Uri secretUri)
...
-
-            if (_keyVaultOptions.SecretRefreshIntervals.TryGetValue(key, out TimeSpan refreshInterval))
+            if (success)
             {
-                refreshSecretAt = DateTimeOffset.UtcNow.Add(refreshInterval);
-            }
-            else if (_keyVaultOptions.DefaultSecretRefreshInterval.HasValue)
-            {
-                refreshSecretAt = DateTimeOffset.UtcNow.Add(_keyVaultOptions.DefaultSecretRefreshInterval.Value);
+                cachedSecret = new CachedKeyVaultSecret(secretValue);
```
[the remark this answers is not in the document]
avanigupta: While its possible to do that, we still need to track the `success` status in `UpdateCacheExpirationTimeForSecret`. So I am still using it in `SetSecretInCache`  for readability.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
avanigupta: Updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RefreshConstants.cs
```diff
@@ -0,0 +1,24 @@
...
+        // Key-values
+        public static readonly TimeSpan DefaultCacheExpirationInterval = TimeSpan.FromSeconds(30);
+        public static readonly TimeSpan MinimumCacheExpirationInterval = TimeSpan.FromMilliseconds(1000);
+
+        // Feature flags
+        public static readonly TimeSpan DefaultFeatureFlagsCacheExpirationInterval = TimeSpan.FromSeconds(30);
+        public static readonly TimeSpan MinimumFeatureFlagsCacheExpirationInterval = TimeSpan.FromMilliseconds(1000);
+
+        // Key Vault secrets
+        public static readonly TimeSpan MinimumSecretRefreshInterval = TimeSpan.FromMilliseconds(1000);
```
avanigupta, line 19: `MinimumSecretRefreshInterval` is not being used yet. I'll create a separate PR for adding this check in KeyVault options. Since its technically a breaking change, we can release it with 5.0.0.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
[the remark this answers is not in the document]
zhenlan: Yeah, I wrote that comment before the other one. What you have now is good.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -50,14 +51,25 @@ public async Task<string> GetSecretValue(Uri secretUri, string key, Cancellation
...
+                    success = true;
+                }
+                finally
+                {
+                    SetSecretInCache(key, cachedSecret, success);
+                }
             }
             else if (_keyVaultOptions.SecretResolver != null)
             {
                 secretValue = await _keyVaultOptions.SecretResolver(secretUri).ConfigureAwait(false);
```
zhenlan, line 70: What if a customer secret resolver throws? I guess we don't to want to call it like a crazy either.

zhenlan:
>                 throw new UnauthorizedAccessException("No key vault credential or secret resolver callback configured, and no matching secret client could be found.");

This exception will be swallowed if called during refresh, right? I guess we still need a way to extend the cache.

---
Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs:76 in a84a990. [](commit_id = a84a99047e60d10f8225d96697d22cbd4331c865, deletion_comment = False)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <param name="attempts">The number of attempts made to backoff.</param>
+        /// <returns>The calculated exponential backoff time, or <paramref name="interval"/> if it is less than <see cref="RefreshConstants.DefaultCacheExpirationInterval"/>.</returns>
+        public static TimeSpan CalculateBackoffTime(this TimeSpan interval, int attempts)
+        {
+            if (attempts < 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(attempts), attempts, "The number of attempts should not be less than 1.");
+            }
+
+            if (interval < RefreshConstants.DefaultCacheExpirationInterval)
```
zhenlan, line 25: >DefaultCacheExpirationInterval [](http://example.com/codeflow?start=44&length=30)

Do you think if it makes sense to define a separate DefaultMinBackoff for 30s?

avanigupta:
> > ```
> >             throw new UnauthorizedAccessException("No key vault credential or secret resolver callback configured, and no matching secret client could be found.");
> > ```
> 
> This exception will be swallowed if called during refresh, right? I guess we still need a way to extend the cache.
> 
> Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs:76 in [a84a990](https://github.com/Azure/AppConfiguration-DotnetProvider/commit/a84a99047e60d10f8225d96697d22cbd4331c865). [](commit_id = [a84a990](https://github.com/Azure/AppConfiguration-DotnetProvider/commit/a84a99047e60d10f8225d96697d22cbd4331c865), deletion_comment = False)

During refresh, this exception can only occur when a sentinel key update triggers RefreshAll, and we received a new KVR that this app has never resolved before. So now we are trying to resolve a KVR that doesn't have any registered SecretClient and there is no SecretResolver callback configured. This KVR can never be resolved no matter how many times we retry. So we never cache null value for this failure case and therefore, this failed KVR cannot affect the next refresh time of other secrets.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
zhenlan: I like the central place for all these consts. Maybe we can have a more generic naming of the class if you can think of one, or we wait till the "fallback" code change to update it to something else :)

avanigupta: Ok, leaving it in RefreshConstants for now. We can update it in future depending on the final implementation of fallback code.

zhenlan:
>                 throw new UnauthorizedAccessException("No key vault credential or secret resolver callback configured, and no matching secret client could be found.");

> During refresh, this exception can only occur when a sentinel key update triggers RefreshAll, and we received a new KVR that this app has never resolved before. So now we are trying to resolve a KVR that doesn't have any registered SecretClient and there is no SecretResolver callback configured. This KVR can never be resolved no matter how many times we retry. So we never cache null value for this failure case and therefore, this failed KVR cannot affect the next refresh time of other secrets.

What if the current KVR is the key being monitored (or the sentinel key)?

---
In reply to: [960161400](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/283#issuecomment-960161400) [](http://example.com/codeflow?ancestors=960161400)

---
Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs:76 in b531bdb. [](commit_id = b531bdb2dbaf3e5a7b9540f9e5b72ce9c571edd0, deletion_comment = False)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <param name="attempts">The number of attempts made to backoff.</param>
+        /// <returns>The calculated exponential backoff time, or <paramref name="interval"/> if it is less than <see cref="RefreshConstants.DefaultMinBackoff"/>.</returns>
+        public static TimeSpan CalculateBackoffTime(this TimeSpan interval, int attempts)
+        {
+            if (attempts < 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(attempts), attempts, "The number of attempts should not be less than 1.");
+            }
+
+            if (interval < RefreshConstants.DefaultMinBackoff)
```
zhenlan, line 25: Small optimization: since the DefaultMinBackoff is the same as DefaultCacheExpirationInterval currently, if we do "<=" here, it probably can save the calculation down below for many people.

zhenlan (approved):
Everything looks great to me. Let's ship it!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/CachedKeyVaultSecret.cs
```diff
@@ -17,10 +17,16 @@ internal class CachedKeyVaultSecret
         /// </summary>
         public DateTimeOffset? RefreshAt { get; set; }
 
-        public CachedKeyVaultSecret(string secretValue, DateTimeOffset? refreshAt)
+        /// <summary>
+        /// The number of times we tried to reload this secret with exponential backoff strategy.
```
jimmyca15, line 21: > with exponential backoff strategy

I don't think attemp tracking needs to be coupled with backoff implementation.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueWatcher.cs
```diff
@@ -33,6 +33,11 @@ internal class KeyValueWatcher
         /// </summary>
         public DateTimeOffset CacheExpires { get; set; }
 
+        /// <summary>
+        /// The number of times we tried to reload this key-value with exponential backoff strategy.
+        /// </summary>
+        public int RefreshAttempts { get; set; } = 0;
```
jimmyca15, line 39: Why default it to 0? Also, we don't do that in CachedKeyVaultSecret

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RefreshConstants.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        // Key-values
+        public static readonly TimeSpan DefaultCacheExpirationInterval = TimeSpan.FromSeconds(30);
+        public static readonly TimeSpan MinimumCacheExpirationInterval = TimeSpan.FromMilliseconds(1000);
+
+        // Feature flags
+        public static readonly TimeSpan DefaultFeatureFlagsCacheExpirationInterval = TimeSpan.FromSeconds(30);
+        public static readonly TimeSpan MinimumFeatureFlagsCacheExpirationInterval = TimeSpan.FromMilliseconds(1000);
+
+        // Key Vault secrets
```
jimmyca15, line 18: TimeSpan.FromMilliseconds(1000) can use timespan.fromseconds(1). Then it's all seconds, no ms in the picture.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueWatcher.cs
avanigupta: We default it to 0 in the CachedKeyVaultSecret constructor.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RefreshConstants.cs
avanigupta: updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/CachedKeyVaultSecret.cs
avanigupta: removed

jimmyca15 (approved):

PR #314, opened 2022-03-23
Update backoff time calculation

avanigupta:
The existing logic for calculating randomized exponential backoff can overflow as there is no upper limit to the number of attempts. This PR:
- handles overflow, 
- removes `Math.Pow` dependency to improve performance, 
- consolidates our backoff logic with the server side logic.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
@@ -10,9 +10,9 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
         private const int MaxAttempts = 63;
 
         /// <summary>
-        /// This method calculates randomized exponential backoff which lies between <paramref name="min"/> and <paramref name="max"/>.
+        /// This method calculates randomized exponential backoff times for operations that occur periodically on a given <paramref name="interval"/>.
         /// </summary>
-        /// <param name="interval">The maximum backoff to be used if <paramref name="interval"/> is less than <paramref name="max"/>.
+        /// <param name="interval">The periodic interval at which the operation occurs. If <paramref name="interval"/> is less than <paramref name="max"/>, <paramref name="interval"/> will be the maximum backoff time..
         /// If <paramref name="interval"/> is less than <paramref name="min"/>, <paramref name="interval"/> is returned.
         /// </param>
         /// <param name="min">The minimum backoff time if <paramref name="interval"/> is greater than <paramref name="min"/>.</param>
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
@@ -7,8 +7,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 {
     internal static class TimeSpanExtensions
     {
+        private const int MaxAttempts = 63;
+
         /// <summary>
-        /// This method calculates a random exponential backoff which lies between <see cref="RefreshConstants.DefaultMinBackoff"/> and <see cref="RefreshConstants.DefaultMaxBackoff"/>.
+        /// This method calculates randomized exponential backoff which lies between <see cref="RefreshConstants.DefaultMinBackoff"/> and <see cref="RefreshConstants.DefaultMaxBackoff"/>.
         /// </summary>
         /// <param name="interval">The maximum backoff to be used if <paramref name="interval"/> is less than <see cref="RefreshConstants.DefaultMaxBackoff"/>.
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -7,8 +7,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 {
     internal static class TimeSpanExtensions
     {
+        private const int MaxAttempts = 63;
```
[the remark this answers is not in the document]
avanigupta: The reason we use 63 max attempts is because 1 << 63 is the first value that overflows. It has nothing to do with the max number of attempts or retries. Attempts can keep growing beyond 63 but `MaxAttempts` is just used to limit the max backoff period for a specific attempt.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -27,10 +29,25 @@ public static TimeSpan CalculateBackoffTime(this TimeSpan interval, int attempts
...
 
-            TimeSpan maxBackoff = TimeSpan.FromTicks(Math.Min(interval.Ticks, RefreshConstants.DefaultMaxBackoff.Ticks));
-            TimeSpan calculatedBackoff = TimeSpan.FromTicks(RefreshConstants.DefaultMinBackoff.Ticks * new Random().Next(1, (int)Math.Min(Math.Pow(2, attempts - 1), int.MaxValue)));
+            TimeSpan min = RefreshConstants.DefaultMinBackoff;
+            TimeSpan max = TimeSpan.FromTicks(Math.Min(interval.Ticks, RefreshConstants.DefaultMaxBackoff.Ticks));
+
+            if (attempts == 1)
+            {
+                return min;
+            }
```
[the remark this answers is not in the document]
avanigupta: Good point.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -7,8 +7,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
...
     internal static class TimeSpanExtensions
     {
+        private const int MaxAttempts = 63;
+
         /// <summary>
-        /// This method calculates a random exponential backoff which lies between <see cref="RefreshConstants.DefaultMinBackoff"/> and <see cref="RefreshConstants.DefaultMaxBackoff"/>.
+        /// This method calculates randomized exponential backoff which lies between <see cref="RefreshConstants.DefaultMinBackoff"/> and <see cref="RefreshConstants.DefaultMaxBackoff"/>.
         /// </summary>
         /// <param name="interval">The maximum backoff to be used if <paramref name="interval"/> is less than <see cref="RefreshConstants.DefaultMaxBackoff"/>.
         /// If <paramref name="interval"/> is less than <see cref="RefreshConstants.DefaultMinBackoff"/>, <paramref name="interval"/> is returned.
```
jimmyca15, line 16: It's a bit odd to me that the self-contained extension has bounds defined by static type with reference to `Refresh`. If you ask me, bounds should be input parameters.

avanigupta: Thats a good point. Right now we use this extension for calculating the backoff time during refresh only. But the idea of a backoff utility is pretty generic and should be reusable in other scenarios too. I can move the min/max to being input parameters.

pratiksanglikar (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -7,30 +7,48 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 {
     internal static class TimeSpanExtensions
     {
+        private const int MaxAttempts = 63;
+
         /// <summary>
-        /// This method calculates a random exponential backoff which lies between <see cref="RefreshConstants.DefaultMinBackoff"/> and <see cref="RefreshConstants.DefaultMaxBackoff"/>.
+        /// This method calculates randomized exponential backoff which lies between <paramref name="min"/> and <paramref name="max"/>.
```
jimmyca15, line 13: the summary isn't correct. It's more like the method helps calculate retry backoff times for operations that occur periodically on a given interval

The interval parameter description looks like it could use some attention as well.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
avanigupta: Updated.

jimmyca15 (approved):

Issue #315, opened 2022-03-29
Dynamic Feature - Microsoft.FeatureManagement v3 support


jimmyca15:
# Microsoft.FeatureManagement 3.0.0 support.

This configuration library provides a mechanism to use feature flags defined in Azure App Configuration with the Microsoft.FeatureManagement library. Version 3.0 of the Microsoft.FeatureManagement library will be introducing a [new schema](https://github.com/microsoft/FeatureManagement-Dotnet/blob/feature/v3/docs/schemas/FeatureManagement.v2.0.0.json) to define feature flags. This new schema will enable [dynamic features](https://github.com/microsoft/FeatureManagement-Dotnet/blob/feature/v3/README.md#dynamic-features) to be [defined](https://github.com/microsoft/FeatureManagement-Dotnet/blob/c816292a8e789be0c8b809c4d172afa05b491f9c/examples/FeatureFlagDemo/appsettings.json#L70).

To enable compatability with the new schema introduced in the Microsoft.FeatureManagement v3 library we should add an opt-in capability into our next 5.x preview package (E.g. 5.1.0-preview). This will enable two scenarios to work with the v3 Microsoft.FeatureManagement library:

* Use dynamic features defined in Azure App Configuration with the Microsoft.FeatureManagement library.
* Define feature flags in appsettings file using the new feature flag schema while simultaneously pulling feature flags from Azure App Configuration.

**Note**: Enabling this opt-in flag will cause feature flags to stop working with version 2.x of the feature management library.

### Opt-in flag

```
public class FeatureFlagOptions
{
    ...

    /// <summary>
    /// Forces feature flags to be written into the configuration system using the feature flag 2.0 schema introduced in
    /// Microsoft.FeatureManagement 3.0.0-preview.
    /// </summary>
    public bool EnablePreviewFeatureFlagSchema { get; set; }

    ...
}
```

### Usage

```
o.UseFeatureFlags(fo =>
{
    fo.EnablePreviewFeatureFlagSchema = true;
}
```

## Guidance

If you would like to try the 3.0.0-preview release of Microsoft.FeatureManagement and get feature flags as well as dynamic features from Azure App Configuration then enable this opt-in flag.

If you would like to continue using the stable 2.x versions of Microsoft.FeatureManagement then this flag can be ignored.

## Preview Functionality

This flag will only be available in preview versions of the 5.x Azure App Configuration .NET configuration provider.

jimmyca15:
[...]

PR #316, opened 2022-04-13
Add an option to support Feature Management V2 schema

avanigupta:
Adding a preview option to support Feature Management V2 schema introduced in Microsoft.FeatureManagement 3.0.0-preview library.
More details in #315


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            var keyValues = new List<KeyValuePair<string, string>>();
+
+            keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeaturesSectionName}:{dynamicFeature.Id}:Assigner", dynamicFeature.ClientAssigner));
+
+            for (int i = 0; i < dynamicFeature.Variants.Count; i++)
+            {
+                FeatureVariant variant = dynamicFeature.Variants[i];
+
+                keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeaturesSectionName}:{dynamicFeature.Id}:Variants:{i}:Name", variant.Name));
```
jimmyca15, line 36: Instead of ":" I'd recommend we use `ConfigurationPath.KeyDelimiter`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
```diff
@@ -0,0 +1,71 @@
...
+                 dynamicFeature = JsonSerializer.Deserialize<DynamicFeature>(setting.Value);
+            }
+            catch (JsonException e)
+            {
+                throw new FormatException(setting.Key, e);
+            }
+
+            var keyValues = new List<KeyValuePair<string, string>>();
+
+            keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeaturesSectionName}:{dynamicFeature.Id}:Assigner", dynamicFeature.ClientAssigner));
```
jimmyca15, line 30: It might help readers to store `$"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeaturesSectionName}"` in a variable and continue referencing it that way.

jimmyca15: Also, it's hard to see everything since the line gets long. It may also help to break it into a few lines.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            for (int i = 0; i < dynamicFeature.Variants.Count; i++)
+            {
+                FeatureVariant variant = dynamicFeature.Variants[i];
+
```
jimmyca15, line 35: Can cache `$"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeaturesSectionName}:{dynamicFeature.Id}:Variants:{i}"` to make the following string concatenations shorter.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
```diff
@@ -0,0 +1,71 @@
...
+                    keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeaturesSectionName}:{dynamicFeature.Id}:Variants:{i}:AssignmentParameters:{kvp.Key}", kvp.Value));
+                }
+            }
+
+            return Task.FromResult<IEnumerable<KeyValuePair<string, string>>>(keyValues);
+        }
+
+        public bool CanProcess(ConfigurationSetting setting)
+        {
+            string contentType = setting?.ContentType?.Split(';')[0].Trim();
```
jimmyca15, line 55: `Split` allocates unnecessarily. If we just need everything before `;`, it's better to use indexof and substring.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagV1KeyValueAdapter.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement
 {
-    internal class FeatureManagementKeyValueAdapter : IKeyValueAdapter
+    internal class FeatureFlagV1KeyValueAdapter : IKeyValueAdapter
```
jimmyca15, line 14: I expected version to be a parameter in the adapter constructor rather than being bundled into the type. Is there a motivation to have separate types?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
```diff
@@ -0,0 +1,71 @@
...
+                keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeaturesSectionName}:{dynamicFeature.Id}:Variants:{i}:ConfigurationReference", variant.ConfigurationReference));
+                
+                if (variant.Default)
+                {
+                    keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeaturesSectionName}:{dynamicFeature.Id}:Variants:{i}:Default", variant.Default.ToString()));
+                }
+
+                foreach (KeyValuePair<string, string> kvp in new JsonFlattener().FlattenJson(variant.AssignmentParameters))
+                {
+                    keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeaturesSectionName}:{dynamicFeature.Id}:Variants:{i}:AssignmentParameters:{kvp.Key}", kvp.Value));
```
jimmyca15, line 46: Seems like `"Variants"`, and the other fields are worth a constant definition,

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeature.cs
```diff
@@ -0,0 +1,24 @@
...
+        [JsonPropertyName("id")]
+        public string Id { get; set; }
+
+        [JsonPropertyName("client_assigner")]
+        public string ClientAssigner { get; set; }
+
+        [JsonPropertyName("variants")]
+        public List<FeatureVariant> Variants { get; set; } = new List<FeatureVariant>();
+
+        // Reserved to allow variants list to be pulled from separate configuration.
```
jimmyca15, line 20: I don't see it being used anywhere, is that right? Reserving it isn't valuable if we don't process it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -28,6 +28,12 @@ public class FeatureFlagOptions
         /// </summary>
         public TimeSpan CacheExpirationInterval { get; set; } = RefreshConstants.DefaultFeatureFlagsCacheExpirationInterval;
 
+        /// <summary>
+        /// Forces feature flags and dynamic features to be written into the configuration system using the  
+        /// feature management V2.0 schema introduced in Microsoft.FeatureManagement 3.0.0-preview library.
+        /// </summary>
+        public bool EnableFeatureManagementV2SchemaPreview { get; set; } = false;
```
jimmyca15, line 35: This API should only be in preview versions of the library. I'm wondering if there is a way to break the build if a stable version is produced with this API exposed. Perhaps a test case could do it?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -28,6 +28,12 @@ public class FeatureFlagOptions
         /// </summary>
         public TimeSpan CacheExpirationInterval { get; set; } = RefreshConstants.DefaultFeatureFlagsCacheExpirationInterval;
 
+        /// <summary>
+        /// Forces feature flags and dynamic features to be written into the configuration system using the  
+        /// feature management V2.0 schema introduced in Microsoft.FeatureManagement 3.0.0-preview library.
+        /// </summary>
```
jimmyca15, line 34: I had a couple of new thoughts on this one.

* The field exists on FeatureFlagOptions, so it's already somewhat expected that the field is related to feature flags + feature management. 
* Specifying V2 in the name *may* get a little confusing since v3.0.0-preview of the feature management is required to consume it.

Do you think with those two things it's fine to just use `bool FeatureFlagOptions.EnablePreviewSchema` ?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
avanigupta: Its making the strings longer and harder to read if we use `ConfigurationPath.KeyDelimiter`. For example:
```
string variantsSectionPrefix = DynamicFeatureSectionPrefix +
                               ConfigurationPath.KeyDelimiter +
                               dynamicFeature.Id +
                               ConfigurationPath.KeyDelimiter + 
                               FeatureManagementConstants.Variants +
                               ConfigurationPath.KeyDelimiter + 
                               i;
```

Vs using the `:` character:

```
string variantsSectionPrefix = $"{DynamicFeatureSectionPrefix}:{dynamicFeature.Id}:{FeatureManagementConstants.Variants}:{i}";

```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagV1KeyValueAdapter.cs
avanigupta: If we use a single adapter, we need to keep track of the schema version within the adapter. This means that the adapter must decide at runtime whether the feature flag will be flattened with V1 schema or V2 schema. 
I imagine it would look something like this:

```
public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, CancellationToken cancellationToken)
{
    if (schema=="v1")
        return ProcessKeyValueInternalV1(....);
    return ProcessKeyValueInternalV2(....);
}
```

This causes an overhead of the `if` check for each feature flag that is loaded. Having two separate  adapters gives us the option to make that decision only once during startup and add only the necessary adapters.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
avanigupta: Do we want to keep symmetry with the option name we'll release with the stable version for opting out of V2 schema? 
Something like `bool FeatureFlagOptions.UseV1Schema` ? If we need symmetry, how about calling this preview option `bool FeatureFlagOptions.UseV2SchemaPreview`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -166,7 +171,35 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
             }
-            
+
+            IsFeatureManagementV2SchemaEnabled = options.EnableFeatureManagementV2SchemaPreview;
+
+            if (IsFeatureManagementV2SchemaEnabled)
+            {
+                if (_adapters.Any(a => a is FeatureFlagV1KeyValueAdapter))
+                {
+                    throw new InvalidOperationException($"Please select either Feature Management V1 schema or V2 schema, not both.");
```
zhenlan, line 181: Can we be more explicit about what's wrong. For example, something like "Opted in to Feature Management V2 schema, but one or more feature flags defined in V1 schema were detected. Use xxx to specify Feature Management schema opt-in preference".

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -25,6 +25,7 @@ internal class RequestTracingConstants
         public const string DevEnvironmentValue = "Dev";
         public const string KeyVaultConfiguredTag = "UsesKeyVault";
         public const string KeyVaultRefreshConfiguredTag = "RefreshesKeyVault";
+        public const string FeatureManagementV2SchemaEnabledTag = "UsesFeatureManagementV2Schema";
```
zhenlan, line 28: For the tracing, I would suggest we have something like "FMSchema=v2preview" instead of a boolean. This works better in case we have more versions. It also avoids changing tracing logic when we change the opt-in/opt-out logic.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
jimmyca15: > Do we want to keep symmetry with the option name we'll release with the stable version for opting out of V2 schema

I don't consider it a requirement. 

If we did though, we'd need to settle on what we want that to be. I don't feel good about the idea of having version as part of a field/method/type name.

So I imagine the opt-out api usage would be something like `featureFlagOptions.Schema = FeatureFlagSchema.V1`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagV1KeyValueAdapter.cs
jimmyca15: > I imagine it would look something like this

My thoughts too. The if check doesn't seem costly, especially compared to what the process methods are doing. Adding new types to a code base does introduce a non-trivial design complexity.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
jimmyca15: Is the concern that `ConfigurationPath.KeyDelimiter` is too big and little but unwieldy? In that case, should be able to do one of the 2.

```
using Delimiter = ConfigurationPath.KeyDelimiter;
```

or 
```
private readonly string Delimiter = ConfigurationPath.KeyDelimiter;
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
zhenlan, line 35: I suggest us not to introduce a new API for opt-in/opt-out.

Similar to the tracing, let people opt-in/out-out through an environment variable. I would like to avoid for people hard-coding choices in code. We all know how hard it is to make people update their code later.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
jimmyca15: I was looking to avoid repeated use of literals.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementConstants.cs
```diff
@@ -5,9 +5,24 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManage
...
-        public const string ContentType = "application/vnd.microsoft.appconfig.ff+json";
-        public const string SectionName = "FeatureManagement";
+        public const string FeatureManagementSectionName = "FeatureManagement";
+        public const string Name = "Name";
+
+        // Feature Flags
         public const string EnabledFor = "EnabledFor";
+        public const string FeatureFlagContentType = "application/vnd.microsoft.appconfig.ff+json";
+        public const string FeatureFlagsSectionName = "FeatureFlags";
+        public const string Parameters = "Parameters";
```
zhenlan, line 17: It's nice you put constants for feature flags and dynamic features in two separated sections. I think it's still valuable to add prefix to all constants. It helps to spot errors in the caller code. For example, `FeatureFlagParameters` and  `DynamicFeatureAssigner`. You already have the prefix for some constants, so it will be more consistent too.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
jimmyca15: The plan was to only have this API available in 5.x preview release. So it will be impossible for users to avoid updating their code if they move to a stable release.

Tracing is a passive component. Opt-in or opt-out has zero effect on the behavior of the application. I definitely see why we went with environment variable there. I don't know if that carries over to this though. I can't imagine it would be good for user to set the environment variable in 5.2-preview and then move to 5.3 stable and be scratching their heads at why it's not working as expected.

zhenlan: I mentioned the tracing for its opt-in/out mechanism, not necessarily for its reasoning.

For previews, people define an environment variable for their apps to opt-in to v2 behavior. For stable release, we will have a new environment variable for people to opt-out v2 (or opt-in to v1) behavior. No code change is required. For people who mess up, we do exceptions with clear guidance on what they need to do.

I understand your desire for a breaking change between preview and stable as a notification. Do you plan to have an API in stable and maintain it forever? I'm more concerned about the stable version.

PS. For testing, I don't mind if we define an interval variable to avoid the environment variable (if that's the problem).

zhenlan: We could have an environment variable `FeatureManagementSchema={v1, v2}` and always honor it. The only change between preview and stable is the default value when the variable is not defined. In the preview, it defaults to v1 and v2 for stable.

zhenlan: In addition, the breaking change only has an impact on people who opt-in v2 during the preview. Why do we want to break them when they move to stable and v2 becomes the default?

jimmyca15: > Do you plan to have an API in stable and maintain it forever? 

Would environment variables that control application behavior have less of a maintenance expectation?

> For people who mess up, we do exceptions with clear guidance on what they need to do.

Wanted to make sure I understood you on this one. Do you mean if they are using a 5.x stable release that doesn't respect the opt-in environment variable that we would throw an exception?

> In addition, the breaking change only has an impact on people who opt-in v2 during the preview. Why do we want to break them when they move to stable and v2 becomes the default

I like that point, the environment variable does save us in that regard.

zhenlan: Your point about the maintenance of the environment variable is valid. The difference is it lives outside of the application. It's easier for us to evolve it in the long run (for example, when we introduce v3 and it's not a breaking change in the future). Once people hard code a version in code, it's part of the build. It is harder to change. It's much easier for people to change the environment.

zhenlan: > Do you mean if they are using a 5.x stable release that doesn't respect the opt-in environment variable that we would throw an exception?

No, we should always honor the user's choice even between the preview and stable. What I meant is that the stable version is used together with v1 feature flags (but not opt-in to v1).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagV1KeyValueAdapter.cs
avanigupta: The `if` check will be done once for each feature flag during startup, multiplied by the number of times we update anything in `IConfiguration` due to a refresh operation. We dont know how many feature flags will be loaded by an application so it seems like a significant optimization if we can avoid the additional check. Thoughts?

jimmyca15: An `if` statement does have non-zero cost, but I can't imagine any observable perf impact by saving on avoiding an `if` statement there.  I wouldn't say that there's much of a reason to avoid it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
avanigupta: Even if we use a shorter name to represent this literal, I think using the `:` character makes the string very easy to read and understand.
For example, which looks more natural to you?
This -
```
            keyValues.Add(
                new KeyValuePair<string, string>(
                    DynamicFeatureSectionPrefix +
                    Delimiter +
                    dynamicFeature.Id +
                    Delimiter +
                    FeatureManagementConstants.DynamicFeatureAssigner,
                    dynamicFeature.ClientAssigner));

```

Or this -
```
            keyValues.Add(
                new KeyValuePair<string, string>(
                    $"{DynamicFeatureSectionPrefix}:{dynamicFeature.Id}:{FeatureManagementConstants.Assigner}",
                    dynamicFeature.ClientAssigner));
```

Are there any side effects of repeatedly using the `:` string literal? If yes, another option is to use `string.join()` which IMO makes the code easier to read, but that has an overhead of creating a string array containing the values to be joined.

jimmyca15: There's no side effects. It's the same reason to avoid typing out the literal value of `DynamicFeatureSectionPrefix` each time.  More of a common practice to avoid repeated use of literals. If the literal needs to change for some reason it can be changed at the source rather than at every occurrence. 

If you see value in using the colon directly and want to keep it that way then you can resolve this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
avanigupta: Not applicable anymore.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -166,7 +174,31 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
+            // Read Feature Management schema version from environment variables
+            string featureManagementSchemaVersion = null;
+
+            try
+            {
+                featureManagementSchemaVersion = Environment.GetEnvironmentVariable(FeatureManagementConstants.FeatureManagementSchemaEnvironmentVariable);
+            }
+            catch (SecurityException) { }
             
+            if (!string.IsNullOrWhiteSpace(featureManagementSchemaVersion))
```
zhenlan, line 187: Should we check whether the value of featureManagementSchemaVersion is valid?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementConstants.cs
```diff
@@ -5,9 +5,28 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManage
 {
     internal class FeatureManagementConstants
     {
+        // Common
+        public const string FeatureManagementSchemaEnvironmentVariable = "AZURE_APP_CONFIGURATION_FEATURE_MANAGEMENT_SCHEMA_VERSION";
+        public const string FeatureManagementSchemaV1 = "1";
+        public const string FeatureManagementSchemaV2 = "2";
```
zhenlan, line 11: We can introduce a `FeatureManagementDefaultSchemaVersion` so we don't have to go back and find out what code we need to change for the default next time.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -114,6 +114,8 @@ private static string CreateCorrelationContextHeader(RequestType requestType, Re
                 correlationContextKeyValues.Add(new KeyValuePair<string, string>(RequestTracingConstants.EnvironmentKey, RequestTracingConstants.DevEnvironmentValue));
             }
 
+            correlationContextKeyValues.Add(new KeyValuePair<string, string>(RequestTracingConstants.FeatureManagementSchemaVersionTag, requestTracingOptions.FeatureManagementSchemaVersion));
```
zhenlan, line 117: Love this. Thanks! :)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagKeyValueAdapter.cs
```diff
@@ -0,0 +1,137 @@
...
+    internal class FeatureFlagKeyValueAdapter : IKeyValueAdapter
+    {
+        private static string _schemaVersion = FeatureManagementConstants.FeatureManagementDefaultSchema;
+
+        public FeatureFlagKeyValueAdapter(string schemaVersion)
+        {
+            if (schemaVersion != FeatureManagementConstants.FeatureManagementSchemaV1
+                && schemaVersion != FeatureManagementConstants.FeatureManagementSchemaV2)
+            {
+                throw new InvalidOperationException(
```
jimmyca15, line 23: As a reader it's unexpected to see InvalidOperationException here instead of ArgumentException. Do we do this because somewhere at a higher layer we handle InvalidOperationException better than ArgumentException? If so, I'd recommend to still throw argumentexception if we throw here, and keep invalidoperationexception as high as possible.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagKeyValueAdapter.cs
```diff
@@ -0,0 +1,137 @@
...
+            try
+            {
+                featureFlag = JsonSerializer.Deserialize<FeatureFlag>(setting.Value);
+            }
+            catch (JsonException e)
+            {
+                throw new FormatException(setting.Key, e);
+            }
+
+            if (cancellationToken.IsCancellationRequested)
```
jimmyca15, line 44: This check doesn't really make sense on synchronous code path.

jimmyca15: I think it was already around in the old adapter, but looks like it would be good to remove. There's no async processing happening in here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagKeyValueAdapter.cs
```diff
@@ -0,0 +1,137 @@
...
+            string featureFlagsSectionPrefix;
+
+            if (_schemaVersion == FeatureManagementConstants.FeatureManagementSchemaV1)
+            {
+                featureFlagsSectionPrefix = FeatureManagementConstants.FeatureManagementSectionName;
+            }
+            else
+            {
+                featureFlagsSectionPrefix = FeatureManagementConstants.FeatureManagementSectionName +
+                                            ConfigurationPath.KeyDelimiter +
```
jimmyca15, line 58: I see that we stuck with ':' everywhere else. Intentional to use `ConfigurationPath.KeyDelimiter` here?

jimmyca15: Also, this can be caculated in constructor and then only have to perform a single concatenation for the lifetime of the adapter.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagKeyValueAdapter.cs
```diff
@@ -0,0 +1,137 @@
...
+                                            FeatureManagementConstants.FeatureFlagSectionName;
+            }
+
+            var keyValues = new List<KeyValuePair<string, string>>();
+
+            if (featureFlag.Enabled)
+            {
+                //if (featureFlag.Conditions?.ClientFilters == null)
```
jimmyca15, line 66: What's the comment for?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -114,6 +114,8 @@ private static string CreateCorrelationContextHeader(RequestType requestType, Re
                 correlationContextKeyValues.Add(new KeyValuePair<string, string>(RequestTracingConstants.EnvironmentKey, RequestTracingConstants.DevEnvironmentValue));
             }
```
jimmyca15, line 116: If they never use feature management, the schema version is null and we shouldn't add this right?

jimmyca15: @zhenlan Do we want this for v1 schema or do we just want to know v2 schema usage?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagKeyValueAdapter.cs
avanigupta: Not sure, I copied the old code. I'll clean it up. Thanks!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
avanigupta: > If they never use feature management, the schema version is null and we shouldn't add this right?

The schema will be "1" by default, not null.

jimmyca15: If they're not using feature flags and we don't use the adapter and we send v1, that may be a bit misleading.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagKeyValueAdapter.cs
avanigupta: That's a good point. We are catching all `InvalidOperationException`s when options are initialized during `Build()` method: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/0ac6b93e7abbb31fbb46090baf244dcc91384447/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs#L55

These exceptions are then ignored if `optional` flag is set to true. So why do we only swallow `InvalidOperationException`s here? Why not all other exceptions thrown during options initialization, like
 `ArgumentException`s? The point of having `optional` set to true was that any errors during AppConfig initialization will be ignored and not crash the app, but here we are ignoring only some exceptions. 

I can see why we might want to throw `ArgumentException` on options initialization - because it helps users quickly fix configuration issues on their end. But by that logic, any exception that occurs during option initialization should be thrown even if `optional` is true. And in some cases, we throw `InvalidOperationException` during options initialization.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
avanigupta: Didn't think of that. I'll fix it, thanks!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagKeyValueAdapter.cs
jimmyca15: So it seems the first thing that we need to decide is what the user experience should be like for specifying an invalid version. Which method is going to result in the exception, and what exception should it be.

I would think that it should be thrown when the user calls `UseFeatureFlags()`.

In that case it comes before `Build()` is invoked, and we don't need to worry about optional at that point. It's the same strategy we use for passing an invalid connection string.

avanigupta: Not sure I follow. Options are initialized during `Build`: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/0ac6b93e7abbb31fbb46090baf244dcc91384447/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs#L33

jimmyca15: Oh got it. Thanks, I missed that. Okay, so then we already know wherever an exception, if any, will be thrown.

I see what you mean now. It's a bit inconsistent right now as far as whether an exception will occur when optional is true. In some cases we let the exception bubble up (`ArgumentException`) and some cases we hide it ([example here](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/0ac6b93e7abbb31fbb46090baf244dcc91384447/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs#L167))

So it sounds like the follow up task is for us to define the behavior we want. Do we want to actually be swallowing argument exceptions for things like an invalid connection string? Or do we want to surface all errors from input parameters like the error I linked above?

My preference would be to fail as early as possible for errors that come from invalid inputs, but that seems to contradict the documentation for the enabled flag.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
zhenlan: I thought we had a "UsesFeatureFlag" tracing, but we didn't. This will serve that purpose then :)

avanigupta: The existing implementation relies on tracing feature flag usage by the ".appconfig.featureflag/" prefix in requests. But this would make it more explicit if we started logging v1 usage too.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagKeyValueAdapter.cs
avanigupta: > My preference would be to fail no matter what for errors that come from invalid inputs, but that seems to contradict the documentation for the enabled flag.

My thoughts too. I think our "optional" behavior comes from the "optional" behavior of FileConfigurationProviders. For configuration files, this flag suppresses the exception when a file is not found. Anything else, like invalid syntax in file or empty file path still throws an exception when optional is true. 
Along similar lines, if optional is true for AppConfig provider, any errors that occur during `Build()` should not be suppressed (these are always due to invalid inputs). However, any errors that occur during `Load()` should be suppressed.

cc @zhenlan

zhenlan: I think 1. we should do the validation in AzureAppConfiguraionOptions where we read the environment variable 2. and fall back to the default if the environment variable is anything we don't understand (no throw. This will be a good place to add a log). If there is a mismatch between the default schema version and the actual schema, we throw whatever errors later. For most users in the future, they won't set this environment variable, it will be the same experience.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -166,7 +172,36 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
             
+            if (!string.IsNullOrWhiteSpace(featureManagementSchemaVersion) 
+                && featureManagementSchemaVersion == FeatureManagementConstants.FeatureManagementSchemaV2)
+            {
+                FeatureManagementSchemaVersion = FeatureManagementConstants.FeatureManagementSchemaV2;
+            }
+            else
+            {
+                FeatureManagementSchemaVersion = FeatureManagementConstants.FeatureManagementDefaultSchema;
+            }
```
zhenlan, line 193: Will it be clearer this way?

```cs
if (featureManagementSchemaVersion == FeatureManagementConstants.FeatureManagementSchemaV1 ||
    featureManagementSchemaVersion == FeatureManagementConstants.FeatureManagementSchemaV2)
{
    FeatureManagementSchemaVersion = featureManagementSchemaVersion;
}
else
{
    FeatureManagementSchemaVersion = FeatureManagementConstants.FeatureManagementDefaultSchema;
}
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagKeyValueAdapter.cs
```diff
@@ -0,0 +1,118 @@
...
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement
+{
+    internal class FeatureFlagKeyValueAdapter : IKeyValueAdapter
+    {
+        private static string _schemaVersion = FeatureManagementConstants.FeatureManagementDefaultSchema;
+        private static string _featureFlagsSectionPrefix;
+
+        public FeatureFlagKeyValueAdapter(string schemaVersion)
+        {
+            _schemaVersion = schemaVersion;
```
jimmyca15, line 21: Public method deserves argument validation.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -166,7 +172,36 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
+            try
+            {
+                featureManagementSchemaVersion = Environment.GetEnvironmentVariable(FeatureManagementConstants.FeatureManagementSchemaEnvironmentVariable);
+            }
+            catch (SecurityException) { }
             
+            if (!string.IsNullOrWhiteSpace(featureManagementSchemaVersion) 
+                && featureManagementSchemaVersion == FeatureManagementConstants.FeatureManagementSchemaV2)
+            {
+                FeatureManagementSchemaVersion = FeatureManagementConstants.FeatureManagementSchemaV2;
```
jimmyca15, line 188: I'd recommend not setting the `FeatureManagementSchemaVersion` member until the adapter is added. The adapter has the if clause to protect against repeated calls. In fact, the environment variable isn't needed unless we are adding the adapter it looks like.

jimmyca15 (approved):

zhenlan (approved):
:shipit:

PR #331, opened 2022-07-11
Add telemetry to indicate the number of endpoints configured.

pratiksanglikar:
This PR adds the telemetry to indicate number of endpoints configured.
The sample log is available [here](https://portal.microsoftgeneva.com/s/8E96A547).

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -716,7 +716,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 IsDevEnvironment = TracingUtils.IsDevEnvironment(),
                 IsKeyVaultConfigured = _options.IsKeyVaultConfigured,
                 IsKeyVaultRefreshConfigured = _options.IsKeyVaultRefreshConfigured,
-                FeatureManagementSchemaVersion = _options.FeatureManagementSchemaVersion
+                FeatureManagementSchemaVersion = _options.FeatureManagementSchemaVersion,
+                ReplicaCount = _options.Endpoints?.Count() ?? 1,
             };
         }
 
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -717,7 +717,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 IsKeyVaultConfigured = _options.IsKeyVaultConfigured,
                 IsKeyVaultRefreshConfigured = _options.IsKeyVaultRefreshConfigured,
                 FeatureManagementSchemaVersion = _options.FeatureManagementSchemaVersion,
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -103,6 +103,7 @@ private static string CreateCorrelationContextHeader(RequestType requestType, Re
             IList<string> correlationContextTags = new List<string>();
             
             correlationContextKeyValues.Add(new KeyValuePair<string, string>(RequestTracingConstants.RequestTypeKey, Enum.GetName(typeof(RequestType), requestType)));
+            correlationContextKeyValues.Add(new KeyValuePair<string, string>(RequestTracingConstants.ReplicaCount, requestTracingOptions.ReplicaCount.ToString()));
```
avanigupta, line 106: How about adding replica count only if the app is using more than 1 endpoint? By default, we know that an app always uses at least 1 endpoint.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -716,7 +716,8 @@ private void SetRequestTracingOptions()
                 IsDevEnvironment = TracingUtils.IsDevEnvironment(),
                 IsKeyVaultConfigured = _options.IsKeyVaultConfigured,
                 IsKeyVaultRefreshConfigured = _options.IsKeyVaultRefreshConfigured,
-                FeatureManagementSchemaVersion = _options.FeatureManagementSchemaVersion
+                FeatureManagementSchemaVersion = _options.FeatureManagementSchemaVersion,
+                ReplicaCount = _options.Endpoints?.Count() ?? 1,
```
avanigupta, line 720: When you say `ReplicaCount`, does that mean number of replicas or number of endpoints? If its number of replicas, should it be number of endpoints -1?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
pratiksanglikar: I think it is a good suggestion and saves us space in the header when only one endpoint is available.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
pratiksanglikar: Good point. In the current implementation, it is number of endpoints. I considered using `EndpointsCount` but didn't go that way because of space constraints. But I like the suggestion of just reducing `1` from the `endpoints` to get replicas.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
pratiksanglikar: Done!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
pratiksanglikar: Done!

avanigupta (approved):

jimmyca15 (approved):

PR #332, opened 2022-07-13
Add Feature Filter Telemetry

avanigupta:
This PR adds telemetry to indicate when applications load feature flags from AppConfig using [built-in feature filters](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/README.md#built-in-feature-filters).

Sample Correlation-Context header values:
```
RequestType=Startup
RequestType=Watch,Filter=PRCNT
RequestType=Watch,Filter=CSTM+PRCNT+TIME
RequestType=Watch,Filter=PRCNT+TRGT
RequestType=Watch,Filter=CSTM+PRCNT+TIME+TRGT,UsesKeyVault
```


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFilterTypes.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement
+{
+    /// <summary>
+    /// Types of built-in feature filters.
+    /// </summary>
+    internal class FeatureFilterTypes
+    {
+        private const string CustomFilter = "Custom";
+        private const string PercentageFilter = "Percentage";
+        private const string TimeWindowFilter = "Time";
```
jimmyca15, line 16: I'd stick with the full name, TimeWindow.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFilterTypes.cs
```diff
@@ -0,0 +1,80 @@
...
+//
+
+using System.Text;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement
+{
+    /// <summary>
+    /// Types of built-in feature filters.
+    /// </summary>
+    internal class FeatureFilterTypes
```
jimmyca15, line 12: Based off the usage, this looks like `FeatureFilterConfiguration` or `FeatureFilterSetup`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFilterTypes.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public void ResetFeatureFilters()
+        {
+            UsesCustomFilter = false;
+            UsesPercentageFilter = false;
+            UsesTimeWindowFilter = false;
+            UsesTargetingFilter = false;
+        }
+
+        public override string ToString()
```
jimmyca15, line 38: The knowledge of this string is specific to the telemetry, I'd like to keep the implementation close by there if possible.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementTelemetryHelper.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement
+{
+    internal class FeatureManagementTelemetryHelper
```
jimmyca15, line 9: How do you feel about moving this in to the adapter to avoid a new type/indirection?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFilterTypes.cs
jimmyca15: Or even FeatureFilterTelemetry

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementTelemetryHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        // Built-in Feature Filter Names
+        private static List<string> PercentageFilterNames = new List<string> { "Percentage", "Microsoft.Percentage", "PercentageFilter", "Microsoft.PercentageFilter" };
+        private static List<string> TimeWindowFilterNames = new List<string> { "TimeWindow", "Microsoft.TimeWindow", "TimeWindowFilter", "Microsoft.TimeWindowFilter" };
+        private static List<string> TargetingFilterNames = new List<string> { "Targeting", "Microsoft.Targeting", "TargetingFilter", "Microsoft.TargetingFilter" };
+
+        public static void UpdateFilterTypesTelemetry(string filterName, FeatureFilterTypes filterTypes)
+        {
+            if (PercentageFilterNames.Any(name => name == filterName))
```
jimmyca15, line 18: we don't need case sensitive match.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFilterTypes.cs
avanigupta: I'd say let's use shorter names for all. How about these - Percent, Time, Target ?

jimmyca15: Why not enum in that case?

jimmyca15: If we want to save characters I feel we need to go all the way.

Thats like this

`RequestType=Watch,Filter=0+1+2+3`

avanigupta: If we use codes, we need to remember what each code number means while looking at logs. Our PowerBI query will also need to convert the codes to names, so anyone modifying the query needs to check the source code to know what each code means. Having somewhat trimmed names provide a balance between saving space and maintaining human readability.

jimmyca15: These names aren't made up, Time could be a valid filter. TimeWindow, is our filter.

jimmyca15: I don't feel great about squeezing characters halfway. If there is a concern of correlation context length, we need to consider real solution which is shortening keys and values to enum where appropriate. If not, I don't see any value to introduce drift from our real names.

avanigupta: How do you feel about using codes like this:
```
private enum FilterType
{
    [Description("CSTM")]
    Custom,
    [Description("PRCNT")]
    Percentage,
    [Description("TIME")]
    TimeWindow,
    [Description("TRGT")]
    Targeting
}
```

So the header looks like this:
`RequestType=Watch,Filter=CSTM+PRCNT+TIME+TRGT`

jimmyca15: I'm fine with that, it's a consistent approach with that method. cc @zhenlan to see if he has any concern about it.

I would stick away from defining an enum and using attribute to get those strings though. Private method where we are generating the correlation context should be fine.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFilterTypes.cs
avanigupta: I combined all telemetry operations in a single class called FeatureFilterTelemetry.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementTelemetryHelper.cs
avanigupta: Moved to FeatureFilterTelemetry.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFilterTelemetry.cs
```diff
@@ -0,0 +1,108 @@
...
+            else if (TargetingFilterNames.Any(name => string.Equals(name, filterName, StringComparison.OrdinalIgnoreCase)))
+            {
+                UsesTargetingFilter = true;
+            }
+            else
+            {
+                UsesCustomFilter = true;
+            }
+        }
+
```
jimmyca15, line 65: Mind putting a comment of an example formatted string?

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFilterTelemetry.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <summary>
+        /// Returns a formatted string containing code names, indicating which feature filters are used by the application.
+        /// </summary>
+        /// <returns>Formatted string like: "CSTM+PRCNT+TIME+TRGT", "PRCNT+TRGT", etc. If no filters are used, empty string will be returned.</returns>
+        public override string ToString()
+        {
+            if (!UsesAnyFeatureFilter())
```
jimmyca15, line 72: Can you add the brackets for the `if`, it's really easy for weird things to happen when updating code that has bracketless ifs.

jimmyca15 (changes requested):

jimmyca15 (approved):

PR #333, opened 2022-07-14
Upgrade ConfigProvider to .NET6

pratiksanglikar:
This PR upgrades the config provider to .NET6, also it adds .NET6 to `TargetFrameworks` so .NET6 applications can use the library.


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
```diff
@@ -17,10 +17,10 @@
     <PackageReference Include="Azure.Data.AppConfiguration" Version="1.2.0" />
     <PackageReference Include="Azure.Messaging.EventGrid" Version="4.7.0" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.0.1" />
-    <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.18" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.18" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.18" />
-    <PackageReference Include="System.Text.Json" Version="4.6.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="6.0.1" />
```
avanigupta, line 20: Since we are upgrading the major version of these packages, should we introduce this PR in the next breaking release (v6.0.0)? @jimmyca15 I cant recall what we decided last time with major version upgrades like this.

pratiksanglikar: I talked with @zhenlan, we want these versions to flow into all of our branches. (Stable, preview as well as next release)

jimmyca15: Major version bump indicates breaking changes. If our dependencies have a breaking change, we need to bump a major version as well. 

So yes, bumping up these major versions of our dependencies would be best done with a major version update ourselves.

jimmyca15: Got it. This project doesn't have any asp.net Core dependencies so it doesn't target .NET Core 3, or .NET 5 and I don't believe we want to target .NET 6 here either.

These projects would need a change to add .NET 6 as a target.
[Function Worker](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/01f2115fcb3f8cb18b81426ffd4766035401ce88/src/Microsoft.Azure.AppConfiguration.Functions.Worker/Microsoft.Azure.AppConfiguration.Functions.Worker.csproj#L5)
[ASP.Net Core integration](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/01f2115fcb3f8cb18b81426ffd4766035401ce88/src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj#L5)

pratiksanglikar: Sound good! Thanks Jimmy, Zhenlan for clarification.
I will close this PR and keep the changes into the branch so that I can just use this branch when we decide to upgrade for the next major version.

I'll open up another branch and another PR for adding `net6.0` to targets.

PR #334, opened 2022-07-14
Add .net6 to targets

pratiksanglikar:
Add `.NET 6` to targets


build/CallPowerShell.cmd
```diff
@@ -0,0 +1 @@
+PowerShell %~dp0..\%*
```
jimmyca15, line 1: I see pack.cmd is removed, doesn't look like build.cmd is removed though. Any reason?

pratiksanglikar: Good catch, it was an oversight on my part.

jimmyca15 (approved):

src/Microsoft.Azure.AppConfiguration.Functions.Worker/Microsoft.Azure.AppConfiguration.Functions.Worker.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\build\NugetProperties.props" />
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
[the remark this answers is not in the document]
pratiksanglikar: Microsoft has officially ended the support for .NET 5 in May 2022, I think we are okay with having just .NET 6.
[.NET Core Support Cycles](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)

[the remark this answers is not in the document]
jimmyca15: Good point, that needs a major version bump if we remove a target.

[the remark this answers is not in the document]
avanigupta: We can keep .net 5 support for now and plan for removing it from all packages in the next major release.

tests/Tests.AzureAppConfiguration.Functions.Worker/Tests.AzureAppConfiguration.Functions.Worker.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
avanigupta, line 4: Can you also add .net5 back here?

pratiksanglikar: Done!

avanigupta (approved):

PR #351, opened 2022-08-30
update keyvault secrets package version to 4.3.0, use TryCreate for validation

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -41,7 +42,7 @@ public AzureKeyVaultKeyValueAdapter(AzureKeyVaultSecretProvider secretProvider)
             }
 
             // Uri validation
-            if (string.IsNullOrEmpty(secretRef.Uri) || !Uri.TryCreate(secretRef.Uri, UriKind.Absolute, out Uri secretUri) || secretUri.Segments.Length < 3)
+            if (string.IsNullOrEmpty(secretRef.Uri) || !Uri.TryCreate(secretRef.Uri, UriKind.Absolute, out Uri secretUri) || !KeyVaultSecretIdentifier.TryCreate(secretUri, out KeyVaultSecretIdentifier identifier))
             {
                 throw CreateKeyVaultReferenceException("Invalid Key vault secret identifier", setting, null, secretRef);
             }
```
avanigupta, line 48: I think we can pass the secret identifier to `_secretProvider.GetSecretValue` method, instead of passing `secretUri`.
The `GetSecretValue` method can be updated to accept `KeyVaultSecretIdentifier` instead of `Uri` as first argument.
That would save us the trouble of extracting secret name and version here: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/28aeecf2d919dd862e39eaa0242e8b199f249af3/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs#L40-L41

avanigupta (approved):

PR #352, opened 2022-09-01
Fix whitespace labelfilter bug

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -129,7 +129,7 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
                 throw new ArgumentNullException(nameof(keyFilter));
             }
 
-            if (labelFilter == null)
+            if (labelFilter == null || string.IsNullOrWhiteSpace(labelFilter))
```
avanigupta, line 132: We dont need to check `labelFilter == null` since `IsNullOrWhiteSpace` will check for null too.

avanigupta (approved):

PR #354, opened 2022-09-08
Add configuration update logging with endpoints

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -278,7 +278,7 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                             {
                                 // Trigger a single load-all operation if a change was detected in one or more key-values with refreshAll: true
                                 applicationSettings = await LoadAll(client, cancellationToken).ConfigureAwait(false);
-
+                                _logger?.LogInformation(LoggingConstants.SuccessfulConfigurationUpdated + " on endpoint: " + AppConfigurationEndpoint); // what endpoint is client coming from? for georeplication
```
avanigupta, line 281: >AppConfigurationEndpoint [](http://example.com/codeflow?start=125&length=24)

This is not the current client's endpoint, its just he first endpoint in the list. We can add a new method in `IConfigurationClientManager` to return the endpoint of the current client.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -13,5 +13,13 @@ internal class LoggingConstants
         public const string RefreshFailedDueToKeyVaultError = "A refresh operation failed while resolving a Key Vault reference.";
         public const string RefreshFailedError = "A refresh operation failed.";
         public const string RefreshCanceledError = "A refresh operation was canceled.";
+
+        // Successful update messages
+        public const string SuccessfulConfigurationUpdated = "IConfiguration updated.";
```
avanigupta, line 18: >IConfiguration updated. [](http://example.com/codeflow?start=62&length=23)

To make it more explicit that the entire configuration changed, we can say something like:
"IConfiguration updated from Azure App Configuration endpoint: "

We can follow the same pattern for all messages.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -61,6 +62,7 @@ public async Task<string> GetSecretValue(Uri secretUri, string key, Cancellation
                 if (client != null)
                 {
                     KeyVaultSecret secret = await client.GetSecretAsync(secretName, secretVersion, cancellationToken).ConfigureAwait(false);
+                    logger?.LogInformation(LoggingConstants.SuccessfulKeyVaultSecretUpdated + secret.Name);
```
avanigupta, line 65: > secret.Name [](http://example.com/codeflow?start=93&length=12)

We can add the KeyVault endpoint here

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -291,15 +291,37 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                                adapter.InvalidateCache(change.Current);
+                            }
+                        }
+
+                        foreach (KeyValueChange change in changedKeyValuesCollection)
                         {
                             if (change.ChangeType == KeyValueChangeType.Deleted)
                             {
                                 applicationSettings.Remove(change.Key);
+                                _logger?.LogInformation(LoggingConstants.SuccessfulFeatureFlagDeleted + change.Key + " on endpoint: " + AppConfigurationEndpoint);
```
avanigupta, line 319: > change.Key  [](http://example.com/codeflow?start=103&length=12)

Key names are not needed for feature flags. We can combine both cases and just say that feature flags were updated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -291,15 +291,37 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                            {
+                                applicationSettings[change.Key] = change.Current;
+                                _logger?.LogInformation(LoggingConstants.SuccessfulKeyValueModified + change.Current + " on endpoint: " + AppConfigurationEndpoint);
+                            }
+
+                            // Invalidate the cached Key Vault secret (if any) for this ConfigurationSetting
+                            foreach (IKeyValueAdapter adapter in _options.Adapters)
+                            {
+                                adapter.InvalidateCache(change.Current);
+                            }
```
avanigupta, line 311: We dont need to do it twice

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -103,5 +104,17 @@ public bool UpdateSyncToken(Uri endpoint, string syncToken)
...
         }
+
+        public Uri GetEndpointForClient(ConfigurationClient client)
+        {
+            ConfigurationClientWrapper currentClient = _clients.FirstOrDefault(c => c.Client == client);
+            
+            if (currentClient == null)
+            {
+                throw new ArgumentNullException(nameof(client));
+            }
```
avanigupta, line 115: ArgumentNullException should be thrown if the passed argument is null. In this case, we should check if `client` is null and throw ArgumentNullException. This argument check should be done _before_ we use `client` to find current client.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -13,5 +13,13 @@ internal class LoggingConstants
         public const string RefreshFailedDueToKeyVaultError = "A refresh operation failed while resolving a Key Vault reference.";
         public const string RefreshFailedError = "A refresh operation failed.";
         public const string RefreshCanceledError = "A refresh operation was canceled.";
+
+        // Successful update messages
+        public const string RefreshConfigurationUpdatedSuccess = "All configuration updated from Azure App Configuration endpoint: ";
+        public const string RefreshKeyValueUpdatedSuccess = "Key-Values updated from Azure App Configuration endpoint: ";
```
avanigupta, line 19: Nit: Key-value(s)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -13,5 +13,13 @@ internal class LoggingConstants
...
+        public const string RefreshKeyVaultSecretUpdatedSuccess = "KeyVault updated from KeyVault Uri: ";
```
avanigupta, line 20: How about "Key Vault secret updated from vault: "?

tests/Tests.AzureAppConfiguration/LoggingTests.cs
```diff
@@ -34,13 +37,13 @@ public class LoggingTests
                 label: "label",
                 value: "TestValue2",
                 eTag: new ETag("31c38369-831f-4bf1-b9ad-79db56c8b989"),
-                contentType: "text")
-        };
+                contentType: "text"),
```
avanigupta, line 40: Trailing comma not needed

tests/Tests.AzureAppConfiguration/MockedConfigurationClientManager.cs
```diff
@@ -40,5 +40,17 @@ public bool UpdateSyncToken(Uri endpoint, string syncToken)
...
             return true;
         }
+
+        public Uri GetEndpointForClient(ConfigurationClient client)
+        {
+            ConfigurationClientWrapper currentClient = _clients.FirstOrDefault(c => c.Client == client);
+
+            if (currentClient == null)
+            {
+                throw new ArgumentNullException(nameof(client));
```
avanigupta, line 50: Same applies here - null check passed argument before using it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -103,5 +104,17 @@ public bool UpdateSyncToken(Uri endpoint, string syncToken)
...
+        public Uri GetEndpointForClient(ConfigurationClient client)
+        {
+            if (client == null)
+            {
+                throw new ArgumentNullException(nameof(client));
+            }
+
+            ConfigurationClientWrapper currentClient = _clients.FirstOrDefault(c => c.Client == client);
+            
+            return currentClient.Endpoint;
```
avanigupta, line 117: You can return `currentClient?.Endpoint`. This makes sure that if no matching client is found, this method would return null Uri instead of throwing null reference exception.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -273,28 +277,33 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
 
                                         break;
                                     }
+                                } else
+                                {
+                                    logDebugBuilder.AppendLine(LoggingConstants.RefreshKeyValueUnchanged + watchedKey);
```
avanigupta, line 282: We dont need to log unchanged key-values. Logging what changed should be enough.

zhenlan: We often hear from customers complaining the config change is not picked up, but the reality is what they are monitoring is different from what they changed. So from debugging perspective, this could be useful.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -271,18 +277,34 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
-
+                                logInfoBuilder.AppendLine(LoggingConstants.RefreshConfigurationUpdatedSuccess + endpoint);
+                                logDebugBuilder.AppendLine(LoggingConstants.RefreshConfigurationUpdatedSuccess + endpoint);
                                 return;
                             }
 
+                            if (keyValueChanges.Any())
+                            {
+                                logInfoBuilder.AppendLine(LoggingConstants.RefreshKeyValueUpdatedSuccess + endpoint);
+                                logDebugBuilder.AppendLine(LoggingConstants.RefreshKeyValueUpdatedSuccess + endpoint);
```
avanigupta, line 298: We dont need to log success message again in the debug logger. We can just add the key name in debug logs, info logs will always contain the generic success message anyway.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -210,13 +211,16 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                    StringBuilder logInfoBuilder = new StringBuilder();
+                    StringBuilder logDebugBuilder = new StringBuilder();
 
                     await ExecuteWithFailOverPolicyAsync(availableClients, async (client) =>
                         {
                             applicationSettings = null;
                             keyValueChanges = new Dictionary<KeyValueWatcher, KeyValueChange>();
                             changedKeyValuesCollection = null;
                             refreshAll = false;
+                            Uri endpoint = _configClientManager.GetEndpointForClient(client);
```
avanigupta, line 223: Can you clear both string builders here? If failover occurred, we dont want log statements from failed endpoints to be present in log builders.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -13,5 +13,15 @@ internal class LoggingConstants
         public const string RefreshFailedDueToKeyVaultError = "A refresh operation failed while resolving a Key Vault reference.";
         public const string RefreshFailedError = "A refresh operation failed.";
         public const string RefreshCanceledError = "A refresh operation was canceled.";
+
+        // Successful update messages
+        public const string RefreshConfigurationUpdatedSuccess = "All configuration updated from Azure App Configuration endpoint: ";
+        public const string RefreshKeyValueUpdatedSuccess = "Key-values updated from Azure App Configuration endpoint: ";
```
avanigupta, line 19: Nit: Some key-value(s) updated from Azure App Configuration endpoint:

tests/Tests.AzureAppConfiguration/LoggingTests.cs
```diff
@@ -308,6 +567,21 @@ private bool ValidateLoggedError(Mock<ILogger> logger, string expectedMessage)
             return true;
         }
 
+        private bool ValidateLoggedSuccess(Mock<ILogger> logger, string expectedMessage)
```
avanigupta, line 570: Can you also add tests to check if the key names are properly logged in debug logger?

avanigupta:
>                     }

I think this is something we can add to debug logs - "Skipping refresh operation because all AppConfig endpoints are backed off due to previous failures."

cc @zhenlan 

---
Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs:194 in 17626c6. [](commit_id = 17626c6c527dd9a99bc76b2ae0bce76cdb172b68, deletion_comment = False)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -16,12 +16,14 @@ internal class LoggingConstants
...
+        public const string RefreshKeyValueUpdatedSuccess = "Some key-value(s) updated from Azure App Configuration endpoint: ";
+        public const string RefreshKeyValueUnchanged = "No new changes for key-value: ";
         public const string RefreshKeyValueChanged = "Key-value changed: ";
         public const string RefreshKeyVaultSecretUpdatedSuccess = "Key Vault secret updated from vault: ";
+        public const string RefreshKeyVaultSecretChanged = "Key Vault secret changed: ";
         public const string RefreshFeatureFlagUpdatedSuccess = "Feature flags updated from Azure App Configuration endpoint: ";
 
-        // add endpoints to all constants, just success for now
+        // Other
+        public const string RefreshCanceledDueToNoAvailableEndpoints = "Skipping refresh operation because all AppConfig endpoints are backed off due to previous failures.";
```
avanigupta, line 27: Nit: RefreshSkippedNoClientAvailable

tests/Tests.AzureAppConfiguration/TestHelper.cs
```diff
@@ -126,6 +128,36 @@ public static List<ConfigurationSetting> LoadJsonSettingsFromFile(string path)
...
+            return true;
+        }
+
+        public static bool ValidateLoggedSuccess(Mock<ILogger> logger, string expectedMessage)
+        {
+            Func<object, Type, bool> state = (v, t) => v.ToString().Contains(expectedMessage);
+
+            logger.Verify(
+                x => x.Log(
+                    It.Is<LogLevel>(l => l == LogLevel.Information || l == LogLevel.Debug),
```
avanigupta, line 153: Can we split the debug and information checks? How about creating 3 helpers instead of 2
 - ValidateLoggedWarning
 - ValidateLoggedInformation 
 - ValidateLoggedDebug

avanigupta (approved):

avanigupta (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -61,6 +62,8 @@ public async Task<string> GetSecretValue(Uri secretUri, string key, Cancellation
                 if (client != null)
                 {
                     KeyVaultSecret secret = await client.GetSecretAsync(secretName, secretVersion, cancellationToken).ConfigureAwait(false);
+                    logger?.LogDebug(String.Format("{0}\"{1}\"", LoggingConstants.RefreshKeyVaultSecretChanged, secret.Name));
```
avanigupta, line 65: Nit: does this need string.format? Can we log using concatenation just like the info log below?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -263,6 +270,8 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                                 // Check if a change has been detected in the key-value registered for refresh
                                 if (change.ChangeType != KeyValueChangeType.None)
                                 {
+                                    logDebugBuilder.AppendLine(String.Format("{0}(key: \"{1}\", label: \"{2}\")", LoggingConstants.RefreshKeyValueChanged, change.Key, change.Label));
```
avanigupta, line 273: Nit: we can use `'KeyName'` instead of `"KeyName"` to avoid escape characters.
Also, we can use [string interpolation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) instead of string.format everywhere.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -354,6 +371,14 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                         // SetData because repeating appconfig calls (by not updating watchers) won't help anything for keyvault calls.
                         // As long as adapter.NeedsRefresh is true, we will attempt to update keyvault again the next time RefreshAsync is called.
                         SetData(await PrepareData(applicationSettings, cancellationToken).ConfigureAwait(false));
+                        if (logDebugBuilder.Length > 0)
```
avanigupta, line 374: `SetData` may throw exceptions unrelated to the key-values and feature flags being watched. So we should log whatever we have so far before calling SetData.

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -866,5 +876,119 @@ public void SelectAndRefreshSingleFeatureFlag()
...
+                    ",
+            label: "label2",
+            contentType: FeatureManagementConstants.FeatureFlagContentType + ";charset=utf-8",
+            eTag: new ETag("c3c231fd-39a0-4cb6-3237-4614474b92c1" + "f"));
+
+            Thread.Sleep(CacheExpirationTime);
+            refresher.LoggerFactory = mockLoggerFactory.Object;
+            refresher.TryRefreshAsync().Wait();
+            Assert.Equal("AllUsers", config["FeatureManagement:MyFeature:EnabledFor:0:Name"]);
+            Assert.True(TestHelpers.ValidateLog(mockLogger, LoggingConstants.RefreshFeatureFlagChanged + "(key: '" + FeatureManagementConstants.FeatureFlagMarker + "myFeature2', label: '')", LogLevel.Debug));
```
avanigupta, line 933: This test should be failing, I'm looking into why the build pipeline didn't report the failed test as pipeline failure. Until then, make sure to run tests locally.

Also, this test should've caught the bug that we are not logging on feature flag updates (only deletes are logged). I'll let you figure out why :)

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -866,5 +876,119 @@ public void SelectAndRefreshSingleFeatureFlag()
...
+
+        Response<ConfigurationSetting> GetIfChanged(ConfigurationSetting setting, bool onlyIfChanged, CancellationToken cancellationToken)
+        {
+            return Response.FromValue(FirstKeyValue, new MockResponse(200));
+        }
+
+        Response<ConfigurationSetting> GetTestKey(string key, string label, CancellationToken cancellationToken)
+        {
+            return Response.FromValue(TestHelpers.CloneSetting(FirstKeyValue), new Mock<Response>().Object);
+        }
```
avanigupta, line 945: Helpers can be moved to the end of the file, after all tests.

tests/Tests.AzureAppConfiguration/LoggingTests.cs
```diff
@@ -4,16 +4,20 @@
...
 using Microsoft.Extensions.Configuration.AzureAppConfiguration;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Moq;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection.Emit;
```
avanigupta, line 18: Can be removed

tests/Tests.AzureAppConfiguration/TestHelper.cs
```diff
@@ -3,8 +3,10 @@
 //
 using Azure;
 using Azure.Core;
+using Azure.Core.Testing;
```
avanigupta, line 6: We dont need this reference.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
```diff
@@ -166,6 +169,9 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
                         Label = options.Label.NormalizeNull(),
                         Current = null
                     });
+                    string key = kvp.Key.Replace(FeatureManagementConstants.FeatureFlagMarker, "");
```
avanigupta, line 172: We should use [Substring](https://learn.microsoft.com/en-us/dotnet/api/system.string.substring?view=netstandard-2.0#system-string-substring(system-int32)) instead. 
Substring will be faster because we are not searching for anything in the string. Moreover, `Replace` will replace all occurrences of FeatureFlagMarker from the key.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -350,6 +367,14 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                             }
                         }
 
+                        if (logDebugBuilder.Length > 0)
+                        {
+                            _logger?.LogDebug(logDebugBuilder.ToString().Trim('\r', '\n'));
```
avanigupta, line 372: We can simply call `Trim()` and it will take care of all whitespaces

avanigupta (approved):

PR #355, opened 2022-09-22
Add Map API for App Configuration Settings

amerjusupovic:


tests/Tests.AzureAppConfiguration/MapTests.cs
```diff
@@ -0,0 +1,332 @@
...
+            mockClient.Setup(c => c.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()))
+                .Returns(new MockAsyncPageable(new List<ConfigurationSetting> { _kvr }));
+
+            var mockClientManager = TestHelpers.CreateMockedConfigurationClientManager(mockClient.Object);
+
+            var mockSecretClient = new Mock<SecretClient>(MockBehavior.Strict);
+            mockSecretClient.SetupGet(client => client.VaultUri).Returns(vaultUri);
+            mockSecretClient.Setup(client => client.GetSecretAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()))
+                .Returns((string name, string version, CancellationToken cancellationToken) =>
+                    Task.FromResult((Response<KeyVaultSecret>)new MockResponse<KeyVaultSecret>(new KeyVaultSecret(name, _certValue))));
```
avanigupta, line 105: Would this setup always return `_certValue` for any secret name? If that's the case, this test will pass even if you dont have the mapper changing secret name.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -296,9 +305,16 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
                             {
                                 applicationSettings.Remove(change.Key);
+                                mappedData.Remove(change.Key);
                             }
                             else if (change.ChangeType == KeyValueChangeType.Modified)
                             {
+                                ConfigurationSetting setting = change.Current;
+                                foreach (Func<ConfigurationSetting, ValueTask<ConfigurationSetting>> func in _options.UserDefinedMappers)
+                                {
+                                    setting = func(setting).Result;
```
avanigupta, line 315: The mapper will change the original ConfigurationSetting object, so mappedData and applicationSettings will both point to the modified object. This would be an issue when we want to inspect the value of the original  ConfigurationSetting object during refresh.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -496,10 +513,12 @@ public void ProcessPushNotification(PushNotification pushNotification, TimeSpan?
         private async Task InitializeAsync(bool ignoreFailures, IEnumerable<ConfigurationClient> availableClients, CancellationToken cancellationToken = default)
         {
             Dictionary<string, ConfigurationSetting> data = null;
+            Dictionary<string, ConfigurationSetting> dataCopy = null;
 
             try
             {
                 data = await ExecuteWithFailOverPolicyAsync(availableClients, (client) => LoadAll(client, cancellationToken), cancellationToken).ConfigureAwait(false);
+                dataCopy = data.ToDictionary(kvp => kvp.Key, kvp => new ConfigurationSetting(kvp.Value.Key, kvp.Value.Value, kvp.Value.Label, kvp.Value.ETag));
```
avanigupta, line 521: `dataCopy` is only used to initialize `_applicationSettings`.  Can we do it without creating this new dictionary?

tests/Tests.AzureAppConfiguration/MapTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        ConfigurationSetting sentinelKv = new ConfigurationSetting("SentinelKey", "SentinelValue");
+
+        TimeSpan CacheExpirationTime = TimeSpan.FromSeconds(1);
+
+        string _certValue = "Certificate Value from KeyVault";
+        Uri vaultUri = new Uri("https://keyvault-theclassics.vault.azure.net");
+
+        [Fact]
+        public void MapTransformKeyValue()
+        {
```
avanigupta, line 49: Can you add tests where Map is changing the key/label? What is the expected behavior in both cases? What is the expected behavior if key modifications lead to duplicate keys in IConfiguration?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -404,7 +404,10 @@ public AzureAppConfigurationOptions ConfigureKeyVault(Action<AzureAppConfigurati
         /// <param name="mapper">A callback registered by the user to transform each configuration setting.</param>
         public AzureAppConfigurationOptions Map(Func<ConfigurationSetting, ValueTask<ConfigurationSetting>> mapper)
         {
-            _mappers.Add(mapper);
+            if (mapper != null)
```
avanigupta, line 407: We can throw ArgumentNullException if mapper is null

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -313,7 +304,7 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                                 ConfigurationSetting setting = change.Current;
                                 foreach (Func<ConfigurationSetting, ValueTask<ConfigurationSetting>> func in _options.UserDefinedMappers)
                                 {
-                                    setting = func(setting).Result;
+                                    setting = await func(setting);
```
avanigupta, line 307: `await` should be accompanied by `.ConfigureAwait(false)`.

`setting = await func(setting).ConfigureAwait(false);`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -540,22 +529,12 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
...
-                    }
-                    mappedData[kvp.Key] = setting;
-                }
+                Dictionary<string, ConfigurationSetting> mappedData = await MapConfigurationData(data);
 
                 try
                 {
                     SetData(await PrepareData(mappedData, cancellationToken).ConfigureAwait(false));
-                    _applicationSettings = dataCopy;
+                    _applicationSettings = data.ToDictionary(kvp => kvp.Key, kvp => new ConfigurationSetting(kvp.Value.Key, kvp.Value.Value, kvp.Value.Label, kvp.Value.ETag));
```
avanigupta, line 537: If we call `MapConfigurationData` on `data` before deep copying it, `data` has already been modified by user defined mappers and we lose the original ConfigurationSetting returned by server.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -296,9 +297,16 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
                                 applicationSettings.Remove(change.Key);
+                                mappedData.Remove(change.Key);
                             }
                             else if (change.ChangeType == KeyValueChangeType.Modified)
                             {
+                                ConfigurationSetting setting = change.Current;
+                                foreach (Func<ConfigurationSetting, ValueTask<ConfigurationSetting>> func in _options.UserDefinedMappers)
+                                {
+                                    setting = await func(setting).ConfigureAwait(false);
+                                }
```
avanigupta, line 308: Can this use the common MapConfigurationData method too?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -278,7 +279,7 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                             {
                                 // Trigger a single load-all operation if a change was detected in one or more key-values with refreshAll: true
                                 applicationSettings = await LoadAll(client, cancellationToken).ConfigureAwait(false);
-
+                                mappedData = await MapConfigurationData(applicationSettings);
```
avanigupta, line 282:
```suggestion
                                mappedData = await MapConfigurationData(applicationSettings).ConfigureAwait(false);

```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -29,6 +29,7 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
         private readonly IConfigurationClientManager _configClientManager;
         private AzureAppConfigurationOptions _options;
         private Dictionary<string, ConfigurationSetting> _applicationSettings;
+        private Dictionary<string, ConfigurationSetting> _mappedData;
```
avanigupta, line 32: Can you help rename the existing dictionary to clarify that it only holds the original data from server?
`_applicationSettings` -> `_serverData`

And any local variables should be renamed accordingly.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -523,8 +531,10 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
 
                 try
                 {
-                    SetData(await PrepareData(data, cancellationToken).ConfigureAwait(false));
-                    _applicationSettings = data;
+                    _applicationSettings = data.ToDictionary(kvp => kvp.Key, kvp => new ConfigurationSetting(kvp.Value.Key, kvp.Value.Value, kvp.Value.Label, kvp.Value.ETag));
+                    Dictionary<string, ConfigurationSetting> mappedData = await MapConfigurationData(data);
+                    _mappedData = mappedData;
```
avanigupta, line 536:
```suggestion
                    _mappedData =  await MapConfigurationData(data).ConfiureAwait(false);
                     SetData(await PrepareData(_mappedData, cancellationToken).ConfigureAwait(false));

```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -850,5 +860,22 @@ private bool IsFailOverable(RequestFailedException rfe)
                    rfe.Status == (int)HttpStatusCode.RequestTimeout ||
                    rfe.Status >= (int)HttpStatusCode.InternalServerError;
         }
+
+        private async Task<Dictionary<string, ConfigurationSetting>> MapConfigurationData(Dictionary<string, ConfigurationSetting> data)
+        {
+            Dictionary<string, ConfigurationSetting> mappedData = new Dictionary<string, ConfigurationSetting>();
```
avanigupta, line 866: We should create a dictionary with case-insensitive keys.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta, line 282: Anytime you call MapConfigurationData, the input dictionary will lose its original ConfigurationSettings because we are not making deep copies before mapping. I think the deep copy logic should move inside MapConfigurationData to avoid any errors.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -521,10 +532,20 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
                     adapter.InvalidateCache();
                 }
 
+<<<<<<< HEAD
```
avanigupta, line 535: Looks like there were merge conflicts.

amerjusupovic: Sorry not done yet, needed to push to remote real quick

amerjusupovic: Sorry about that, fixed now, tests are still in progress

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -883,5 +897,42 @@ private bool IsFailOverable(RequestFailedException rfe)
...
+        }
+
+        private void UpdateWatchedSettings()
+        {
+            _watchedSettings.Clear();
+
+            foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
+            {
+                string watchedKey = changeWatcher.Key;
+                string watchedLabel = changeWatcher.Label;
```
avanigupta, line 925: >changeWatcher.Label [](http://example.com/codeflow?start=38&length=19)

We need to call `NormalizeNull()` on the label so that it can be compared properly.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -883,5 +897,42 @@ private bool IsFailOverable(RequestFailedException rfe)
...
+                KeyValueIdentifier watchedKeyLabel = new KeyValueIdentifier(watchedKey, watchedLabel);
+
+                ConfigurationSetting watchedKvCopy = _serverData.Values.FirstOrDefault((kv) => kv.Key.Equals(watchedKey) && kv.Label == watchedLabel);
```
avanigupta, line 928: >watchedKvCopy [](http://example.com/codeflow?start=37&length=13)

Nit: watchedKv 

You can add a comment that the watchedKv will be a deep copy of the original setting from server because we are extracting from _serverData dictionary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -343,42 +355,45 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                                if (keyValueChange.ChangeType == KeyValueChangeType.Modified)
+                                {
+                                    _watchedSettings[kvIdentifier] = keyValueChange.Current;
+                                }
+                                else if (keyValueChange.ChangeType == KeyValueChangeType.Deleted)
+                                {
+                                    _watchedSettings.Remove(kvIdentifier);
+                                }
 
-                            // Already updated cache expiration time if refreshAll is true.
```
avanigupta, line 363: >// Already updated cache expiration time if refreshAll is true. [](http://example.com/codeflow?start=28&length=63)

We can keep this comment for future reference

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,16 +209,17 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
 
                     //
                     // Avoid instance state modification
-                    Dictionary<string, ConfigurationSetting> applicationSettings = null;
+                    Dictionary<string, ConfigurationSetting> serverData = null;
                     Dictionary<KeyValueWatcher, KeyValueChange> keyValueChanges = null;
                     List<KeyValueChange> changedKeyValuesCollection = null;
+                    Dictionary<string, ConfigurationSetting> mappedData = new Dictionary<string, ConfigurationSetting>(_mappedData);
```
avanigupta, line 215: > Dictionary<string, ConfigurationSetting> mappedData = new Dictionary<string, ConfigurationSetting>(_mappedData); [](http://example.com/codeflow?start=19&length=113)

It would be better to keep the same initialization pattern here as serverData. Since both dictionaries need to be updated at the same time, keeping their code style in sync would help with readability.
We can create a new dictionary for mappedData when its needed (we do the same for serverData later).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,16 +209,17 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
                     List<KeyValueChange> changedKeyValuesCollection = null;
+                    Dictionary<string, ConfigurationSetting> mappedData = new Dictionary<string, ConfigurationSetting>(_mappedData);
                     bool refreshAll = false;
                     StringBuilder logInfoBuilder = new StringBuilder();
                     StringBuilder logDebugBuilder = new StringBuilder();
 
                     await ExecuteWithFailOverPolicyAsync(availableClients, async (client) =>
                         {
-                            applicationSettings = null;
+                            serverData = null;
```
avanigupta, line 222: >serverData = null; [](http://example.com/codeflow?start=28&length=18)

Also set mappedData to null here.

avanigupta:
>                                 string watchedLabel = changeWatcher.Label;

This isn't part of your PR, but can you call `NormalizeNull()` on this label too? We need to do this everytime we compare KeyValueWatcher.Label with ConfigurationSetting.Label

---
Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs:233 in 7beacca. [](commit_id = 7beacca71d1911c5771a032c2f7fe302f2c7e887, deletion_comment = False)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -883,5 +897,42 @@ private bool IsFailOverable(RequestFailedException rfe)
                    rfe.Status == (int)HttpStatusCode.RequestTimeout ||
                    rfe.Status >= (int)HttpStatusCode.InternalServerError;
         }
+
+        private async Task<Dictionary<string, ConfigurationSetting>> MapConfigurationData(Dictionary<string, ConfigurationSetting> data)
```
avanigupta, line 901: >MapConfigurationData [](http://example.com/codeflow?start=69&length=20)

Can we call it `MapConfigurationSettings`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -883,5 +897,42 @@ private bool IsFailOverable(RequestFailedException rfe)
...
+            foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers)
+            {
+                string watchedKey = changeWatcher.Key;
+                string watchedLabel = changeWatcher.Label;
+                KeyValueIdentifier watchedKeyLabel = new KeyValueIdentifier(watchedKey, watchedLabel);
+
+                ConfigurationSetting watchedKvCopy = _serverData.Values.FirstOrDefault((kv) => kv.Key.Equals(watchedKey) && kv.Label == watchedLabel);
+
+                // If the key-value was found, store it for updating the settings
+                if (watchedKvCopy != null)
```
avanigupta, line 931: >if (watchedKvCopy != null) [](http://example.com/codeflow?start=16&length=26)

The reason watchedKvCopy could be null is when we have multiple refresh registered key-values with same key but different label. When we were populating _watchedSettings in `LoadKeyValuesRegisteredForRefresh`, we would add all refresh registered key-values to _watchedSettings and the overwritten key-values were not lost. That's what we want to achieve here too. We need to rethink how to extract the logic to update _watchedSettings from LoadKeyValuesRegisteredForRefresh, without losing the overwritten key-values.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -553,7 +560,7 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
                 UpdateCacheExpirationTime(changeWatcher);
             }
 
-            if (data != null)
+            if (data != null && watchedSettings != null)
```
avanigupta, line 563: Watched settings can be null when there are no refresh registered settings. We dont need to check for null here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -646,6 +655,7 @@ private async Task LoadKeyValuesRegisteredForRefresh(ConfigurationClient client,
                 if (data.TryGetValue(watchedKey, out ConfigurationSetting loadedKv)
                     && watchedKeyLabel.Equals(new KeyValueIdentifier(loadedKv.Key, loadedKv.Label)))
                 {
+                    watchedSettings[watchedKeyLabel] = new ConfigurationSetting(loadedKv.Key, loadedKv.Value, loadedKv.Label, loadedKv.ETag);
```
avanigupta, line 658: We need to call NormalizeNull() on `changeWatcher.Label` here

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -356,12 +360,9 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
                     {
                         _serverData = serverData;
+                        _watchedSettings = watchedSettings;
 
-                        if (refreshAll)
-                        {
-                            UpdateWatchedSettings();
-                        } 
-                        else
+                        if (!refreshAll)
```
avanigupta, line 365: Can you revert the changes to this code block since the underlying logic has not changed here. 
Just want to minimize the code we touch - less chance of introducing new bugs :)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -616,9 +641,9 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
             return serverData;
         }
 
-        private async Task LoadKeyValuesRegisteredForRefresh(ConfigurationClient client, IDictionary<string, ConfigurationSetting> data, CancellationToken cancellationToken)
+        private async Task<Dictionary<KeyValueIdentifier, ConfigurationSetting>> LoadKeyValuesRegisteredForRefresh(ConfigurationClient client, IDictionary<string, ConfigurationSetting> data, CancellationToken cancellationToken)
```
avanigupta, line 644: We need to remove the call to LoadKeyValuesRegisteredForRefresh inside LoadSelectedKeyValues

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -391,6 +398,21 @@ public AzureAppConfigurationOptions ConfigureKeyVault(Action<AzureAppConfigurati
             return this;
         }
 
+        /// <summary>
+        /// Transforms the configuration settings passed to the Azure App Configuration Provider before being passed to internal adapters.
```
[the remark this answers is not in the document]
jimmyca15: So maybe something like

`Provides a way to transform settings retrieved from App Configuration before they are processed by the configuration provider."

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -77,6 +79,11 @@ internal IEnumerable<IKeyValueAdapter> Adapters
             set => _adapters = value?.ToList();
         }
 
+        /// <summary>
+        /// A collection of user defined functions that transform each <see cref="ConfigurationSetting"/>.
+        /// </summary>
+        internal IEnumerable<Func<ConfigurationSetting, ValueTask<ConfigurationSetting>>> UserDefinedMappers => _mappers;
```
jimmyca15, line 85: Are there system defined mappers?

amerjusupovic: I think I just overlooked this, will change it to "Mappers"

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -29,7 +29,8 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
         private readonly bool _requestTracingEnabled;
         private readonly IConfigurationClientManager _configClientManager;
         private AzureAppConfigurationOptions _options;
-        private Dictionary<string, ConfigurationSetting> _applicationSettings;
+        private Dictionary<string, ConfigurationSetting> _serverData;
```
jimmyca15, line 32: I was wondering if we had to store `ConfigurationSetting` in there or if it could be a type that just has key, label and etag. Unfortunately, it looks like the SDK requires `ConfigurationSetting` as a param. I'd like to see if we could work with SDK team to get an overload that doesn't require us to keep the entire `ConfigurationSetting` object around.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -289,7 +292,11 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                             if (refreshAll)
                             {
                                 // Trigger a single load-all operation if a change was detected in one or more key-values with refreshAll: true
-                                applicationSettings = await LoadAll(client, cancellationToken).ConfigureAwait(false);
+                                // Preserve reference to loaded data with deep copy into serverData before calling mapper functions
+                                Dictionary<string, ConfigurationSetting> data = await LoadSelectedKeyValues(client, cancellationToken).ConfigureAwait(false);
+                                watchedSettings = await LoadKeyValuesRegisteredForRefresh(client, data, cancellationToken).ConfigureAwait(false);
```
jimmyca15, line 297: Is the change to watched settings required for Map API, or can it be fixed in a separate PR?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: I was wondering if we had to store `ConfigurationSetting` in there or if it could be a type that just has key, label and etag. Unfortunately, it looks like the SDK requires `ConfigurationSetting` as a param. I'd like to see if we could work with SDK team to get an overload that doesn't require us to keep the entire `ConfigurationSetting` object around.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -548,8 +570,11 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
 
                 try
                 {
-                    SetData(await PrepareData(data, cancellationToken).ConfigureAwait(false));
-                    _applicationSettings = data;
+                    _serverData = data.ToDictionary(kvp => kvp.Key, kvp => new ConfigurationSetting(kvp.Value.Key, kvp.Value.Value, kvp.Value.Label, kvp.Value.ETag));
```
jimmyca15, line 573: Shouldn't _serverData be the union of data and watchedSettings?

jimmyca15: I see, we are modifying data in `LoadKeyValuesRegisteredForRefresh`

jimmyca15: That's a bit hard to follow, may want to consider bringing that out.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Looks like we can with the [preview version of the SDK](https://learn.microsoft.com/en-us/dotnet/api/azure.data.appconfiguration.configurationclient.getconfigurationsettingasync?view=azure-dotnet-preview#azure-data-appconfiguration-configurationclient-getconfigurationsettingasync(system-string-system-string-system-string-system-collections-generic-ienumerable((system-string))-azure-matchconditions-azure-requestcontext))

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -289,7 +292,11 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                             if (refreshAll)
                             {
                                 // Trigger a single load-all operation if a change was detected in one or more key-values with refreshAll: true
-                                applicationSettings = await LoadAll(client, cancellationToken).ConfigureAwait(false);
+                                // Preserve reference to loaded data with deep copy into serverData before calling mapper functions
+                                Dictionary<string, ConfigurationSetting> data = await LoadSelectedKeyValues(client, cancellationToken).ConfigureAwait(false);
+                                watchedSettings = await LoadKeyValuesRegisteredForRefresh(client, data, cancellationToken).ConfigureAwait(false);
+                                serverData = data.ToDictionary(kvp => kvp.Key, kvp => new ConfigurationSetting(kvp.Value.Key, kvp.Value.Value, kvp.Value.Label, kvp.Value.ETag));
+                                mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
```
jimmyca15, line 299: `mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);` doesn't need to execute with failover policy, I'd suggest moving it to the if/else after failover policy execution.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Also, it doesn't look like we need to store every key-value in _serverData. It's only used for refreshing key-value collections.

Going further, if it's only being used for change detection, why isn't it merged with _watchedSettings?

amerjusupovic: Going off what we talked about last week, I noticed that trying to pass in null for the value parameter for the settings that I deep copied into watchedsettings ended up failing some unit tests.

jimmyca15: But we didn't aim to nullify values in this PR right?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -905,5 +905,46 @@ private bool IsFailOverable(RequestFailedException rfe)
...
+            Dictionary<string, ConfigurationSetting> mappedData = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);
+
+            foreach (KeyValuePair<string, ConfigurationSetting> kvp in data)
+            {
+                ConfigurationSetting setting = kvp.Value;
+                foreach (Func<ConfigurationSetting, ValueTask<ConfigurationSetting>> func in _options.Mappers)
+                {
+                    setting = await func(setting).ConfigureAwait(false);
+                }
+                mappedData[kvp.Key] = setting;
```
avanigupta, line 920: We should check if the setting is null before storing in mappedData. If its null, we can ignore this key.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -308,18 +308,23 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
-                        applicationSettings = new Dictionary<string, ConfigurationSetting>(_applicationSettings, StringComparer.OrdinalIgnoreCase);
+                        mappedData = new Dictionary<string, ConfigurationSetting>(_mappedData, StringComparer.OrdinalIgnoreCase);
                         watchedSettings = new Dictionary<KeyValueIdentifier, ConfigurationSetting>(_watchedSettings);
 
                         foreach (KeyValueChange change in keyValueChanges.Values.Concat(changedKeyValuesCollection))
                         {
                             if (change.ChangeType == KeyValueChangeType.Deleted)
                             {
-                                applicationSettings.Remove(change.Key);
+                                mappedData.Remove(change.Key);
```
avanigupta, line 318: We are running mappers on all KeyValueChanges before updating `_watchedSettings`. That means we are not creating deep copies of the settings in _watchedSettings dictionary when refreshAll is false.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -308,18 +308,23 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
                             }
                             else if (change.ChangeType == KeyValueChangeType.Modified)
                             {
-                                applicationSettings[change.Key] = change.Current;
+                                ConfigurationSetting setting = change.Current;
+                                foreach (Func<ConfigurationSetting, ValueTask<ConfigurationSetting>> func in _options.Mappers)
+                                {
+                                    setting = await func(setting).ConfigureAwait(false);
+                                }
+                                mappedData[change.Key] = setting;
```
avanigupta, line 327: Check for null `setting` before updating cache. If its null, delete the key from mappedData.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,17 +208,18 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
 
                     //
                     // Avoid instance state modification
-                    Dictionary<string, ConfigurationSetting> applicationSettings = null;
                     Dictionary<KeyValueIdentifier, ConfigurationSetting> watchedSettings = null;
                     Dictionary<KeyValueWatcher, KeyValueChange> keyValueChanges = null;
                     List<KeyValueChange> changedKeyValuesCollection = null;
+                    Dictionary<string, ConfigurationSetting> mappedData = null;
+                    Dictionary<string, ConfigurationSetting> data = null;
```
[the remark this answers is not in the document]
amerjusupovic: This was just my solution to moving the line `mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);` to outside of failover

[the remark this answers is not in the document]
avanigupta: You could replace `data` with `mappedData` everywhere and then do this outside failover:
`_mappedData = await MapConfigurationSettings(mappedData).ConfigureAwait(false);`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -346,7 +353,6 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
 
                     if (_options.Adapters.Any(adapter => adapter.NeedsRefresh()) || changedKeyValuesCollection?.Any() == true || keyValueChanges.Any())
                     {
-                        _applicationSettings = applicationSettings;
```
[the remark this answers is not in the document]
amerjusupovic: My fault, I think this was a result of merging incorrectly

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -673,11 +668,21 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
                 }
             }
 
+            foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
+            {
+                IEnumerable<ConfigurationSetting> currentKeyValues = GetCurrentKeyValueCollection(changeWatcher.Key, changeWatcher.Label, existingSettings.Values);
+
+                foreach (ConfigurationSetting setting in currentKeyValues)
+                {
+                    watchedSettings[new KeyValueIdentifier(setting.Key, setting.Label)] = new ConfigurationSetting(setting.Key, setting.Value, setting.Label, setting.ETag);
```
jimmyca15, line 682: So once the API that takes etag only comes out of preview, we won't need to store the entire setting right?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -290,8 +289,8 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                             if (refreshAll)
                             {
                                 // Trigger a single load-all operation if a change was detected in one or more key-values with refreshAll: true
-                                applicationSettings = await LoadSelectedKeyValues(client, cancellationToken).ConfigureAwait(false);
-                                watchedSettings = await LoadKeyValuesRegisteredForRefresh(client, applicationSettings, cancellationToken).ConfigureAwait(false);
+                                mappedData = await LoadSelectedKeyValues(client, cancellationToken).ConfigureAwait(false);
+                                watchedSettings = await LoadKeyValuesRegisteredForRefresh(client, mappedData, cancellationToken).ConfigureAwait(false);
```
jimmyca15, line 293: You're passing a parameter here named `mappedData`, but that's not really mapped data right?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -673,11 +668,21 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
                 // If the key-value was found, store it for updating the settings
                 if (watchedKv != null)
                 {
-                    watchedSettings[watchedKeyLabel] = watchedKv;
+                    watchedSettings[watchedKeyLabel] = new ConfigurationSetting(watchedKv.Key, watchedKv.Value, watchedKv.Label, watchedKv.ETag);
                     existingSettings[watchedKey] = watchedKv;
                 }
             }
 
+            foreach (KeyValueWatcher changeWatcher in _options.MultiKeyWatchers)
```
jimmyca15, line 676: This doesn't seem to have to do with loading the key-values registered for refresh.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -910,5 +898,49 @@ private bool IsFailOverable(RequestFailedException rfe)
...
+            }
+
+            return mappedData;
+        }
+
+        private IEnumerable<ConfigurationSetting> GetCurrentKeyValueCollection(string key, string label, IEnumerable<ConfigurationSetting> existingSettings)
+        {
+            IEnumerable<ConfigurationSetting> currentKeyValues;
+
+            if (key.EndsWith("*"))
```
jimmyca15, line 926: What's the case where we won't end up with it ending in '*'?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -574,9 +569,9 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
 
                 try
                 {
-                    SetData(await PrepareData(data, cancellationToken).ConfigureAwait(false));
-                    _applicationSettings = data;
                     _watchedSettings = watchedSettings;
+                    _mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
```
jimmyca15, line 573: What's the reasoning for the order switch? Initialization hasn't really finished until SetData has executed successfully, and `_mappedData` is used as the indicator for successful initialization.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -308,45 +307,38 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                                {
+                                    setting = await func(setting).ConfigureAwait(false);
+                                }
+                                if (setting == null)
+                                {
+                                    mappedData.Remove(change.Key);
+                                }
+                                else
+                                {
+                                    mappedData[change.Key] = setting;
```
jimmyca15, line 335: Shouldn't that be `setting.Key`?

If so, raises an interesting question. How do we know what to remove if they modify the key.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: It was originally named data but that meant having an extra variable defined at the top of the method, so now it's reusing mappedData. It is correct though that it's not mapped at this point, is that too misleading?

amerjusupovic: alright, I can change that back

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: From what I understand, it's when a key value is specified in the select method for feature flags options

jimmyca15: I see now. Unfortunate that we lump in single key refresh with collection based refresh. I'd like if we could rework the single key case to somehow use the single key refresh pipeline. That's out of the scope for this PR though.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -910,5 +898,49 @@ private bool IsFailOverable(RequestFailedException rfe)
...
+        }
+
+        private IEnumerable<ConfigurationSetting> GetCurrentKeyValueCollection(string key, string label, IEnumerable<ConfigurationSetting> existingSettings)
+        {
+            IEnumerable<ConfigurationSetting> currentKeyValues;
+
+            if (key.EndsWith("*"))
+            {
+                // Get current application settings starting with changeWatcher.Key, excluding the last * character
+                var keyPrefix = key.Substring(0, key.Length - 1);
```
jimmyca15, line 929: `var keyPrefix`

Do you mind to switch that to `string keyPrefix`. We usually prefer explicit type when the type itself is not used in the assignment.

E.g. we use the following

`var dict = new Dictionary<string, string>();

but not

`var foo = Bar();`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: I think I don't fully understand the reason for the ordering, which is why I put it last to save a couple extra lines of code. I can put setdata first again

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: I agree, I can rename the method to something like UpdateWatchedSettings instead

jimmyca15: I think the key vaule collection piece needs to be pulled out of this method.

amerjusupovic: Oh I see, I can work on that then

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,18 +209,19 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                    Dictionary<string, ConfigurationSetting> mappedData = null;
                     bool refreshAll = false;
                     StringBuilder logInfoBuilder = new StringBuilder();
                     StringBuilder logDebugBuilder = new StringBuilder();
 
                     await ExecuteWithFailOverPolicyAsync(availableClients, async (client) =>
                         {
-                            applicationSettings = null;
-                            keyValueChanges = new Dictionary<KeyValueIdentifier, KeyValueChange>();
+                            mappedData = null;
```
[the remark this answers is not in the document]
jimmyca15: Is it possible that you meant to nullify `data` here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Moving the code into a private method doesn't really address the concern. The state modification is still taking place in the `LoadKeyValuesRegisteredForRefresh` method.

By pulled out of the method, I meant that it should not happen in the `LoadKeyValuesRegisteredForRefresh` method.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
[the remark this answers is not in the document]
amerjusupovic: Yea I think I missed changing that back to what it was before, should be data. I can also declare mappedData later since it's not used until after ExecuteWithFailOverPolicyAsync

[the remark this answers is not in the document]
avanigupta: That's a good point. We probably dont even need the temp dictionary for mappedData because its always set outside ExecuteWithFailOverPolicyAsync. It looks like we could directly modify _mappedData at that point.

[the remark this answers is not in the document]
amerjusupovic: Oh I see, if mappedData was altered then we would always set _mappedData equal to it later anyway.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Ok sorry about that, so instead does it make sense to just create this new private method and call it each time after LoadKeyValuesRegisteredForRefresh is called so it's not taking place in that method?

jimmyca15: That works. Although the name shouldn't be `Load...` because it's not loading anything.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -208,18 +209,18 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                    Dictionary<string, ConfigurationSetting> data = null;
                     bool refreshAll = false;
                     StringBuilder logInfoBuilder = new StringBuilder();
                     StringBuilder logDebugBuilder = new StringBuilder();
 
                     await ExecuteWithFailOverPolicyAsync(availableClients, async (client) =>
                         {
-                            applicationSettings = null;
-                            keyValueChanges = new Dictionary<KeyValueIdentifier, KeyValueChange>();
+                            data = null;
```
amerjusupovic, line 222: Should I also be setting watchedSettings to null here following the same idea as data?

jimmyca15: Looks like with current usage the answer is no since it's only set right before exiting `ExecuteWithFailoverPolicyAsync`. But I wouldn't argue against adding it for symmetry purposes and just saying that the variables that are declared outside should be reset at the beginning of `ExecuteWithFailoverPolicyAsync`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: I almost feel like we need a wrapper around these three calls now. There are two places we need to call these methods in the same order. How about something like this:

```
LoadAndWatchConfigurationSettings()
{
    LoadSelectedKeyValues()
    LoadKeyValuesRegisteredForRefresh()
    UpdateWatchedKeyValueCollection()
}
```

avanigupta: Amer pointed out that we need to return two different dictionaries from these methods, data and watchedSettings. So to have a wrapper around the three methods, we would need to either return both dictionaries as a tuple/valuetuple, or make `LoadAndWatchConfigurationSettings` take out parameters. I think it would be better to return two dictionaries.  @jimmyca15 thoughts?

jimmyca15: I would just skip having the wrapper. Returning multiple types seems more complicated than having 3 method calls in two places.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -574,9 +575,11 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
 
                 try
                 {
-                    SetData(await PrepareData(data, cancellationToken).ConfigureAwait(false));
-                    _applicationSettings = data;
+                    Dictionary<string, ConfigurationSetting> mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
+                    SetData(await PrepareData(mappedData, cancellationToken).ConfigureAwait(false));
+                    watchedSettings = UpdateWatchedKeyValueCollections(watchedSettings, data);
```
jimmyca15, line 580: Wondering what the motivation was to put this here instead of right after `LoadKeyValuesRegisteredForRefresh`?

amerjusupovic: It can be inside of the try statement and might fit better with the pattern of calling all three in the same spot, willing to change that. I just saw that it didn't need to be inside of the ExecuteWithFailOverPolicyAsync call so I looked further down to where watchedsettings was used.

jimmyca15: It is fine to be here, but if it stays here i'd recommend doing `_watchedSettings = UpdateWatchedKeyValueCollections(watchedSettings, data);`

My initial expectation was it to be right after the try

jimmyca15: We call it in `ExecuteWithFailOverPolicyAsync` in the other call location. Perhaps it is a good idea to just move it in there, right after `LoadKeyValuesRegisteredForRefresh`

amerjusupovic: Oh that's true, makes sense thanks

jimmyca15 (approved):

avanigupta (approved):

PR #358, opened 2022-09-26
Remove select optimization

amerjusupovic:


tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -361,10 +364,10 @@ public void PreservesDefaultQuery()
             builder.AddAzureAppConfiguration(options =>
             {
                 options.Client = new ConfigurationClient(TestHelpers.CreateMockEndpointString(), clientOptions);
-                options.UseFeatureFlags();
+                options.UseFeatureFlags(o => o.Label = "myLabel");
```
avanigupta, line 367: If you add label for this test, it becomes exactly the same as `QueriesFeatureFlags`.
If you remove the label filter, this test verifies that even though Feature flag query will load the same label as default query, we have two requests corresponding to both default query and feature flag query.

avanigupta (approved):

PR #359, opened 2022-10-10
Fix Optional Parameter Bug for AddAzureAppConfiguration

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -597,7 +598,7 @@ private async Task SetData(IDictionary<string, ConfigurationSetting> data, bool
                 {
                     keyValuePairs = await ProcessAdapters(kvp.Value, cancellationToken).ConfigureAwait(false);
                 }
-                catch (KeyVaultReferenceException)
+                catch (Exception e) when (e is KeyVaultReferenceException || e is JsonException || e is FormatException || e is ArgumentNullException)
```
avanigupta, line 601: We only need to catch `FormatException` since that's what we throw while deserializing feature flags: 
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/aa782803912f5c898bb4cf48eeb1d37d50a08fbc/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs#L30-L33

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -13,6 +13,7 @@
 using System.Linq;
 using System.Net;
 using System.Security;
+using System.Text.Json;
```
avanigupta, line 16: Do you still need this reference?

avanigupta (approved):

PR #362, opened 2022-10-21
Fix watchedsettings logic

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -521,10 +524,15 @@ public void ProcessPushNotification(PushNotification pushNotification, TimeSpan?
...
             Dictionary<string, ConfigurationSetting> data = null;
+            Dictionary<KeyValueIdentifier, ConfigurationSetting> watchedSettings = null;
 
             try
             {
-                data = await ExecuteWithFailOverPolicyAsync(availableClients, (client) => LoadAll(client, cancellationToken), cancellationToken).ConfigureAwait(false);
+                await ExecuteWithFailOverPolicyAsync(availableClients, async (client) =>
+                {
+                    data = await LoadSelectedKeyValues(client, cancellationToken);
+                    watchedSettings = await LoadKeyValuesRegisteredForRefresh(client, data, cancellationToken);
```
avanigupta, line 534: .ConfigureAwait(false)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -228,7 +229,7 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                             foreach (KeyValueWatcher changeWatcher in cacheExpiredWatchers)
                             {
                                 string watchedKey = changeWatcher.Key;
-                                string watchedLabel = changeWatcher.Label;
+                                string watchedLabel = changeWatcher.Label.NormalizeNull();
```
jimmyca15, line 232: Why is `NormalizeNull` needed? I believe it already takes place inside of KeyValueIdentifier.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -209,6 +209,7 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                     //
                     // Avoid instance state modification
                     Dictionary<string, ConfigurationSetting> applicationSettings = null;
+                    Dictionary<KeyValueIdentifier, ConfigurationSetting> watchedSettings = new Dictionary<KeyValueIdentifier, ConfigurationSetting>(_watchedSettings);
```
jimmyca15, line 212: Why is this initialized here? I'd expect behavior similar to `applicationSettings`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: I thought the reason might be because we pass watchedLabel into GetConfigurationSettingAsync, if not then I can remove it

jimmyca15: Well the main branch today doesn't NormalizeNull here, so is this a bug you're fixing?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -344,6 +346,7 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                     if (_options.Adapters.Any(adapter => adapter.NeedsRefresh()) || changedKeyValuesCollection?.Any() == true || keyValueChanges.Any())
                     {
                         _applicationSettings = applicationSettings;
+                        _watchedSettings = watchedSettings;
```
jimmyca15, line 349: Where are we updating the watched setting for individual keys?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: It wasn't a specific bug, I was just told to call normalizenull on anywhere we used changewatcher.label in the map pr

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -611,26 +620,24 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
                 }).ConfigureAwait(false);
             }
 
-            // Load key-values registered for refresh that are not already loaded
-            await LoadKeyValuesRegisteredForRefresh(client, serverData, cancellationToken).ConfigureAwait(false);
             return serverData;
         }
 
-        private async Task LoadKeyValuesRegisteredForRefresh(ConfigurationClient client, IDictionary<string, ConfigurationSetting> data, CancellationToken cancellationToken)
+        private async Task<Dictionary<KeyValueIdentifier, ConfigurationSetting>> LoadKeyValuesRegisteredForRefresh(ConfigurationClient client, IDictionary<string, ConfigurationSetting> data, CancellationToken cancellationToken)
```
jimmyca15, line 626: I don't think the parameter name `data` is descriptive enough here. How about `existingSettings`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Right now it looks orthogonal to the watched settings update and also seems like it would have some behavioral change.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -521,10 +524,15 @@ public void ProcessPushNotification(PushNotification pushNotification, TimeSpan?
...
             Dictionary<string, ConfigurationSetting> data = null;
+            Dictionary<KeyValueIdentifier, ConfigurationSetting> watchedSettings = null;
 
             try
             {
-                data = await ExecuteWithFailOverPolicyAsync(availableClients, (client) => LoadAll(client, cancellationToken), cancellationToken).ConfigureAwait(false);
+                await ExecuteWithFailOverPolicyAsync(availableClients, async (client) =>
+                {
+                    data = await LoadSelectedKeyValues(client, cancellationToken).ConfigureAwait(false);
+                    watchedSettings = await LoadKeyValuesRegisteredForRefresh(client, data, cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
jimmyca15: I think it's fine. refactoring it out would require a way to preserve the label precedence currently established within the method by iterating over change watchers.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: From what I'm seeing, it may have been a bug introduced when the preview branch was created. But I'm not sure if I'm just missing it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: My bad, I missed that we are calling it inside KeyValueIdentifier. It can be removed from both places in this PR.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Its being updated in the loop right after this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -209,6 +209,7 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                     //
                     // Avoid instance state modification
                     Dictionary<string, ConfigurationSetting> applicationSettings = null;
+                    Dictionary<KeyValueIdentifier, ConfigurationSetting> watchedSettings = null;
```
jimmyca15, line 212: Should be cleared at the beginning of ExecuteWithFailOverPolicyAsync like `applicationSettings` is. The reason is that if it's not it's possible a failover could see half-modified state of previous attempt.

avanigupta (approved):

amerjusupovic:
Merging continues to be blocked despite checks succeeding, will open new PR

PR #365, opened 2022-10-28
Fix incorrect cache expiration for watched settings

amerjusupovic:
## Expected

When setting up App Configuration options, users are able to set a cache expiration interval for registered key-values and feature flags. Whenever a change is observed to one of these key-values in App Configuration or the cache expiration interval has elapsed, the provider will make a request to App Configuration to update that value. It will then reset the cache expiration interval back to the user-specified value. 

## Actual

After the first time the cache expiration interval elapses or a change is observed, the cache expiration interval is not updated back to its specified value unless the registered key-value has refreshAll set to true. If a key-value is registered with refreshAll set to false or is a feature flag, the provider is requesting App Configuration to update the value every time the application refreshes after the first cache expiration.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -311,7 +311,30 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                         applicationSettings = new Dictionary<string, ConfigurationSetting>(_applicationSettings, StringComparer.OrdinalIgnoreCase);
                         watchedSettings = new Dictionary<KeyValueIdentifier, ConfigurationSetting>(_watchedSettings);
 
-                        foreach (KeyValueChange change in keyValueChanges.Values.Concat(changedKeyValuesCollection))
+                        foreach (KeyValueWatcher changeWatcher in cacheExpiredWatchers.Concat(cacheExpiredMultiKeyWatchers))
+                        {
+                            UpdateCacheExpirationTime(changeWatcher);
+                        }
+
+                        foreach (KeyValuePair<KeyValueWatcher, KeyValueChange> kvp in keyValueChanges)
+                        {
+                            KeyValueChange keyValueChange = kvp.Value;
+                            KeyValueWatcher changeWatcher = kvp.Key;
+                            KeyValueIdentifier kvIdentifier = new KeyValueIdentifier(changeWatcher.Key, changeWatcher.Label);
+
+                            if (keyValueChange.ChangeType == KeyValueChangeType.Modified)
+                            {
+                                applicationSettings[keyValueChange.Key] = keyValueChange.Current;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -311,7 +311,30 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                         applicationSettings = new Dictionary<string, ConfigurationSetting>(_applicationSettings, StringComparer.OrdinalIgnoreCase);
                         watchedSettings = new Dictionary<KeyValueIdentifier, ConfigurationSetting>(_watchedSettings);
 
-                        foreach (KeyValueChange change in keyValueChanges.Values.Concat(changedKeyValuesCollection))
+                        foreach (KeyValueWatcher changeWatcher in cacheExpiredWatchers.Concat(cacheExpiredMultiKeyWatchers))
+                        {
+                            UpdateCacheExpirationTime(changeWatcher);
+                        }
+
+                        foreach (KeyValuePair<KeyValueWatcher, KeyValueChange> kvp in keyValueChanges)
```
avanigupta, line 319: Do we still need `keyValueChanges` to have `KeyValueWatcher` as dictionary key? Can we change the dictionary key to be `KeyValueIdentifier`?

amerjusupovic: Oh yes I agree, forgot to change

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -311,21 +311,42 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
-                                applicationSettings[change.Key] = change.Current;
+                                applicationSettings.Remove(keyValueChange.Key);
+                                watchedSettings.Remove(kvp.Key);
                             }
 
                             // Invalidate the cached Key Vault secret (if any) for this ConfigurationSetting
                             foreach (IKeyValueAdapter adapter in _options.Adapters)
                             {
-                                adapter.InvalidateCache(change.Current);
+                                adapter.InvalidateCache(keyValueChange.Current);
```
avanigupta, line 337: Historically, we have invalidated adapter cache whenever key-values or feature flags change. But I dont think we need to invalidate adapter cache for feature flag changes because we dont cache feature flags in Key Vault adapter, and all other adapters dont have any caching at all.
That being said, I dont see an issue with invalidating the adapter cache for feature flag changes - it would have no effect on anything, and it would prevent any bugs in future if we ever decide to implement `InvalidateCache` for feature management adapter. This behavior would also be consistent with what we have in the main and preview branches currently. 
Can you add the same cache invalidation logic to the `changedKeyValuesCollection` loop below?

avanigupta (approved):

PR #366, opened 2022-11-16
Add .net7 as target framework, build with .net7

amerjusupovic:


avanigupta:
Can you also update sample projects to target net7.0? Just want to make sure nothing is broken.

PR #367, opened 2022-11-21
Update logging with SDK model

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -415,7 +416,7 @@ public AzureAppConfigurationOptions Map(Func<ConfigurationSetting, ValueTask<Con
 
         private static ConfigurationClientOptions GetDefaultClientOptions()
         {
-            var clientOptions = new ConfigurationClientOptions(ConfigurationClientOptions.ServiceVersion.V1_0);
+            ConfigurationClientOptions clientOptions = new ConfigurationClientOptions(ConfigurationClientOptions.ServiceVersion.V1_0);
```
zhenlan, line 419: Seems unnecessary change

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Models;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.Tracing;
```
zhenlan, line 12: I don't see any other changes related to this namespace in this file. Is it needed?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -377,12 +378,26 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
 
                         if (logDebugBuilder.Length > 0)
                         {
-                            _logger?.LogDebug(logDebugBuilder.ToString().Trim());
+                            if (_logger != null)
+                            {
+                                _logger?.LogDebug(logDebugBuilder.ToString().Trim());
```
zhenlan, line 383: Now that you checked the _logger was not null, you don't need the ? anymore.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -439,11 +454,11 @@ public async Task<bool> TryRefreshAsync(CancellationToken cancellationToken)
             {
                 if (IsAuthenticationError(e))
                 {
-                    _logger?.LogWarning(e, LoggingConstants.RefreshFailedDueToAuthenticationError);
+                    AzureAppConfigurationProviderEventSource.Log.LogWarning(e, LoggingConstants.RefreshFailedDueToAuthenticationError);
```
zhenlan, line 457: I'm confused. Are we not logging with ILogger anymore even if it's available?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -377,12 +378,26 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
-                            _logger?.LogDebug(logDebugBuilder.ToString().Trim());
+                            if (_logger != null)
+                            {
+                                _logger?.LogDebug(logDebugBuilder.ToString().Trim());
+                            } 
+                            else
+                            {
+                                AzureAppConfigurationProviderEventSource.Log.LogDebug("\n\t" + logDebugBuilder.ToString().Trim());
+                            }
                         }
```
zhenlan, line 389: I hope we can put most of these logic in a helper function. This code doesn't need to know which logger it's using.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -273,8 +273,9 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                                 // Check if a change has been detected in the key-value registered for refresh
                                 if (change.ChangeType != KeyValueChangeType.None)
                                 {
-                                    logDebugBuilder.AppendLine($"{LoggingConstants.RefreshKeyValueChanged}(key: '{change.Key}', label: '{change.Label}')");
-                                    logInfoBuilder.AppendLine($"{LoggingConstants.RefreshKeyValueSettingUpdated}'{change.Key}' from endpoint {endpoint}");
+                                    string changeTypeString = change.ChangeType == KeyValueChangeType.Modified ? "Modified" : "Deleted";
+                                    logDebugBuilder.AppendLine($"{LoggingConstants.RefreshKeyValueLoaded} Change: {changeTypeString}. Key: '{change.Key}'. Label: '{change.Label}'.");
+                                    logInfoBuilder.AppendLine($"{LoggingConstants.RefreshKeyValueSettingUpdated} Key: '{change.Key}'. Endpoint: [ {endpoint} ].");
```
zhenlan, line 278: We are trying to establish some structured logging formatting. It's good to keep that logging format in a central place instead of inline here.

zhenlan:
>     <TargetFramework>net5.0</TargetFramework>

net5.0 is out of support. We should update the example to net6.0 or net7.0. How about you do that in a separate PR including any dependency package update?

---
Refers to: examples/ConsoleAppWithFailOver/ConsoleAppWithFailOver.csproj:5 in a1d8039. [](commit_id = a1d80397d01f0d3e8ba13f34846bbf8c755cf389, deletion_comment = False)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: I was just told that the convention was to use the type if there was an explicit type being used for the initialization. I don't mind taking it out if it's unnecessary though.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: I'll add a helper to structure messages and solve this in those next changes.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
zhenlan: Declare the type explicitly when you cannot tell what the type is. In this case, it's clear what the type is from the new operator.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
```diff
@@ -23,12 +23,6 @@ public AzureAppConfigurationRefresherProvider(IConfiguration configuration, ILog
                 {
                     if (provider is IConfigurationRefresher refresher)
                     {
-                        // Use _loggerFactory only if LoggerFactory hasn't been set in AzureAppConfigurationOptions
```
avanigupta, line 26: This is where we get the logger factory from DI and assign it to the `IConfigurationRefresher` refresher, which is actually an instance of `AzureAppConfigurationProvider`. 
If we are deprecating LoggerFactory from `IConfigurationRefresher`, we would need to cast this `provider` to `AzureAppConfigurationProvider` and assign `_loggerFactory` obtained from DI to the concrete instance of `AzureAppConfigurationProvider`. Otherwise, `AzureAppConfigurationProvider.LoggerFactory` will always be null.

```
if (provider is AzureAppConfigurationProvider appConfigurationProvider)
{
    appConfigurationProvider.LoggerFactory = _loggerFactory;                 
    refreshers.Add(appConfigurationProvider);
}
```

@jimmyca15, do you see an issue with casting to `AzureAppConfigurationProvider` instead of `IConfigurationRefresher`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/LoggingExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            if (endpoint != null)
+            {
+                fullLogMessage.Append($" Endpoint:'{endpoint.TrimEnd('/')}'");
+            }
+
+            return fullLogMessage.ToString();
+        }
+
+        public static void HandleLog(ILogger logger, LogLevel logLevel, Exception e, string message)
```
avanigupta, line 37: We should split it into LogWarning, LogInfo and LogDebug. Keeps the code easy to understand and avoids unnecessary input arguments (like Exception e), which are not applicable to 2 out of 3 log types here.

examples/ConsoleAppWithFailOver/ConsoleAppWithFailOver.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="1.5.0" />
+    <PackageReference Include="Azure.Identity" Version="1.7.0" />
```
avanigupta, line 9: Does this need to be updated as part of this PR?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -10,20 +10,19 @@ internal class LoggingConstants
...
-        public const string RefreshConfigurationUpdatedSuccess = "Configuration reloaded for all selected key-values from endpoint: ";
-        public const string RefreshKeyValueChanged = "Change detected for key-value ";
-        public const string RefreshKeyValueUnchanged = "No change detected for key-value ";
-        public const string RefreshKeyValueSettingUpdated = "Value updated for key: ";
-        public const string RefreshKeyVaultSecretChanged = "Secret loaded from Key Vault for key-value ";
-        public const string RefreshKeyVaultSettingUpdated = "Value updated from Key Vault for key: ";
-        public const string RefreshFeatureFlagChanged = "Change detected for feature flag ";
-        public const string RefreshFeatureFlagsUnchanged = "No change detected for feature flags.";
-        public const string RefreshFeatureFlagValueUpdated = "Value updated for feature flag: ";
+        public const string RefreshConfigurationUpdatedSuccess = "Configuration reloaded.";
```
avanigupta, line 18: I know we have a separate section on top for error messages, which are all logged as warnings. But for success, we sometimes print info and sometimes debug. It would be helpful to remember which statements are used for debug logs vs info logs. Can you split this into sections for info vs debug?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProviderRefreshEventSource.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    [EventSource(Name = EventSourceName)]
+    internal class AzureAppConfigurationProviderRefreshEventSource : EventSource
+    {
+        /// <summary>The name to use for the event source.</summary>
+        private const string EventSourceName = "Microsoft.Extensions.Configuration.AzureAppConfiguration.Refresh[0]";
```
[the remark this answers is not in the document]
amerjusupovic: Oh yes I spaced out on that, it should be dashes

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProviderRefreshEventSource.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                LogWarning(message + " " + e.Message);
+            }
+            else
+            {
+                LogWarning(message);
+            }
+        }
+
+        protected AzureAppConfigurationProviderRefreshEventSource()
```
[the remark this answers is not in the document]
amerjusupovic: Yea that's odd, think I just copied it from an sdk example and forgot to double check

examples/ConsoleAppWithFailOver/ConsoleAppWithFailOver.csproj
amerjusupovic: It has some compile errors if I don't add it but it doesn't seem like it happens on preview branch without my changes. That's why I wasn't really sure if I should make a separate PR.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/LoggingExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                fullLogMessage.Append($" Change:'{changeType}'");
+            }
+
+            if (key != null)
+            {
+                fullLogMessage.Append($" Key:'{key}'");
+            }
+
+            if (label != LabelFilter.Null)
```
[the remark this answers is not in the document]
amerjusupovic: Oh I see, I was using LabelFilter.Null as a null value to get around this, wasn't sure if that's a bad idea

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
```diff
@@ -17,10 +17,11 @@
     <PackageReference Include="Azure.Data.AppConfiguration" Version="1.2.0" />
     <PackageReference Include="Azure.Messaging.EventGrid" Version="4.7.0" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.0.1" />
+    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.6.0" />
```
avanigupta, line 20: Why is this package added? I think this is the package that requires Azure.Identity to be 1.7.0 or higher.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProviderRefreshEventSource.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <summary>
+        ///   Provides a singleton instance of the event source for callers to
+        ///   use for logging.
+        /// </summary>
+        public static AzureAppConfigurationProviderRefreshEventSource Log { get; } = new AzureAppConfigurationProviderRefreshEventSource();
+
+        [Event(LogDebugEvent, Message = "{0}", Level = EventLevel.Verbose)]
+        public void LogDebug(string message)
+        {
+            WriteEvent(1, message);
```
avanigupta, line 25:
```suggestion
            WriteEvent(LogDebugEvent, message);
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProviderRefreshEventSource.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        [Event(LogWarningEvent, Message = "{0}", Level = EventLevel.Warning)]
+        public void LogWarning(string message, Exception e)
+        {
+            if (e != null)
+            {
+                message += " " + e.Message;
+            }
+            WriteEvent(3, message);
```
avanigupta, line 41: Nit: empty line after closing bracket

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProviderRefreshEventSource.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        private AzureAppConfigurationProviderRefreshEventSource()
+           : base(
+                EventSourceName,
+                EventSourceSettings.Default,
+                "AzureEventSource",
```
avanigupta, line 48: Since these are constant string values, can we define these traits on top too?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
jimmyca15: No. I don't see a problem.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/LoggingExtensions.cs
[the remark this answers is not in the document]
jimmyca15: I'd go further and suggest having a method for each type of format we want.

`LogHelper.BuildKeyValueChangeMessage(changeType, key, label)` Being over generic can induce limitations by trying to satisfy all the different call-sites, and reduce flexibility.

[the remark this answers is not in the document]
avanigupta: Our new logs will always have the same "structure", right? So we know that debug logs always need "Change", "Label" and "Endpoint" arguments but info logs dont need anything except "Key".
If we just split the formatting helper into multiple little helpers, each accepting only the input arguments it needs, we can avoid using default arguments to deduce what needs to be logged.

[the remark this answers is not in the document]
jimmyca15: I don't see the relation between debug logs and those parameters. If they all take the parameter, it seems happenstance.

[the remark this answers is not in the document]
avanigupta: I didnt refresh the page, so I didnt see your earlier comment and thought I was responding to Amer :) 
I like the idea of `LogHelper.BuildKeyValueChangeMessage`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProviderRefreshEventSource.cs
amerjusupovic: My fault I made this change and undid it on accident

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
amerjusupovic: Oh this was my mistake, I originally had a different approach in this PR but this isn't needed now. Thanks!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/LoggingExtensions.cs
```diff
@@ -0,0 +1,75 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+using System.Text;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+    internal static class LoggingExtensions
```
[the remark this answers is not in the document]
jimmyca15: I agree with the idea that these don't really belong as extension methods. However, using a static class means we have to pass around ILogger alot which is a little cumbersome. Was it considered to have a class that takes optionally takes ILogger and wraps it to avoid passing it every time?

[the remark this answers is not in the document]
amerjusupovic: I hadn't really considered it, does that mean that class would get stored in AzureAppConfigurationProvider?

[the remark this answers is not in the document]
jimmyca15: > does that mean that class would get stored in AzureAppConfigurationProvider?

It would be it's own class that `AzureAppConfigurationProvider` instantiates.

```
class Logger
{
    public Logger();

    public Logger(ILogger logger);

    public LogInformation(string message);

    public LogWarning(string message, Exception e);
}
```

Usage
```
_logger = logger != null ?
    new Logger(logger) :
    new Logger();
```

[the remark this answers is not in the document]
amerjusupovic: Ok thanks that's what I was expecting. Don't see any problem with this at the moment.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProviderRefreshEventSource.cs
```diff
@@ -0,0 +1,55 @@
...
+    [EventSource(Name = EventSourceName)]
+    internal class AzureAppConfigurationProviderRefreshEventSource : EventSource
+    {
+        /// <summary>The name to use for the event source.</summary>
+        private const string EventSourceName = "Microsoft-Extensions-Configuration-AzureAppConfiguration-Refresh";
+
+        private const int LogDebugEvent = 1;
+        private const int LogInformationEvent = 2;
+        private const int LogWarningEvent = 3;
+
```
avanigupta, line 16: The more I read about EventId ([1](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventattribute.eventid?view=net-7.0)), ([2](https://github.com/microsoft/dotnet-samples/blob/master/Microsoft.Diagnostics.Tracing/EventSource/docs/EventSource.md)), ([3](https://github.com/Azure/azure-sdk-for-net/blob/c915bdae1ad9ec44f1435faade3b0b25e4542ded/sdk/core/Azure.Core/src/Diagnostics/AzureCoreEventSource.cs#L14-L34)), the more I feel like "LogDebug", "LogInfo" and "LogWarning" are not really events - they are just log levels. Events are what we are describing in the log message - i.e., each message in the LoggingConstants class should be a separate event. 
For example, "RefreshKeyValueSettingUpdated" should be a unique event called "RefreshKeyValueSettingUpdatedEvent" and should have its own method in this file. 

```
[Event(RefreshKeyValueSettingUpdatedEvent, Level = EventLevel.Informational, Message = "Setting updated. Key:'{0}'")]
public void RefreshKeyValueSettingUpdated(string key)
{
    WriteEvent(RefreshKeyValueSettingUpdatedEvent, key);
}
```

Then the usage would look like this:
```
AzureAppConfigurationProviderRefreshEventSource.Log.RefreshKeyValueSettingUpdated(key);
```
It's also interesting that it might be possible to filter EventSource logs based on the exact EventId, so users dont have to rely on filtering the log message strings if they want to monitor our logs. It does require some changes to how we currently keep track of what needs to be logged inside ExecuteWithFailOverPolicyAsync method. 
@jimmyca15 do you think it's worth the effort?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Logger.cs
```diff
@@ -0,0 +1,97 @@
...
+            {
+                AzureAppConfigurationProviderRefreshEventSource.Log.LogInformation(message);
+            }
+        }
+
+        public void LogWarning(string message, Exception e)
+        {
+            if (_logger != null)
+            {
+                _logger.LogWarning(message);
```
avanigupta, line 89: Exception is not being logged.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -452,23 +453,23 @@ public async Task<bool> TryRefreshAsync(CancellationToken cancellationToken)
...
             catch (KeyVaultReferenceException e)
             {
-                _logger?.LogWarning(e, LoggingConstants.RefreshFailedDueToKeyVaultError);
+                _logger.LogWarning(LoggingConstants.RefreshFailedDueToKeyVaultError, e);
                 return false;
             }
             catch (OperationCanceledException)
             {
-                _logger?.LogWarning(LoggingConstants.RefreshCanceledError);
+                _logger.LogWarning(LoggingConstants.RefreshCanceledError, null);
```
avanigupta, line 472: If we have to pass null for an argument, there is scope for improving that method. 
In this case, we can change LogWarning method to accept just the final message and pre-build the message from exception before calling LogWarning.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -524,7 +525,7 @@ public void ProcessPushNotification(PushNotification pushNotification, TimeSpan?
             }
             else
             {
-                _logger.LogWarning($"Ignoring the push notification received for the unregistered endpoint '{pushNotification.ResourceUri}'");
+                _logger.LogWarning($"Ignoring the push notification received for the unregistered endpoint '{pushNotification.ResourceUri}'.", null);
```
avanigupta, line 528: Looks like we missed putting this error message in the LoggingConstants class.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IKeyValueAdapter.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     internal interface IKeyValueAdapter
     {
-        Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, ILogger logger, CancellationToken cancellationToken);
+        Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, Logger logger, CancellationToken cancellationToken);
```
avanigupta, line 14: Remove usused using statement

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProviderRefreshEventSource.cs
jimmyca15: @avanigupta I do agree that the log level should not be confused with the event. If we're not explicitly passing event ids, we should hardcode 0 for the event id. Similar to [built-in usage](https://github.com/aspnet/Logging/blob/2d2f31968229eddb57b6ba3d34696ef366a6c71b/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs#L377).

At the moment I wouldn't want to scope-creep and start coming up with unique event ids. We can always update the event ids in the future.

amerjusupovic: Based on the first link mentioned it seems like event id has to be greater than 0 for EventSource. It could be possible to hardcode a different value and pass that to ILogger too.

jimmyca15: Sure, then can use 1.

avanigupta: Sounds good, let's use 1 for all events.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -525,7 +528,7 @@ public void ProcessPushNotification(PushNotification pushNotification, TimeSpan?
             }
             else
             {
-                _logger.LogWarning($"Ignoring the push notification received for the unregistered endpoint '{pushNotification.ResourceUri}'.", null);
+                _logger.LogWarning(LoggingConstants.PushNotificationUnregisteredEndpoint + pushNotification.ResourceUri + "'.");
```
avanigupta, line 531:
```suggestion
                _logger.LogWarning(LoggingConstants.PushNotificationUnregisteredEndpoint + pushNotification.ResourceUri + ".");

```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/DynamicFeatureKeyValueAdapter.cs
```diff
@@ -16,7 +16,7 @@ internal class DynamicFeatureKeyValueAdapter : IKeyValueAdapter
         private static readonly string DynamicFeatureSectionPrefix =
             $"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.DynamicFeatureSectionName}";
                                                                     
-        public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, ILogger logger, CancellationToken cancellationToken)
+        public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, Logger logger, CancellationToken cancellationToken)
```
avanigupta, line 19: Remove Microsoft.Extensions.Logging?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Logger.cs
```diff
@@ -0,0 +1,92 @@
...
+        ILogger _logger;
+
+        public Logger() { }
+
+        public Logger(ILogger logger)
+        {
+            _logger = logger;
+        }
+
+        public static string BuildKeyValueReadMessage(KeyValueChangeType changeType, string key, string label, string endpoint)
```
avanigupta, line 16: Can we move the static helper methods to a different class like LogHelper?

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -928,19 +940,18 @@ public void ValidateCorrectFeatureFlagLoggedIfModifiedOrRemovedDuringRefresh()
             eTag: new ETag("c3c231fd-39a0-4cb6-3237-4614474b92c1" + "f"));
 
             Thread.Sleep(CacheExpirationTime);
-            refresher.LoggerFactory = mockLoggerFactory.Object;
             refresher.TryRefreshAsync().Wait();
             Assert.Equal("AllUsers", config["FeatureManagement:MyFeature:EnabledFor:0:Name"]);
-            Assert.True(TestHelpers.ValidateLog(mockLogger, LoggingConstants.RefreshFeatureFlagChanged + "(key: 'myFeature1', label: '')", LogLevel.Debug));
-            Assert.True(TestHelpers.ValidateLog(mockLogger, LoggingConstants.RefreshFeatureFlagValueUpdated + "'myFeature1'", LogLevel.Information));
+            Assert.Contains(LoggingConstants.RefreshFeatureFlagRead + " Key:'myFeature1' Label:''", verboseInvocation);
```
avanigupta, line 945: We might benefit from reusing the helper methods like BuildFeatureFlagReadMessage to validate the logged errors. That way, if we change the formatting of error messages, we dont have to update all the tests.

amerjusupovic: Do you mean creating a helper for each kind of error message as well? I agree with using the existing helpers to validate logging tests but not sure what you meant by reusing the feature flag method for logged errors.

avanigupta: I meant that we could do something like this:
`Assert.Equal(verboseInvocation, LogHelper.BuildFeatureFlagReadMessage("myFeature1", null))`

amerjusupovic: Ah ok understood, just thought you meant something else as well

avanigupta (approved):
Can you add the Copyright header in all new files?

avanigupta:
@jimmyca15 @zhenlan can you take another look at this PR so it can be closed?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Logger.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    internal class Logger
+    {
+        ILogger _logger;
+
+        public Logger() { }
+
+        public Logger(ILogger logger)
+        {
+            _logger = logger;
```
jimmyca15, line 16: The expectation of this constructor is `logger` is not null right? If so, needs null check.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -524,7 +528,7 @@ public void ProcessPushNotification(PushNotification pushNotification, TimeSpan?
             }
             else
             {
-                _logger.LogWarning($"Ignoring the push notification received for the unregistered endpoint '{pushNotification.ResourceUri}'");
+                _logger.LogWarning(LoggingConstants.PushNotificationUnregisteredEndpoint + pushNotification.ResourceUri + ".");
```
jimmyca15, line 531: Some places use `Build` paradigm, some don't. What dictates the decision?

jimmyca15 (approved):

PR #371, opened 2022-12-12
Fix configuration refresh when ChainedConfigurationProvider is present for .NET7

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
```diff
@@ -26,20 +39,24 @@ public AzureAppConfigurationRefresherProvider(IConfiguration configuration, ILog
...
                         refreshers.Add(refresher);
                     }
-                }
-            }
 
-            if (!refreshers.Any())
-            {
-                throw new InvalidOperationException("Unable to access the Azure App Configuration provider. Please ensure that it has been configured correctly.");
-            }
+                    PropertyInfo p = typeof(ChainedConfigurationProvider).GetProperty("Configuration", BindingFlags.Public | BindingFlags.Instance);
```
avanigupta, line 48: We need to do this only if the provider is ChainedConfigProvider. I think it can be placed inside an `else if` block:
```
if (provider is IConfigurationRefresher refresher)
{

}
else if (provider is ChainedConfigurationProvider chainedProvider)
{
  // Reflection
}
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
```diff
@@ -26,20 +39,24 @@ public AzureAppConfigurationRefresherProvider(IConfiguration configuration, ILog
...
 
-            Refreshers = refreshers;
+                    if (p != null)
```
avanigupta, line 50: nit: `propertyInfo` instead of `p`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
```diff
@@ -26,20 +39,23 @@ public AzureAppConfigurationRefresherProvider(IConfiguration configuration, ILog
...
+                            refresher.LoggerFactory = loggerFactory;
                         }
 
                         refreshers.Add(refresher);
                     }
-                }
-            }
+                    else if (provider is ChainedConfigurationProvider chainedProvider)
+                    {
+                        PropertyInfo propertyInfo = typeof(ChainedConfigurationProvider).GetProperty("Configuration", BindingFlags.Public | BindingFlags.Instance);
```
jimmyca15, line 49: Why reflection?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
```diff
@@ -26,20 +39,23 @@ public AzureAppConfigurationRefresherProvider(IConfiguration configuration, ILog
...
 
-            if (!refreshers.Any())
-            {
-                throw new InvalidOperationException("Unable to access the Azure App Configuration provider. Please ensure that it has been configured correctly.");
+                        if (propertyInfo != null)
+                        {
+                            var chainedProviderConfigurationRoot = propertyInfo.GetValue(chainedProvider) as IConfigurationRoot;
```
jimmyca15, line 53: `chainedProviderConfigurationRoot` could end up null here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
amerjusupovic: I wasn't really sure how to approach this, I'm only aware of reflection or using predefined symbols like `#if NET7_0` to identify the version. There just has to be a way to identify whether this version has the `Configuration` property.

jimmyca15: The two approaches have different behaviors at runtime. Do we need this to happen on runtimes other than .net 7?

avanigupta: If we use net7_0 macro, we wouldn't be able to extend this feature to future versions of .NET, right?

zhenlan: There are all kinds of problems with version detection, so feature detection (using reflection on public API) is normally preferred. Please make sure the reflection won't occur repeatedly.

amerjusupovic: There is a net7_0_or_greater symbol for the version issue, but since it sounds like that's not preferred I'll go with reflection. I don't think it's called more than needed right now for this issue.

avanigupta: PropertyInfo can be a private field so that we only do this feature detection once.

amerjusupovic: Oh that makes sense actually

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
```diff
@@ -26,20 +41,21 @@ public AzureAppConfigurationRefresherProvider(IConfiguration configuration, ILog
                         // Use _loggerFactory only if LoggerFactory hasn't been set in AzureAppConfigurationOptions
                         if (refresher.LoggerFactory == null)
                         {
-                            refresher.LoggerFactory = _loggerFactory;
+                            refresher.LoggerFactory = loggerFactory;
```
avanigupta, line 44: We dont need to check for null anymore. This comment is not applicable because there is no other way to set LoggerFactory now.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
```diff
@@ -5,18 +5,33 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     internal class AzureAppConfigurationRefresherProvider : IConfigurationRefresherProvider
     {
+        private PropertyInfo _propertyInfo = typeof(ChainedConfigurationProvider).GetProperty("Configuration", BindingFlags.Public | BindingFlags.Instance);
```
avanigupta, line 14: This can be made static readonly

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -1132,6 +1127,42 @@ public void RefreshTests_UpdateCacheExpirationTimeForFailedRefreshOperations()
...
+                {
+                    Configuration = new ConfigurationBuilder()
+                        .AddAzureAppConfiguration(optionsInitializer, optional: true)
+                        .Build()
+                })
+                .Build()
+            }
+            .Build(new ConfigurationBuilder()) as ChainedConfigurationProvider;
+
+            PropertyInfo propertyInfo = typeof(ChainedConfigurationProvider).GetProperty("Configuration", BindingFlags.Public | BindingFlags.Instance);
```
avanigupta, line 1147: I dont think we need reflection in unit tests. We can test the following:
- Create IConfiguration that loads data from AppConfig.
- Create another IConfiguration that loads data using the previous config. 
- Now create  AzureAppConfigurationRefresherProvider with the second config. This refresher provider should have a single refresher.

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
```diff
@@ -24,6 +24,14 @@
     <ProjectReference Include="..\..\src\Microsoft.Extensions.Configuration.AzureAppConfiguration\Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net7.0'">
+	<PackageReference Include="Microsoft.Extensions.Configuration" Version="7.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' != 'net7.0'">
+	<PackageReference Include="Microsoft.Extensions.Configuration" Version="5.0.0" />
```
avanigupta, line 32: This package is only needed for net7.0 tests.

tests/Tests.AzureAppConfiguration/RefreshTests.cs
```diff
@@ -12,6 +12,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
```
avanigupta, line 15: Remove unused package

avanigupta (approved):

avanigupta (approved):

Issue #312, opened 2022-02-26
Conditionally add App Config provider depending on the environment that the app is running in


avanigupta:
### Background:

In some situations, users may want to load their configuration differently based on the app's current hosting environment. Some examples include:
1. Load key-values from a different AppConfig store in different environments
2. Load key-values with a different label in different environments
3. Don't load anything from AppConfig in development environment

Currently, users can do this by checking the hosting environment themselves. For example: 
```
.ConfigureAppConfiguration((hostingContext, config) => 
{ 
    if (hostingContext.HostingEnvironment.IsDevelopment()) 
    { 
        config.AddAzureAppConfiguration(options => 
        { 
            options.Connect(appconfig_endpoint, credentials) 
                   .Select(KeyFilter.Any, "dev"); 
        }); 
    } 
    else
    { 
        config.AddAzureAppConfiguration(options => 
        { 
            options.Connect(appconfig_endpoint, credentials) 
                   .Select(KeyFilter.Any, "prod"); 
        }); 
    } 
} 
```

### New API Proposal:

We want to make this easier and more declarative by introducing a new API that takes the name of the environment for which the AppConfig provider options are applicable. 

New API to be added in AzureAppConfigurationExtensions: 
```
public static IConfigurationBuilder AddAzureAppConfiguration( 
            this IConfigurationBuilder configurationBuilder, 
            string environmentName, 
            Action<AzureAppConfigurationOptions> action, 
            bool optional = false) 
```

### Environment name:

To determine the current hosting environment, we will read the value of these environment variables:
- DOTNET_ENVIRONMENT
- ASPNETCORE_ENVIRONMENT

The ASPNETCORE_ENVIRONMENT value overrides any value defined in DOTNET_ENVIRONMENT. 

### Sample Usage:

This sample code will:
- not load any key-values from AppConfig when the environment is set to "Development".
- only load key-values from `stage_endpoint` AppConfig store when the environment is set to "Staging"
- only load key-values from `prod_endpoint` AppConfig store when the environment is set to "Production"

```
.ConfigureAppConfiguration((hostingContext, config) => 
{ 
    config.AddAzureAppConfiguration(
    environmentName: "Staging",
    options => 
    { 
        options.Connect(stage_endpoint, stage_creds) 
               .Select(KeyFilter.Any, LabelFilter.Null); 
    }); 

    config.AddAzureAppConfiguration(
    environmentName: "Production",
    options => 
    { 
[...]

PR #372, opened 2022-12-22
Conditionally add App Config provider depending on the environment that the app is running in

amerjusupovic:


amerjusupovic:
Could add a log message when key-values aren't loaded or just print the environment name to let the user know

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -50,7 +54,7 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
                     throw new ArgumentException($"Please call {nameof(AzureAppConfigurationOptions)}.{nameof(AzureAppConfigurationOptions.Connect)} to specify how to connect to Azure App Configuration.");
                 }
 
-                provider = new AzureAppConfigurationProvider(client, options, _optional);
+                provider = new AzureAppConfigurationProvider(client, options, _environmentName, _environmentNameEnabled, _optional);
```
avanigupta, line 57: Do we want to create an instance of AzureAppConfigurationProvider even if the environment doesn't match?
Can we do the environment variable checks here instead of doing them in the constructor of AzureAppConfigurationProvider?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -11,15 +11,19 @@ internal class AzureAppConfigurationSource : IConfigurationSource
         private readonly bool _optional;
         private readonly Func<AzureAppConfigurationOptions> _optionsProvider;
         private readonly IConfigurationClientFactory _configurationClientFactory;
+        private readonly string _environmentName;
+        private readonly bool _environmentNameEnabled;
 
-        public AzureAppConfigurationSource(Action<AzureAppConfigurationOptions> optionsInitializer, bool optional = false, IConfigurationClientFactory configurationClientFactory = null)
+        public AzureAppConfigurationSource(Action<AzureAppConfigurationOptions> optionsInitializer, string environmentName = null, bool environmentNameEnabled = false, bool optional = false, IConfigurationClientFactory configurationClientFactory = null)
```
avanigupta, line 17: We dont need `environmentNameEnabled` flag. We can check `environmentName == null` wherever needed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -82,12 +83,26 @@ public ILoggerFactory LoggerFactory
...
             _client = client ?? throw new ArgumentNullException(nameof(client));
             _options = options ?? throw new ArgumentNullException(nameof(options));
             _optional = optional;
 
+            if (environmentNameEnabled)
+            {
+                try
+                {
+                    string envType = Environment.GetEnvironmentVariable(RequestTracingConstants.AspNetCoreEnvironmentVariable) ??
+                                        Environment.GetEnvironmentVariable(RequestTracingConstants.DotNetCoreEnvironmentVariable);
```
avanigupta, line 97: Starting with .NET 7, the precedence order of these environment variables has changed: [ASPNET-prefixed environment variable precedence](https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/7.0/environment-variable-precedence)  

The upside of following the same logic is that our environment detection behavior will be consistent with dotnet. The downside is that it would add complexity to the implementation, since we also need to detect the type of host to make this decision: [Environments](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-7.0#environments)

@jimmyca15 @zhenlan  thoughts?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
```diff
@@ -25,6 +25,23 @@ public static class AzureAppConfigurationExtensions
...
+        /// Adds key-value data from an Azure App Configuration store to a configuration builder.
+        /// </summary>
+        /// <param name="configurationBuilder">The configuration builder to add key-values to.</param>
+        /// <param name="environmentName">The name of the environment for which the Azure App Configuration options are applicable.</param>
+        /// <param name="action">A callback used to configure Azure App Configuration options.</param>
+        /// <param name="optional">Determines the behavior of the App Configuration provider when an exception occurs. If false, the exception is thrown. If true, the exception is suppressed and no settings are populated from Azure App Configuration.</param>
+        /// <returns>The provided configuration builder.</returns>
+        public static IConfigurationBuilder AddAzureAppConfiguration(
+            this IConfigurationBuilder configurationBuilder,
+            string environmentName,
```
avanigupta, line 40: Can you add null check for environmentName?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
amerjusupovic: I guess I should've asked before, but is passing null for environmentName not a valid option? In other words, is there no scenario where the environment variables would both be null and the user could pass that in to check specifically for null? That was the only reason I added this.

avanigupta: From usage perspective, "null" environment doesn't make sense. Dotnet also defaults to "Production" environment if the environment is not explicitly set ([reference](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-7.0#environments)). 
So I think we should explicitly forbid null as an environment name.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -82,7 +82,7 @@ public ILoggerFactory LoggerFactory
             }
         }
 
-        public AzureAppConfigurationProvider(ConfigurationClient client, AzureAppConfigurationOptions options, bool optional)
+        public AzureAppConfigurationProvider(ConfigurationClient client, AzureAppConfigurationOptions options, string environmentName, bool optional)
```
avanigupta, line 85: Unused argument

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -11,15 +12,18 @@ internal class AzureAppConfigurationSource : IConfigurationSource
         private readonly bool _optional;
         private readonly Func<AzureAppConfigurationOptions> _optionsProvider;
         private readonly IConfigurationClientFactory _configurationClientFactory;
+        private readonly string _environmentName;
+        private readonly bool _environmentNameEnabled;
```
avanigupta, line 16: Unused variable

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -50,7 +54,16 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
                     throw new ArgumentException($"Please call {nameof(AzureAppConfigurationOptions)}.{nameof(AzureAppConfigurationOptions.Connect)} to specify how to connect to Azure App Configuration.");
                 }
 
-                provider = new AzureAppConfigurationProvider(client, options, _optional);
+                try
```
avanigupta, line 57: This can be done in the beginning to avoid creating a ConfigurationClient that will never be used.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -50,7 +54,16 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
...
+                {
+                    string envType = Environment.GetEnvironmentVariable(RequestTracingConstants.AspNetCoreEnvironmentVariable) ??
+                                        Environment.GetEnvironmentVariable(RequestTracingConstants.DotNetCoreEnvironmentVariable);
+                    if (envType.Equals(_environmentName, StringComparison.OrdinalIgnoreCase))
```
avanigupta, line 61: .NET assumes that the environment is Production by default if these env var are not set. If we want to follow that pattern, we should still check if user specified environment name is Production and return a valid instance of AzureAppConfigurationProvider.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
amerjusupovic: I think we need to pass in a non-null client to the provider constructor, is there a way around that?

avanigupta: I was thinking of something like this:

```cs
public IConfigurationProvider Build(IConfigurationBuilder builder)
{
    try
    {
        string envType = Environment.GetEnvironmentVariable(RequestTracingConstants.AspNetCoreEnvironmentVariable) ??
                            Environment.GetEnvironmentVariable(RequestTracingConstants.DotNetCoreEnvironmentVariable);
        if (!envType.Equals(_environmentName, StringComparison.OrdinalIgnoreCase))
        {
            return new EmptyConfigurationProvider();
        }
    }
    catch (SecurityException) 
    {
        // cant read environment variables - assume prod
        if (!_environmentName.Equals(Environments.Production, StringComparison.OrdinalIgnoreCase))
        {
            return new EmptyConfigurationProvider();
        }
    }

    IConfigurationProvider provider = null;

    try
    {

        ... // no change needed
    }
    catch (InvalidOperationException e)
    {
        ... // no change needed
    }

    return provider ?? new EmptyConfigurationProvider();
}
```

amerjusupovic: Oh thanks, forgot about using empty provider

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: I was hoping we don't have to read from any environment variables directly but get the environment from something like `IHostEnvironment.EnvironmentName`, so we don't care which environment variable wins in which version of .NET.

Another scenario... if either of those *__ENVIRONMENT was set, they would not match "Production" for `_environmentName`, but shouldn't we consider this scenario as production?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -11,21 +13,44 @@ internal class AzureAppConfigurationSource : IConfigurationSource
         private readonly bool _optional;
         private readonly Func<AzureAppConfigurationOptions> _optionsProvider;
         private readonly IConfigurationClientFactory _configurationClientFactory;
+        private readonly string _environmentName;
 
-        public AzureAppConfigurationSource(Action<AzureAppConfigurationOptions> optionsInitializer, bool optional = false, IConfigurationClientFactory configurationClientFactory = null)
+        public AzureAppConfigurationSource(Action<AzureAppConfigurationOptions> optionsInitializer, string environmentName = null, bool optional = false, IConfigurationClientFactory configurationClientFactory = null)
```
zhenlan, line 18: >environmentName = null [](http://example.com/codeflow?start=107&length=22)

We are implicitly saying null means "any environment" and empty string is not. Not sure I like that subtle difference from a user perspective. Is this designed with some scenarios in mind?

avanigupta: That's a good point. We should block both null and empty string input for environment name.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -11,21 +13,44 @@ internal class AzureAppConfigurationSource : IConfigurationSource
...
+                    {
+                        return new EmptyConfigurationProvider();
+                    }
+                }
+                catch (SecurityException)
+                {
+                    // Can't read environment variables - assume production environment
+                    if (!_environmentName.Equals(Environments.Production, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return new EmptyConfigurationProvider();
```
zhenlan, line 49: This will be a silent failure. In a way, I think we should throw unless optional is true. But I really hope we can avoid reading environment variables altogether.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
zhenlan, line 18: NIT: I'd prefer we don't set default values for the parameters (environmentName and optional). It forces the caller to the set them explicitly. It's easier to read the code.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: I'm not sure if I understand your second point, I just know the default environment is production if neither are set, but is there another meaning if either one is set? I'm also looking to see if there's a way to get an instance of IHostEnvironment at that point in the code, but I don't see how to yet.

zhenlan: > the default environment is production if neither are set

Exactly. If neither is set, does our code match it as "production"?

amerjusupovic: Yes, the aim of this section was to check if the user specified "production" as the environment name when neither variable is set https://github.com/Azure/AppConfiguration-DotnetProvider/blob/ajusupovic/environmentname-api/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs#L46

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
zhenlan: I don't see the default value of `configurationClientFactory` is used anywhere.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
```diff
@@ -39,7 +61,7 @@ public static class AzureAppConfigurationExtensions
             Action<AzureAppConfigurationOptions> action,
             bool optional = false)
         {
-            return configurationBuilder.Add(new AzureAppConfigurationSource(action, optional));
+            return configurationBuilder.Add(new AzureAppConfigurationSource(action, optional: optional));
```
zhenlan, line 64: >optional: optional [](http://example.com/codeflow?start=84&length=18)

Are we adopting some new coding practice? "optional: optional" doesn't make code easier to read, IMO.

amerjusupovic: I only added this because it was would be an error otherwise when there were default values for environmentName and optional parameters. Removing both of those default values so I'll remove this too.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: The line you referenced is _not_ "neither is set". It's "either is not accessible".

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
zhenlan: Cool. We have one more reason to remove the default values :)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Ah I misunderstood thanks, I'll add that case

zhenlan: This is one of the reasons we should avoid reading from environment variables.

amerjusupovic: I don't think there is a way to access that `IHostEnvironment.EnvironmentName` value or any other value for the environment name given only the `IConfigurationBuilder`, which is all we have access to. All I can find is that the user could pass in the `WebHostBuilderContext` parameter for `ConfigureAppConfiguration` to `AddAzureAppConfiguration` as well and we could get the environment from there, similar to how it was described in the Background section in #312. This would mean passing in the `environmentName` parameter as well as the hosting context to `AddAzureAppConfiguration` for this new API. Does this sound like a better solution?

avanigupta: We dont have access to host configuration. Users will have to pass it to us if we want to avoid reading environment variables (which might not be a bad idea). In fact, it might be better to explicitly make users pass their current environment name instead of reading environment variables. 
We could modify the new API implemented in the PR to accept a second argument that corresponds to the current environment.
```cs
public static IConfigurationBuilder AddAzureAppConfiguration( 
            this IConfigurationBuilder configurationBuilder, 
            string enabledForEnvironment, 
            string currentEnvironment, 
            Action<AzureAppConfigurationOptions> action, 
            bool optional = false) 
```

Usage would look like this:

```cs
.ConfigureAppConfiguration((hostingContext, config) => 
{
    config.AddAzureAppConfiguration(
    enabledForEnvironment: "Production",
    currentEnvironment: hostingContext.HostingEnvironment.EnvironmentName,
    options => 
    { 
        options.Connect(prod_endpoint, prod_creds) 
               .Select(KeyFilter.Any, LabelFilter.Null); 
    }); 
} 
```

zhenlan: This comes back to the value of this API.
- This API is essentially a `if... else...` clause, so it's value is not about saving customer code. When a customer knows the hosting environment and what they want, it's not a problem to do the `if... else...` by themselves.
- This API's true value is relieving a customer from learning all those concepts. The API is environment-aware and does the right thing automatically.

Maybe it's okay to pass in `hostingContext`.

amerjusupovic: Did you mean passing in `hostingContext` instead of `hostingContext.HostingEnvironment.EnvironmentName`? I feel like that would make sense because we at least do some of the work of getting the environment name. Otherwise I'm also worried that doing an if/else for the user is not that much of an improvement.

zhenlan: Yes, something like that. But before we do that, it will be good to see what the code looks like for .NET/ASP.NET 6/7.

amerjusupovic:
```
public IConfigurationProvider Build(IConfigurationBuilder builder)
{
    IEnumerable<IConfigurationSource> sourceList = builder.Sources.Where(item => item is ChainedConfigurationSource);
    string currentEnvironment = null;
    if (sourceList != null && sourceList.Any())
    {
        ChainedConfigurationSource source = (ChainedConfigurationSource)sourceList.First();
        ConfigurationRoot root = (ConfigurationRoot)source.Configuration;
        if (root != null && root.Providers != null && root.Providers.Any())
        {
            ConfigurationProvider environmentVariables = (ConfigurationProvider)root.Providers.ElementAt(0);
            Console.WriteLine(environmentVariables.TryGet("ENVIRONMENT", out currentEnvironment));
        }
    }

    if (!string.IsNullOrEmpty(_environmentName))
    {
        if (string.IsNullOrEmpty(currentEnvironment))
        {
            if (!_environmentName.Equals("Production"))
            {
                return new EmptyConfigurationProvider();
            }
        }
        else
        {
            if (!currentEnvironment.Equals(_environmentName))
            {
                return new EmptyConfigurationProvider();
            }
        }
    }
```

Seems like this works to directly access the environment name from the `IConfigurationBuilder`, although it does look a bit ugly and I'm not sure if all of the ways to build an ASP.NET app would have a `ChainedConfigurationSource` yet. It did work for the ConfigStoreDemo in the examples folder from what I can tell.

amerjusupovic: Also it seems like for .NET and ASP.NET 6/7 the main thing I see is that environment variables are managed using [ConfigurationManager](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationmanager?view=dotnet-plat-ext-8.0) and are used [like this](https://github.com/dotnet/aspnetcore/blob/7264c87d64fce4420b382ddf9b834dbc97dd88d2/src/DefaultBuilder/src/WebApplicationBuilder.cs#L93) in `WebApplicationBuilder`. They just call `AddEnvironmentVariables` with a prefix and it overrides the "ENVIRONMENT" variable based on whatever is last, so "DOTNET" is prioritized in this case.

PR #383, opened 2023-02-13
Fix Intellisense for Register API

amerjusupovic:


zhenlan (approved):
Thanks, @amerjusupovic.

PR #384, opened 2023-02-14
Add new Azure Pipelines

amerjusupovic:
New pipelines created as a result of CDPX being deprecated.


.pipelines/windows-buddy-main.yml
```diff
@@ -0,0 +1,127 @@
...
+- none
+
+parameters: # parameters are shown up in ADO UI in a build queue time
+- name: 'debug'
+  displayName: 'Enable debug output'
+  type: boolean
+  default: false
+
+variables:
+  Codeql.Enabled: true
```
avanigupta, line 15: Do we need to enable codeql twice?

amerjusupovic: Left that in by mistake, will take this out and test again

.pipelines/windows-buddy-main.yml
```diff
@@ -0,0 +1,124 @@
...
+        Contents: '**/*.trx'
+        TargetFolder: '$(Build.SourcesDirectory)\out\outputs\test\_post_command__run_log_alerts_schedular_tests\_testresults'
+
+    - task: PublishTestResults@2
+      displayName: 'Unit Tests'
+      inputs:
+        testResultsFormat: 'vstest'
+        testResultsFiles: '**/*.trx'
+        searchFolder: ''
+        failTaskOnFailedTests: False
```
avanigupta, line 99: Can we fail the pipeline if any tests are failing?

avanigupta: Though we didnt do that in our previous cdpx pipelines. @jimmyca15 do you know if we did that on purpose?
I dont remember if there's a reason why we wouldn't want to fail the build pipeline if there are test failures.

avanigupta (approved):

.pipelines/windows-buddy-main.yml
jimmyca15: I can't imagine any reason we wouldn't want to require passing tests.

PR #394, opened 2023-03-21
Revert "Revert "Revert "Add an option to support Feature Management V2 schema…

amerjusupovic:


avanigupta (approved):
This should be completed with merge commit. cc @jimmyca15

Issue #317, opened 2022-04-15
A feature flag obtained via refresh may override a feature flag with higher precedence.


jimmyca15:
See [this discussion](https://github.com/Azure/AppConfiguration/issues/616#issuecomment-1062040088) for details of the original detection of this issue.

When using a setup like below to retrieve feature flags

```
.UseFeatureFlags(ffOptions =>
    { 
        ffOptions.Select(KeyFilter.Any, LabelFilter.Null);
        ffOptions.Select(KeyFilter.Any, "prod");
    });
```

It is expected that feature flags from the prod label override the flags with the null label. This works as expected as long as feature flags are not refreshed.

An issue can occur when feature flags are refreshed. When a change is detected for a feature flag with the null label, the new value will be used in the application. This happens even if there **was** a flag with the same name from the prod label that was in use.

## Expected

No matter how many times I refresh feature flags, and regardless of what has changed, a feature flag with a higher precedence label will override a similarly named flag with the a lower precedence label.

## Actual

If a feature flag with a low precedence label is changed and a refresh is triggered, that feature flag will override any feature flags with the same name even if they have a higher precedence.

xelossan:
Are there any updates on this? We're having the same issue - sometimes flags with label are overwritten by values with null label, which is really bad.

amerjusupovic:
Hi @xelossan, we're working on getting a fix out with the linked pull request for the next major version stable release. I can't give you an exact timeline on this at the moment but I can add an update here if anything changes.

danyltsiv:
I can't reproduce the issue mentioned in this ticket, however, we have the same weird thing going on with feature flags. At some random moment, the value(false) from non-labeled overrides labeled ones(true). The weirdest part is that it randomly appears and randomly disappears, hard to find something that correlates with the issue to figure out the root cause. In our case, feature flags are manually updated, no automatism there, and there is no correlation between update times and issue times.

muzopraha:
We have same issue on version 7.0.0

knopa:
Any updates?

amerjusupovic:
The PR #536 made changes to the feature flag refresh behavior, including monitoring feature flags as a collection instead of on an individual basis. Those changes fixed this existing bug.

PR #404, opened 2023-04-17
Fix feature flag overriding higher precedence feature flags bug

amerjusupovic:
This PR fixes a bug mentioned in #317 about selecting feature flags with the same key but different labels. The precedence of the label is based on the order of calls to ```FeatureFlagOptions.Select```, with later calls having higher precedence. With these changes, a feature flag with a higher precedence label will always override a feature flag with a lower precedence label that has the same key. If the highest precedence feature flag is deleted, the configuration will update with the next highest precedence feature flag for that key if one exists.  This change will also apply to refresh-registered key-values, where later calls to ```AzureAppConfigurationRefreshOptions.Register``` will have higher precedence. 

**Example feature flag setup**

```
.UseFeatureFlags(ffOptions =>
    { 
        ffOptions.Select(KeyFilter.Any, LabelFilter.Null);
        ffOptions.Select(KeyFilter.Any, "prod");
    });
```

In this scenario, if there are two feature flags that have the same key, the feature flag with the "prod" label will always override the feature flag with the null label.

**Example refresh-registered key-value setup**

```
.ConfigureRefresh(refreshOptions =>
{
    refreshOptions.Register("TestKeyWithMultipleLabels", "label1")
                  .Register("TestKeyWithMultipleLabels", "label2");
});
```

In this scenario, the key-value with key "TestKeyWithMultipleLabels" and label "label2" will always override the key-value with key "TestKeyWithMultipleLabels" and label "label1".


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -379,6 +380,16 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                     {
                         _watchedSettings = watchedSettings;
 
+                        Dictionary<string, ConfigurationSetting> precedenceData = OrderDataByPrecedence(out Dictionary<string, string> logIdentifiers);
```
avanigupta, line 383: Why dont we directly pass `logInfoBuilder` here? Is it because we want to rearrange the logging order depending on the precedence order?

avanigupta: Its generally not a good practice to have a method that returns a value and also has out parameters.

avanigupta: This is only reordering logs for feature flags - if we are reordering, I think we should do it for key-values too.

Alternatively, instead of rearranging logs, we log everything as it happens but while preparing data, we add an additional log statement for all overwrites of changed settings. I'm thinking of something like this:

During RefreshAsync:
`Key-value read from App Configuration. Change:'Modified' Key:'FontColor' Label:'Dev' Endpoint:'contoso.azconfig.io'`
`Setting updated. Key: FontColor`

During PrepareData:
`Value of Key:'FontColor' Label:'Prod' will override previous values.`

I think we dont need the new helper method for ordering precedence data, it can be part of `PrepareData`, and `PrepareData` can be updated to accept a list of key value changes:

```
private async Task<Dictionary<string, string>> PrepareData(IEnumerable<KeyValueChange> changes, CancellationToken cancellationToken = default)
```

Notice that we dont really need to pass a dictionary of data to PrepareData because its always operating on `_mappedData`.

amerjusupovic: Yea, I see now that I missed reordering the info level logs for key-values.

I mostly understand and I'm willing to change it, but I think this log feels like it could be confusing because of the order. Did you suggest it because the log rearranging adds some iteration that we don't necessarily need?

avanigupta: I'm trying to think of a way to do it without having to pass around a new placeholder data structure - logInfoBuilder is itself a placeholder, now we're passing around a placeholder for a placeholder :)
It also feels odd to reorder info logs, but not debug logs. So either we do both or neither.
If my previous suggestion feels confusing, how about getting rid of the logInfoBuilder and logDebugBuilder string builders and just pass around `cachedInfoLogs` and `cachedDebugLogs`?

amerjusupovic: Makes sense, my original thought was that the debug messages didn't really need to be reordered since they just mention reading key-values so I didn't want to add `cachedDebugLogs`. I agree that it looks odd to use a different variable for each log level, but I also think we should only add `cachedDebugLogs` if we want to reorder the debug logs, which I don't see a reason for right now. I know that didn't help much, but that's just how I see it right now.

avanigupta: Do you mind giving a sample of what the debug and info logs will look like with the changes you currently have?

amerjusupovic: Code to select feature flags: 
```
 options.UseFeatureFlags(ffoptions =>
 {
     ffoptions.Select(KeyFilter.Any, null);
     ffoptions.Select(KeyFilter.Any, "prod");
 });
```

Changing flag `alpha` with label `null`:
```
dbug: Microsoft.Extensions.Configuration.AzureAppConfiguration.Refresh[0]
      Feature flag read from App Configuration. Key:'alpha' Label:'' Endpoint:'https://ajusupovic-ac2.azconfig.io'
```

Changing flag `alpha` with label `prod`:
```
dbug: Microsoft.Extensions.Configuration.AzureAppConfiguration.Refresh[0]
      Feature flag read from App Configuration. Key:'alpha' Label:'prod' Endpoint:'https://ajusupovic-ac2.azconfig.io'
info: Microsoft.Extensions.Configuration.AzureAppConfiguration.Refresh[0]
      Feature flag updated. Key:'alpha'
```

avanigupta: Thanks for the example. The current code behaves differently though. Your first scenario will produce logs like the second scenario. Changing flag `alpha` with label `null` will still have an info log saying `Feature flag updated. Key:'alpha'`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -587,10 +618,10 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
 
                 try
                 {
-                    Dictionary<string, ConfigurationSetting> mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
-                    SetData(await PrepareData(mappedData, cancellationToken).ConfigureAwait(false));
+                    Dictionary<KeyValueIdentifier, ConfigurationSetting> mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
                     _watchedSettings = watchedSettings;
                     _mappedData = mappedData;
+                    SetData(await PrepareData(null, cancellationToken).ConfigureAwait(false));
```
amerjusupovic, line 624: This looks wrong to me but not sure what the best way to do this is, since if I split up `PrepareData` I can't find how to do it without repeating code or needing to have an out param in a helper. I could make an overload that calls the main `PrepareData` instead?

avanigupta: One option is to create a helper method - and that method directly calls PrepareData. Something like this:

```cs
Task<Dictionary<string, string>> ReorderAndPrepareData(Dictionary<KeyValueIdentifier, string> cachedInfoLogs, CancellationToken cancellationToken = default)
{
    // Reorder data

    // Call PrepareData
}

Task<Dictionary<string, string>> PrepareData(Dictionary<string, ConfigurationSetting> orderedData, CancellationToken cancellationToken = default)
{
    // No change needed
}
```

Then `InitializeAsync` can call `PrepareData` after converting `_mappedData` to a dictionary with string keys and `RefreshAsync` can call `ReorderAndPrepareData`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -587,10 +634,17 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
 
                 try
                 {
-                    Dictionary<string, ConfigurationSetting> mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
-                    SetData(await PrepareData(mappedData, cancellationToken).ConfigureAwait(false));
+                    Dictionary<KeyValueIdentifier, ConfigurationSetting> mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
```
avanigupta, line 637: I dont think we need `mappedData` here. We can directly assign to `_mappedData`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -537,9 +542,51 @@ private void SetDirty(TimeSpan? maxDelay)
             return applicationData;
         }
 
+        private async Task<Dictionary<string, string>> ReorderAndPrepareData(Dictionary<KeyValueIdentifier, string> cachedInfoLogs, CancellationToken cancellationToken = default)
+        {
+            Dictionary<string, ConfigurationSetting> precedenceData = new Dictionary<string, ConfigurationSetting>();
```
avanigupta, line 547: nit: can we call it something like `orderedData`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -277,13 +276,15 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                                    cachedInfoLogs[changeIdentifier] = LogHelper.BuildKeyValueSettingUpdatedMessage(change.Key);
                                     keyValueChanges.Add(change);
 
                                     if (changeWatcher.RefreshAll)
                                     {
                                         refreshAll = true;
+                                        cachedInfoLogs[changeIdentifier] += $"{Environment.NewLine}{LogHelper.BuildConfigurationUpdatedMessage()}";
                                         break;
                                     }
                                 } else
```
avanigupta, line 290: nit: can you move `else` to a new line?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -230,7 +230,6 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                             refreshAll = false;
                             Uri endpoint = _configClientManager.GetEndpointForClient(client);
                             logDebugBuilder.Clear();
-                            logInfoBuilder.Clear();
```
avanigupta, line 233: do we need to clear cachedInfoLogs here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -512,6 +507,16 @@ private void SetDirty(TimeSpan? maxDelay)
...
+            foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers.Concat(_options.MultiKeyWatchers))
+            {
+                IEnumerable<ConfigurationSetting> currentKeyValues = GetCurrentKeyValueCollection(changeWatcher.Key, changeWatcher.Label, _watchedSettings.Values);
+
+                foreach (ConfigurationSetting setting in currentKeyValues)
+                {
+                    data[setting.Key] = _mappedData[new KeyValueIdentifier(setting.Key, setting.Label)];
+                }
+            }
+
```
avanigupta, line 519: This is not needed in PrepareData

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -537,9 +542,51 @@ private void SetDirty(TimeSpan? maxDelay)
...
         }
 
+        private async Task<Dictionary<string, string>> ReorderAndPrepareData(Dictionary<KeyValueIdentifier, string> cachedInfoLogs, CancellationToken cancellationToken = default)
+        {
+            Dictionary<string, ConfigurationSetting> precedenceData = new Dictionary<string, ConfigurationSetting>();
+            Dictionary<string, string> logIdentifiers = new Dictionary<string, string>();
+
+            foreach (KeyValuePair<KeyValueIdentifier, ConfigurationSetting> kvp in _mappedData)
+            {
+                precedenceData[kvp.Key.Key] = kvp.Value;
```
avanigupta, line 552: Do we need to keep track of logIdentifiers here too?

amerjusupovic: Shouldn't be necessary I think, the `logIdentifiers` object was just to keep track of which logs to actually print out, but we only log for values that are tracked by the change watchers.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -587,10 +626,26 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
...
+
+                    foreach (KeyValueWatcher changeWatcher in _options.ChangeWatchers.Concat(_options.MultiKeyWatchers))
+                    {
+                        IEnumerable<ConfigurationSetting> currentKeyValues = GetCurrentKeyValueCollection(changeWatcher.Key, changeWatcher.Label, _watchedSettings.Values);
+
+                        foreach (ConfigurationSetting setting in currentKeyValues)
+                        {
+                            orderedData[setting.Key] = _mappedData[new KeyValueIdentifier(setting.Key, setting.Label)];
+                        }
+                    }
```
avanigupta, line 646: During initialization, we don't need to reorder watched settings because we just finished reloading the entire configuration. `_mappedData` should already be in the right order of precedence.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -548,9 +545,51 @@ private void SetDirty(TimeSpan? maxDelay)
             return applicationData;
         }
 
+        private Dictionary<string, ConfigurationSetting> ReorderData(Dictionary<KeyValueIdentifier, string> cachedInfoLogs = default)
```
jimmyca15, line 548: `cachedInfoLogs` I'm not seeing where logging is coming into the picture here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -548,9 +545,51 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                {
+                    if (cachedInfoLogs.TryGetValue(new KeyValueIdentifier(kvp.Key, kvp.Value), out string log))
+                    {
+                        logInfoBuilder.AppendLine(log);
+                    }
+                }
+
+                if (logInfoBuilder.Length > 0)
+                {
+                    _logger.LogInformation(logInfoBuilder.ToString().Trim());
```
jimmyca15, line 583: What is this log? Is this for debugging? I don't see any mention of logging in the PR description.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -548,9 +545,51 @@ private void SetDirty(TimeSpan? maxDelay)
             return applicationData;
         }
 
+        private Dictionary<string, ConfigurationSetting> ReorderData(Dictionary<KeyValueIdentifier, string> cachedInfoLogs = default)
+        {
+            Dictionary<string, ConfigurationSetting> orderedData = new Dictionary<string, ConfigurationSetting>();
+            Dictionary<string, string> logIdentifiers = new Dictionary<string, string>();
+
+            foreach (KeyValuePair<KeyValueIdentifier, ConfigurationSetting> kvp in _mappedData)
```
jimmyca15, line 553: Ideally mapped data should come as a parameter.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -548,9 +545,51 @@ private void SetDirty(TimeSpan? maxDelay)
             return applicationData;
         }
 
+        private Dictionary<string, ConfigurationSetting> ReorderData(Dictionary<KeyValueIdentifier, string> cachedInfoLogs = default)
+        {
+            Dictionary<string, ConfigurationSetting> orderedData = new Dictionary<string, ConfigurationSetting>();
+            Dictionary<string, string> logIdentifiers = new Dictionary<string, string>();
+
```
jimmyca15, line 552: Doesn't this break simple label override scenario?

```
options.Connect()
    .Select(KeyFilter.All, LabelFilter.Null)
    .Select(KeyFilter.All, "Prod")
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: It replaces the log that was removed on line 387. I can see how it looks confusing to split these up, but I did it to prevent some issues with passing around params. I could try to move it or add comments to clarify

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Related to other comment about removed log from line 387.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Since I used a `Dictionary<TKey, TValue>` this scenario could break when the values in the dictionary change, so I could use an `OrderedDictionary` instead. However, `OrderedDictionary` doesn't let you specify types like `<TKey, TValue>` when creating one, which is a drawback. I'm still trying to find a better way to solve this.

amerjusupovic: https://github.com/dotnet/runtime/issues/24826

VeronikaKorzhevyh:
Hi @amerjusupovic, do you have any information on when this PR will be ready to merge and when the next release will be? Thanks!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -293,14 +298,13 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                             if (refreshAll)
                             {
                                 // Trigger a single load-all operation if a change was detected in one or more key-values with refreshAll: true
-                                data = await LoadSelectedKeyValues(client, cancellationToken).ConfigureAwait(false);
+                                (data, orderedSelectedKeyValues) = await LoadSelectedKeyValues(client, cancellationToken).ConfigureAwait(false);
```
avanigupta, line 301: We're doing this because there's no generic OrderedDictionary in .NET, however, I still feel that maintaining a new data structure to track the insertion order might be an overkill. It also adds the extra complication of having to return 2 objects from `LoadSelectedKeyValues`. 

Is it possible to convert `_mappedData` to `List<ConfigurationSetting>` instead of a dictionary? That would preserve the insertion order and we can still perform lookups because `ConfigurationSetting` already contains `key` and `label`.

amerjusupovic: That's definitely fair, I considered making `_mappedData` a list but I was just concerned about having to iterate over the entire list for accessing/removing a specific key upon refresh. Then again, I don't really know how much of an impact that has.

muzopraha:
Is there any update on this issue? Our significant government project relies on this feature. However, it is currently a major obstacle for us. We have a couple of weeks to make a decision. Thank you for your attention to this matter.

knopa:
Any updates?

amerjusupovic:
> Any updates?

This PR will likely be closed, as I've updated the linked issue with a pull request (#574) which will solve this bug in addition to adding a new API for refresh. The other PR still needs to be reviewed and merged, so it will take some more time before it's released. I'm sorry for the inconvenience.

knopa:
Thanks for quick reply. I will keep an an on that one. Seems it 's good idea work with them as collection

PR #408, opened 2023-04-26
Throw input errors on build optional sources

### LINKED ISSUE #318: Inconsistent exception handling when `AddAzureAppConfiguration` is invoked with `optional=true`
> My preference would be to fail no matter what for errors that come from invalid inputs, but that seems to contradict the documentation for the enabled flag.

My thoughts too. I think our "optional" behavior comes from the "optional" behavior of FileConfigurationProviders. For configuration files, this flag suppresses the exception when a file is not found. Anything else, like invalid syntax in file or empty file path still throws an exception when optional is true. 
Along similar lines, if optional is true for AppConfig provider, any errors that occur during `Build()` should not be suppressed (these are always due to invalid inputs). However, any errors that occur during `Load()` should be suppressed.

cc @zhenlan

_Originally posted by @avanigupta in https://github.com/Azure/AppConfiguration-DotnetProvider/pull/316#discussion_r857996210_

Eskibear:
This PR follows https://github.com/Azure/AppConfiguration-DotnetProvider/issues/318#issuecomment-1522874825 to fix #318 , it:
- removes check for `optional` flag, always throw user input error like malformed connection string, even if for optional configuration sources.
- adds some test cases guarding the behavior


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -51,17 +51,11 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
...
-                }
+                throw new ArgumentException(ex.Message, ex);
             }
             catch (FormatException fe) // FormatException is thrown when the connection string is not a well formed connection string.
             {
-                if (!_optional)
-                {
-                    throw new ArgumentException(fe.Message, fe);
-                }
+                throw new ArgumentException(fe.Message, fe);
```
avanigupta, line 58: Could you also update the summary of optional flag in `AzureAppConfigurationExtensions`?
Maybe we can change the first sentence to something like - "Determines the behavior of the App Configuration provider when an exception occurs while loading data from server."

Eskibear: Good catch!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
```diff
@@ -1,13 +1,13 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 //
-using Microsoft.Extensions.Configuration.AzureAppConfiguration;
-using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
```
avanigupta, line 4: Looks like there are some unintentional whitespace changes in this file. Could you reset the whitespaces and update only the method summary?

Eskibear: I intended to do that. Because I see inconsistency in line endings, some lines are LF and some CRLF. Do you suggest that I don't touch them in this PR?

part of `git diff -R` results as below:

```diff
diff --git b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
index 3dabf4c..657512e 100644
--- b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
+++ a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
@@ -1,13 +1,13 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 //
-using Microsoft.Extensions.Configuration.AzureAppConfiguration;
-using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;^M
+using Microsoft.Extensions.DependencyInjection;^M
 using System;
-using System.Collections.Generic;
-
+using System.Collections.Generic;^M
+^M
 namespace Microsoft.Extensions.Configuration
-{
+{^M
```

Eskibear: I prefer to make it consistent, and also we won't see the warning in VS as below:
image

BTW I glanced at other files, their EOL are all CRLF. Only `AzureAppConfigurationExtensions.cs` has mixed EOLs. So probably I should update it to CRLF instead of LF. Never mind, it's a linting issue and I can create a separate PR to do that if necessary.

I'll follow your suggestion don't touch it here. WDYT?

avanigupta: I think it's a good idea to update it in a different PR. This looks good now. Thanks!

avanigupta:
Changes look good to me! However, we'll hold off on merging this PR because it's a breaking change. We can revisit it when we are ready for the next major release. Thanks @Eskibear !

amerjusupovic (approved):

PR #410, opened 2023-05-01
Enforce calling ConfigureAwait() when awaiting a Task

amerjusupovic:
A bug caused by a missing call to ConfigureAwait() caused applications running on .NET Framework to fail when trying to refresh the configuration. These missing calls have been added, along with a rule to enforce that ConfigureAwait() is called whenever we await a task.

### DIFF
--- a/src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
+++ b/src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
@@ -38,7 +38,7 @@
   <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'!='Official'">
     <Version>$(OfficialVersion)-$(CDP_PATCH_NUMBER)-$(Revision)</Version>
   </PropertyGroup>
-
+  
   <PropertyGroup>
     <CodeAnalysisRuleSet>..\..\AzureAppConfigurationRules.ruleset</CodeAnalysisRuleSet>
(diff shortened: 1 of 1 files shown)


zhenlan (approved):

AzureAppConfigurationRules.ruleset
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Default" ToolsVersion="17.0">
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.NetAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.NetAnalyzers">
+    <Rule Id="CA2007" Action="Error" />
```
jimmyca15, line 4: Can you add a comment describing the purpose of the rule.

src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
```diff
@@ -38,5 +38,10 @@
   <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'!='Official'">
     <Version>$(OfficialVersion)-$(CDP_PATCH_NUMBER)-$(Revision)</Version>
   </PropertyGroup>
-  
+
+  <PropertyGroup>
+	<CodeAnalysisRuleSet>..\..\AzureAppConfigurationRules.ruleset</CodeAnalysisRuleSet>
```
jimmyca15, line 43: The indenting looks a bit consistent in this file.

avanigupta (approved):

jimmyca15 (approved):

PR #411, opened 2023-05-02
Upgrade package versions to 6.0.1

amerjusupovic:
This is part of a release for the fix in #410


zhenlan (approved):
Thanks, Amer. Can we test it with the repro app again once you have the official build?

Issue #406, opened 2023-04-21
Support RequirementType property for feature flags


avanigupta:
With Microsoft.FeatureManagement 2.6.0-preview release, feature flags now have an additional property called RequirementType that needs to be parsed by the FeatureManagement adapter. See this PR for the new syntax: https://github.com/microsoft/FeatureManagement-Dotnet/pull/221

Joerio:
This is great and works very well. I do however have some issue with this new property in combination with Azure App Configuration. Is this already supported, and if so, what property do I need to add in that configuration?

I only get it to work at this moment if I add the feature also in mine appsettings.json with the property RequirementType, I can leave the rest of the configuration (filters) in Azure. It would be great if I didn't have to put anything at all in my local appsettings.json.

avanigupta:
Hi @Joerio, thanks for the feedback! Right now, it will only work when you specify RequirementType in your config file. We are working on adding this support in Azure App Configuration too.

Eskibear:
I have some questions while trying to implement the feature. Probably some clarification required.
- Do we always set property `FeatureManagement:<FeatureId>:RequirementType`, even if `requirement_type` is not specified in raw data? If so, I think we can use its default value "Any".
- What if `requirement_type` has an invalid value? Do we throw an exception or silently fallback to set "Any" (which is not a good idea from my gut feeling)?

avanigupta:
Config provider is responsible for parsing feature flags stored in AppConfig service and adding them to application configuration. Then the Feature Management library consumes these feature flags through `IConfiguration`. This means that config provider doesn't need to add default value for RequirementType - it just needs to understand how to parse a feature flag if it contains RequirementType. Even if the value of RequirementType is invalid, config provider shouldn't worry about the semantics of RequirementType and just treat it as a free form string. The FeatureManagement library will handle these cases of missing/invalid RequirementType appropriately.

Eskibear:
Thank you for the clarification, and it makes a lot sense to me now. Then I think I can simply regard requirement type as a string here in the provider.

PR #418, opened 2023-05-04
parse requirement type of feature flags

Eskibear:
See #406


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureConditions.cs
```diff
@@ -10,5 +10,10 @@ internal class FeatureConditions
     {
         [JsonPropertyName("client_filters")]
         public List<ClientFilter> ClientFilters { get; set; } = new List<ClientFilter>();
+
+        //
+        // TODO: Need a JSON Converter for enum type
```
Eskibear, line 15: Here I create a new type `RequirementType` as an enum, but looks like there's no built-in JSON converter for enum type.

I find related issues:
- https://github.com/dotnet/runtime/issues/31081
- https://github.com/dotnet/runtime/issues/29975

Not sure whether I should implement a Json converter for enum on my own, or introduce third party packages?

Eskibear: Or to use [JsonStringEnumConverter](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonstringenumconverter?view=net-6.0)?

Eskibear: > Or to use [JsonStringEnumConverter](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonstringenumconverter?view=net-6.0)?

It seems to work!
I'm not yet familiar with .NET, let me know if there's any consideration about introducing it.

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -1129,5 +1144,66 @@ Response<ConfigurationSetting> GetTestKey(string key, string label, Cancellation
...
+            var config = new ConfigurationBuilder()
+                .AddAzureAppConfiguration(options =>
+                {
+                    options.ClientManager = TestHelpers.CreateMockedConfigurationClientManager(mockClient.Object);
+                    options.UseFeatureFlags();
+                })
+                .Build();
+
+            Assert.Null(config["FeatureManagement:MyFeature2:RequirementType"]);
+            Assert.Null(config["FeatureManagement:Feature_NoFilters:RequirementType"]);
```
Eskibear, line 1186: #### TL;DR
I have a question that how should we parse from as below, requirement_type is specified but client_filters is empty:
```ts
conditions: {
  requirement_type: all/any,
  client_filters: []
}
```
Option A) `FeatureManagement:FeatureId = True` or Option B) `FeatureManagement:FeatureId:RequirementType=All/Any`?

#### Thoughts
In previous impl, if `client_filters` is empty, property `FeatureManagement:FeatureId` is set to `True` and skip further processing, because `client_filters` is the only member of `FeatureConditions`.
Two ways to go:
- Now `RequirementType` doesn't make any sense if there's no filter at all. Pick Option A.
- We do not consider any semantic/logic stuffs, simply do parsing work. Pick Option B.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -51,6 +51,13 @@ internal class FeatureManagementKeyValueAdapter : IKeyValueAdapter
                             keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.SectionName}:{featureFlag.Id}:{FeatureManagementConstants.EnabledFor}:{i}:Parameters:{kvp.Key}", kvp.Value));
                         }
                     }
+
+                    //
+                    // process RequirementType only when filters are not empty
+                    if (featureFlag.Conditions.RequirementType != null)
+                    {
+                        keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.SectionName}:{featureFlag.Id}:{FeatureManagementConstants.RequirementType}", featureFlag.Conditions.RequirementType));
```
rossgrambo, line 59:
```suggestion
                        keyValues.Add(new KeyValuePair<string, string>(
                            $"{FeatureManagementConstants.SectionName}:{featureFlag.Id}:{FeatureManagementConstants.RequirementType}", 
                            featureFlag.Conditions.RequirementType));
```

rossgrambo: nit: I prefer this spacing to make it easier to read ^. Might have to also do it to the other large lines above.

rossgrambo (approved):

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
avanigupta: On AppConfig side, we look at the `enabled` property of a feature flag to decide whether a feature is enabled, disabled or conditional. When the provider parses a feature flag with `enabled=true` and non-empty `client_filters` list, it means that the feature is conditional, i.e., we need to parse all the filters and `requirement_type`. But if `enabled=true` but `client_filters` list is empty, it means that the feature is always enabled.

However, on feature management side, features without any filters are considered disabled. So when a feature is supposed to be enabled (i.e., `FeatureManagement:FeatureId = True`), we automatically inject a special filter called "AlwaysOn" to enable the feature ([code](https://github.com/microsoft/FeatureManagement-Dotnet/blob/5c3e941afc0d1e88a57253364fbce61b2138394f/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L144-L156)).

So from your example, we cannot go for Option B without making changes in the Feature Management library to handle this scenario. Option A should be fine to implement here.
Thoughts @jimmyca15 @rossgrambo ?

jimmyca15: Option a is my vote

Eskibear: It makes sense to me now. This PR already follows Option A, and I'll keep it as is.

Thank you all for elaborated explanations.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
Eskibear: Thanks, committed as suggested.

> Might have to also do it to the other large lines above.

For other lines unrelated to this feature, I would prefer to create a separate PR for such style/format changes, against main branch.

avanigupta (approved):

PR #419, opened 2023-05-04
Allow users to select only feature flags

amerjusupovic:
This allows users to select only feature flags or key vault values, but it's a breaking change since it removes the default select query.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -597,34 +597,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         {
             var serverData = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);
 
-            // Use default query if there are no key-values specified for use other than the feature flags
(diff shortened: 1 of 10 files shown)


zhenlan:
I don't think we want to change the default behavior. The question is more about whether we want to offer users an option to load feature flags only. There is a simple workaround like below. Not sure if we need it.

```cs
    configurationBuilder.AddAzureAppConfiguration(options =>
    {
        options. Connect(Environment.GetEnvironmentVariable("ConnectionString"))
               .Select("_")
               .UseFeatureFlags();
    });
```

amerjusupovic:
> I don't think we want to change the default behavior. The question is more about whether we want to offer users an option to load feature flags only. There is a simple workaround like below. Not sure if we need it.
> 
> ```cs
>     configurationBuilder.AddAzureAppConfiguration(options =>
>     {
>         options. Connect(Environment.GetEnvironmentVariable("ConnectionString"))
>                .Select("_")
>                .UseFeatureFlags();
>     });
> ```

I see, it is a bit much to change the default behavior. So do you think this is clear enough to where we don't need to add a way to explicitly disable it, like a new method/overload?

zhenlan:
If we do, I'm thinking something like `UseFeatureFlags(ffOptions, loadFeatureFlagsOnly: true)` but I vote for doing nothing as there is a simple workaround, and few people (if anyone) asked about it.

PR #422, opened 2023-05-12
Add SelectSnapshot to support loading configuration from snapshots

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -153,7 +153,33 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
...
+                    LabelFilter = labelFilter,
+                    SnapshotName = string.Empty
+                });
+            }
+
+            return this;
+        }
+
+        /// <summary>
+        /// Specify a snapshot and include its captured key-values in the configuration provider.
```
jimmyca15, line 165: Instead of 'captured' can you say 'contained'

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,21 +618,34 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
                 }).ConfigureAwait(false);
             }
 
-                foreach (var loadOption in _options.KeyValueSelectors)
+            foreach (var loadOption in _options.KeyValueSelectors)
```
jimmyca15, line 621: Can you replace the `var` here with a concrete type?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,21 +618,34 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
+                    await CallWithRequestTracing(async () =>
                     {
-                        serverData[setting.Key] = setting;
-                    }
-                }).ConfigureAwait(false);
+                        await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsForSnapshotAsync(loadOption.SnapshotName, cancellationToken).ConfigureAwait(false))
+                        {
+                            serverData[setting.Key] = setting;
+                        }
+                    }).ConfigureAwait(false);
```
jimmyca15, line 647: How about writing like this to save a bit of duplication ?

```
            foreach (var loadOption in _options.KeyValueSelectors)
            {
                IAsyncEnumerable<ConfigurationSetting> settingsEnumerable;

                if (string.IsNullOrEmpty(loadOption.SnapshotName))
                {
                    settingsEnumerable = client.GetConfigurationSettingsAsync(
                        new SettingSelector
                        {
                            KeyFilter = loadOption.KeyFilter,
                        },
                        cancellationToken);
                }
                else
                {
                    settingsEnumerable = client.GetConfigurationSettingsForSnapshotAsync(
                        loadOption.SnapshotName,
                        cancellationToken);
                }

                await CallWithRequestTracing(async () =>
                {
                    await foreach (ConfigurationSetting setting in settingsEnumerable.ConfigureAwait(false))
                    {
                        serverData[setting.Key] = setting;
                    }
                })
                .ConfigureAwait(false);
            }
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -153,7 +153,33 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
...
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if (!_kvSelectors.Any(s => s.SnapshotName.Equals(name)))
+            {
+                _kvSelectors.Add(new KeyValueSelector
+                {
+                    KeyFilter = KeyFilter.Any,
```
jimmyca15, line 180: This doesn't need to be set does it? Same with label filter.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -153,7 +153,33 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
                 _kvSelectors.Add(new KeyValueSelector
                 {
                     KeyFilter = keyFilter,
-                    LabelFilter = labelFilter
+                    LabelFilter = labelFilter,
+                    SnapshotName = string.Empty
```
jimmyca15, line 157: Why add it?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: I was setting these to avoid a null reference error in `LoadSelectedKeyValues` when we iterate through the key-value selectors. The other option was to check for null in that method instead, but I wasn't sure if it would be better to just prevent errors by setting the selector filters or just always check for null when referencing them. I think I'll just add the null checks because it makes more sense conceptually to have specific filters being set.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: Same deal as above comment

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -153,7 +153,33 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
...
+
+            return this;
+        }
+
+        /// <summary>
+        /// Specify a snapshot and include its captured key-values in the configuration provider.
+        /// <see cref="SelectSnapshot"/> can be called multiple times to include key-values from multiple snapshots.
+        /// </summary>
+        /// <param name="name">The name of the snapshot in Azure App Configuration.</param>
+        public AzureAppConfigurationOptions SelectSnapshot(string name)
```
zhenlan, line 169: Please add a check for the composition_type.key before loading a snapshot; throw otherwise.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,21 +618,45 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
                 {
-                    var selector = new SettingSelector
+                    ConfigurationSettingsSnapshot snapshot = await client.GetSnapshotAsync(loadOption.SnapshotName);
+
+                    if (snapshot != null)
                     {
-                        KeyFilter = loadOption.KeyFilter,
-                        LabelFilter = loadOption.LabelFilter
-                    };
+                        if (snapshot.CompositionType == null || !snapshot.CompositionType.ToString().Equals("key"))
```
jimmyca15, line 641: how can snapshot composition type be numb. Isn't it an enum?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,21 +618,45 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
+                        {
+                            throw new ArgumentException($"{nameof(snapshot.CompositionType)} for the selected snapshot must be 'key', found '{snapshot.CompositionType}'.");
```
jimmyca15, line 643: It's not an argument exception. No passed argument is invalid. It's more of an invalid operation exception.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Okay, it's nullable due to it being used for create flow.

Is the null check necessary though? I'd expect `if (snapshot.CompositionType != CompositionType.Key)` to suffice.

Also, this if check can be combined with the enclosing if statement.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,21 +618,45 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
+                            LabelFilter = loadOption.LabelFilter
+                        },
+                        cancellationToken);
+                }
+                else
                 {
-                    var selector = new SettingSelector
+                    ConfigurationSettingsSnapshot snapshot = await client.GetSnapshotAsync(loadOption.SnapshotName);
+
+                    if (snapshot != null)
```
jimmyca15, line 639: Perhaps we should require snapshot to exist as well.

@zhenlan

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Makes sense, the null check was just me being extra cautious.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: So the SDK doesn't throw if a snapshot doesn't exist but throws if a key-value doesn't exist?

zhenlan: Anyways, yes, it makes sense to throw if a snapshot doesn't exist.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,21 +618,45 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
-                        KeyFilter = loadOption.KeyFilter,
-                        LabelFilter = loadOption.LabelFilter
-                    };
+                        if (snapshot.CompositionType == null || !snapshot.CompositionType.ToString().Equals("key"))
+                        {
+                            throw new ArgumentException($"{nameof(snapshot.CompositionType)} for the selected snapshot must be 'key', found '{snapshot.CompositionType}'.");
+                        }
+                    }
+
+                    settingsEnumerable = client.GetConfigurationSettingsForSnapshotAsync(
```
zhenlan, line 647: What are our thoughts about caching the snapshot results so we don't reload it during refresh?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: I expect consistent behavior between key-value and snapshot retrieval. @amerjusupovic can you validate?

amerjusupovic: Yes, I'll test this to confirm the expected behavior

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: The only thing that comes to mind is we'd need to be able to tell if a snapshot was deleted and one with the same name was created with different key-values. Maybe we could see that it's a different snapshot based on the time it was created?

jimmyca15: I'm a little torn on the idea. 

On one hand snapshots are a perfect candidate for caching because they are mostly static.

However, an app setup with refresh is already establishing a dependency on querying key-values during runtime, so it waters down the value of caching snapshots for me a little bit. 

The downside of the caching would be added memory pressure of caching the snapshot response kvs. The previous comment about refresh deters me from suggesting adding more things in memory.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -598,7 +598,7 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
             var serverData = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);
 
             // Use default query if there are no key-values specified for use other than the feature flags
-            bool useDefaultQuery = !_options.KeyValueSelectors.Any(selector => !selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker));
+            bool useDefaultQuery = !_options.KeyValueSelectors.Any(selector => (!string.IsNullOrEmpty(selector.KeyFilter) && !selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker)) || !string.IsNullOrEmpty(selector.SnapshotName));
```
zhenlan, line 601: My brain hurts when reading this line :)

I think we shouldn't use the default query (load all) when a snapshot is specified.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: When refresh is enabled, we will need the data of snapshots anyway, so I won't worry about the memory if we cache them. It doesn't make sense to reload the snapshots every few seconds. We can skip caching if the refresh is not enabled.

It's an interesting scenario Amer brought up. The snapshot is not expected to change, but it could change in the scenario you mentioned. I would rather cache it, so the configuration won't change due to a refresh.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Oh my fault I can simplify it, and yes I agree

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: > It doesn't make sense to reload the snapshots every few seconds.

Is there a scenario that would cause snapshots to be reloaded every few seconds? The only way I would imagine them being reloaded is with sentinel key refresh.

zhenlan: Sure, you are right. Reloading the snapshot could be less frequent. Everything else stays true.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: I just overlooked it before, but SDK does throw. The snapshot null check shouldn't be necessary then.

jimmyca15: Right, no null check, but we do need to catch the exception and convert it to our exception.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: > I'm not sure why would refresh change what we are storing for snapshots?

Not sure I follow the question. When a sentinel key is changed and we need to reload all, don't we need to reload the snapshot if it's not cached?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan, line 1: The GitHub shows the whole file has changed in the last iteration. Is it just me?

amerjusupovic: Oh I must have changed the line endings again, there was a PR making them all uniform. I'll fix it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Ok, so do you mean instead of RequestFailedException change it something else, like ArgumentException? And I think I should be catching these exceptions in InitializeAsync then right? We wouldn't want these exceptions to stop the application if optional is true.

jimmyca15: > Ok, so do you mean instead of RequestFailedException change it something else, like ArgumentException? 

Yes that we shouldn't be throwing RequestFailedException, but I don't suggest ArgumentException. There is no invalid argument. The only argument the user is passing is a snapshot name which is valid. For lack of a better built-in option I'd go with InvalidOperationException, similarly to how we are doing for composition type.

jimmyca15: > And I think I should be catching these exceptions in InitializeAsync then right? We wouldn't want these exceptions to stop the application if optional is true.

Agreed

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Discussed offline with Zhenlan. 

The main scenario is sentinel key refresh
* Snapshots don't play a role in detection of sentinel key change
* A reload all operation would take a service dependency on more than snapshots, otherwise refresh wouldn't be configured. This waters down the value of snapshot caching.

We agreed that while there could be a perceived benefit of snapshot caching by reducing requests to the service during a refresh, it doesn't seem like the benefit is enough to justify the memory utilization for caching.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,17 +619,45 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
+                        snapshot = await client.GetSnapshotAsync(loadOption.SnapshotName).ConfigureAwait(false);
+                    }
+                    catch (RequestFailedException)
+                    {
+                        throw new InvalidOperationException($"Could not load snapshot with name '{loadOption.SnapshotName}'");
+                    }
+
+                    if (snapshot.CompositionType != CompositionType.Key)
+                    {
+                        throw new InvalidOperationException($"{nameof(snapshot.CompositionType)} for the selected snapshot must be 'key', found '{snapshot.CompositionType}'.");
```
zhenlan, line 650: Can we please include the snapshot name in the exception, as a user may select more than one snapshot?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -148,7 +149,7 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
                 throw new ArgumentException("The characters '*' and ',' are not supported in label filters.", nameof(labelFilter));
             }
 
-            if (!_kvSelectors.Any(s => s.KeyFilter.Equals(keyFilter) && s.LabelFilter.Equals(labelFilter)))
+            if (!_kvSelectors.Any(s => !string.IsNullOrEmpty(s.KeyFilter) && s.KeyFilter.Equals(keyFilter) && !string.IsNullOrEmpty(s.LabelFilter) && s.LabelFilter.Equals(labelFilter)))
```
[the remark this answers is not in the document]
jimmyca15: I'm curious about the addition of the nullorempty checks for key filter and label filter. Why are they needed, and also, it doesn't look like it's possible for them to be null or empty given the parameter checks above.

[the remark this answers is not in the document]
jimmyca15: Ah, I see because it's used for snapshot now. Perhaps it would be better to check for snapshot null or empty

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -160,6 +161,29 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
...
+        /// </summary>
+        /// <param name="name">The name of the snapshot in Azure App Configuration.</param>
+        public AzureAppConfigurationOptions SelectSnapshot(string name)
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if (!_kvSelectors.Any(s => !string.IsNullOrEmpty(s.SnapshotName) && s.SnapshotName.Equals(name)))
```
jimmyca15, line 176: `!string.IsNullOrEmpty(s.SnapshotName)`

That check doesn't seem necessary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -598,7 +599,7 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
             var serverData = new Dictionary<string, ConfigurationSetting>(StringComparer.OrdinalIgnoreCase);
 
             // Use default query if there are no key-values specified for use other than the feature flags
-            bool useDefaultQuery = !_options.KeyValueSelectors.Any(selector => !selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker));
+            bool useDefaultQuery = !_options.KeyValueSelectors.Any(selector => (!string.IsNullOrEmpty(selector.KeyFilter) && !selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker)) || !string.IsNullOrEmpty(selector.SnapshotName));
```
jimmyca15, line 602: Can you multi-line this. It's a bit long.

jimmyca15: It seems like it should be

`bool useDefaultQuery = !_options.KeyValueSelectors.Any(selector => selector.KeyFilter == null || !selector.KeyFilter.StartsWith(FeatureManagementConstants.FeatureFlagMarker));`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
amerjusupovic: They can be null if SelectSnapshot is called before a Select. I was originally setting keyfilter and labelfilter to a default value but we changed it to leaving it null because we decided it made more sense.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: SnapshotName can be null if Select is called before SelectSnapshot. Wasn't sure of a different way to handle it

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
jimmyca15: Would it be okay to write it as

if (!_kvSelectors.Any(s => string.Equals(s.KeyFilter, keyFilter) &&  string.Equals(s.LabelFilter, labelFilter)))

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
jimmyca15: if (!_kvSelectors.Any(s => string.Equals(s.SnapshotName, name)))

amerjusupovic: Thanks!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,17 +619,46 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
                     {
-                        KeyFilter = loadOption.KeyFilter,
-                        LabelFilter = loadOption.LabelFilter
-                    };
+                        snapshot = await client.GetSnapshotAsync(loadOption.SnapshotName).ConfigureAwait(false);
+                    }
+                    catch (RequestFailedException)
+                    {
+                        throw new InvalidOperationException($"Could not load snapshot with name '{loadOption.SnapshotName}'");
+                    }
```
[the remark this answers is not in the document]
amerjusupovic: That was the point, should we check for that specifically instead? We could see if it's a 404 and otherwise just rethrow the original exception?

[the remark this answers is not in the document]
jimmyca15: > We could see if it's a 404 and otherwise just rethrow the original exception?

Indeed. We should only catch an exception representing the error we want to handle. RequestFailedException is the propagation mechanism for other types of errors (e.g. authentication or authorization failure) that we don't want to catch here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,17 +620,51 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
+                }
+                else
+                {
+                    ConfigurationSettingsSnapshot snapshot;
+
+                    try
+                    {
+                        snapshot = await client.GetSnapshotAsync(loadOption.SnapshotName).ConfigureAwait(false);
+                    }
+                    catch (RequestFailedException rfe)
```
jimmyca15, line 645: `catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.NotFound)`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,17 +620,51 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
                     {
-                        KeyFilter = loadOption.KeyFilter,
-                        LabelFilter = loadOption.LabelFilter
-                    };
+                        if (rfe.Status == (int)HttpStatusCode.NotFound)
+                        {
+                            throw new InvalidOperationException($"Could not load snapshot with name '{loadOption.SnapshotName}'");
```
jimmyca15, line 649: We should be specific. The snapshot was not found. Saying we could not load it could allude to a multitude of issues.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -562,7 +562,8 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
             catch (Exception exception) when (ignoreFailures &&
                                              (exception is RequestFailedException ||
                                              ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
-                                             exception is OperationCanceledException))
+                                             exception is OperationCanceledException ||
+                                             exception is InvalidOperationException))
             { }
```
jimmyca15, line 567: We also need to add handling in TryRefreshAsync

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -618,17 +620,46 @@ await foreach (ConfigurationSetting setting in client.GetConfigurationSettingsAs
...
+                    try
                     {
-                        KeyFilter = loadOption.KeyFilter,
-                        LabelFilter = loadOption.LabelFilter
-                    };
+                        snapshot = await client.GetSnapshotAsync(loadOption.SnapshotName).ConfigureAwait(false);
+                    }
+                    catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.NotFound)
+                    {
+                        throw new InvalidOperationException($"Could not find snapshot with name '{loadOption.SnapshotName}'.");
```
jimmyca15, line 647: Inner exception needs to be chained.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -14,6 +14,7 @@ internal class LoggingConstants
         public const string RefreshCanceledError = "A refresh operation was canceled.";
         public const string RefreshFailedDueToKeyVaultError = "A refresh operation failed while resolving a Key Vault reference.";
         public const string PushNotificationUnregisteredEndpoint = "Ignoring the push notification received for the unregistered endpoint";
+        public const string RefreshFailedDueToInvalidOperation = "A refresh operation failed due to an invalid operation.";
```
jimmyca15, line 17: The existing RefreshFailedError seems enough.

jimmyca15 (approved):

jimmyca15:
Needs an update to reference the preview app configuration sdk.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -27,7 +27,8 @@ public class AzureAppConfigurationOptions
         private List<IKeyValueAdapter> _adapters = new List<IKeyValueAdapter>() 
         { 
             new AzureKeyVaultKeyValueAdapter(new AzureKeyVaultSecretProvider()),
-            new JsonKeyValueAdapter() 
+            new JsonKeyValueAdapter(),
+            new FeatureManagementKeyValueAdapter()
```
avanigupta, line 31: Why are we adding the feature management adapter by default? This would mean that we are flattening feature flags even when `UseFeatureFlags` is not specified.

jimmyca15: There was an impromptu discussion about this. the conclusion was that we need to handle feature flags that are in snapshots, even if UseFeatureFlags isn't called. The proposal was to add the feature flag adapater by default similarly to json and key vault adapters.

avanigupta: That's what I thought, but we should call this out as a potential breaking change for users who might be expecting raw feature flags in their application.

jimmyca15: Sounds like a major version bump is needed for select snapshot.

avanigupta (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
elonmallin: Hi, I just tried the latest preview version 7.0.0-preview.2 and tried to load a snapshot that has the key+label composition type and got this exception. Is the key+label just not implemented yet? I assume it will be since it's a little strange to be able to create those snapshots but not load them or am I missing something?

PR #429, opened 2023-06-06
Add environment variable for disabling provider

amerjusupovic:
### Disable Azure App Configuration for specific hosting environment:

Set the new environment variable `AZURE_APP_CONFIGURATION_PROVIDER_DISABLED` equal to "true" to disable Azure App Configuration in your application. Multiple environments can be set up in a variety of ways, including [using a `launchSettings.json` file](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-7.0#development-and-launchsettingsjson) which we recommend for local machine development.

If you don't already have a file at `Properties/launchSettings.json` in your project, you can add one.

#### ASP.NET Core

Here is an example of a `launchSettings.json` file for an ASP.NET core app called `ConfigStoreDemo` where Azure App Configuration is disabled for development, but not in production:

```
{
  "profiles": {
    "ConfigStoreDemo": {
      "commandName": "Project",
      "launchBrowser": true,
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development",
        "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED": "true"
      },
      "applicationUrl": "https://localhost:7152;http://localhost:5105"
    },
    "ConfigStoreDemo-Production": {
      "commandName": "Project",
      "launchBrowser": true,
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Production",
        "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED": "false"
      },
      "applicationUrl": "https://localhost:7152;http://localhost:5105"
    }
  }
}
```

#### .NET Core

Here is a similar `launchSettings.json` file, but for a .NET Core app called `ConsoleApplication`.

```
{
  "profiles": {
    "ConsoleApplication": {
      "commandName": "Project",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development",
        "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED": "true"
      }
    },
    "ConsoleApplication-Production": {
      "commandName": "Project",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Production",
        "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED": "false"
      }
    }
  }
}
```

In either example, you can edit this file directly and add the `AZURE_APP_CONFIGURATION_PROVIDER_DISABLED` environment variable under `environmentVariables`. Alternatively, if you're using Visual Studio 2022, you can edit the environment variables by going to the `Debug` tab in project properties and clicking `Open debug launch profiles UI`. This is explained further in the link above.


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -23,6 +24,11 @@ public AzureAppConfigurationSource(Action<AzureAppConfigurationOptions> optionsI
 
         public IConfigurationProvider Build(IConfigurationBuilder builder)
         {
+            if (Environment.GetEnvironmentVariable(ConditionalProviderConstants.DisableProviderEnvironmentVariable) == "true")
```
avanigupta, line 27: This can throw

avanigupta: I meant reading the environment variable could throw Security Exception. We should catch and ignore it in that case.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -23,6 +25,19 @@ public AzureAppConfigurationSource(Action<AzureAppConfigurationOptions> optionsI
 
         public IConfigurationProvider Build(IConfigurationBuilder builder)
         {
+            try
+            {
+                if (Environment.GetEnvironmentVariable(ConditionalProviderConstants.DisableProviderEnvironmentVariable) == "true")
```
avanigupta, line 30: This would require case insensitive comparison. Read the env var as a string and then use `bool.TryParse` to get the value.

avanigupta: Sample here: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/9a76d2a1b43d137e139e329a575eee09641e51d3/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L115

tests/Tests.AzureAppConfiguration/Tests.cs
```diff
@@ -263,5 +264,37 @@ public void TestTurnOffRequestTracing()
...
+        [Fact]
+        public void DisableProviderWithEnvironmentVariable()
+        {
+            var mockResponse = new Mock<Response>();
+            var mockClient = new Mock<ConfigurationClient>(MockBehavior.Strict);
+
+            mockClient.Setup(c => c.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()))
+                .Returns(new MockAsyncPageable(_kvCollectionPageOne));
+
+            Environment.SetEnvironmentVariable(ConditionalProviderConstants.DisableProviderEnvironmentVariable, "true");
```
avanigupta, line 277: This could lead to other tests failing when they are run in parallel. We need a way to run this test such that it never runs in parallel with any other test. Can you check if there's a way to run this one test sequentially and not in parallel?

avanigupta: This answer had some suggestions about using `Collection` attribute, but that may not work for our scenario: https://stackoverflow.com/questions/1408175/execute-unit-tests-serially-rather-than-in-parallel

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ConditionalProviderConstants.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Constants
+{
+    internal class ConditionalProviderConstants
+    {
+        public const string DisableProviderEnvironmentVariable = "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED";
```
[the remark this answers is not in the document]
amerjusupovic: Ok makes sense, I wasn't sure between the two options since we also have a file just for the HttpStatusCode constant that we use once too. I think I should just remove that file then too, what do you think?

tests/Tests.AzureAppConfiguration/Tests.cs
amerjusupovic: Oh ok thanks, yea I didn't consider that I'll look into it

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ConditionalProviderConstants.cs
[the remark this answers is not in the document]
avanigupta: You're right, that couldve been a private constant in the class where its used. However, I think we added a new file for that because HttpStatusCode is possibly extendible if we want to add more error codes in future. But this new environment variable doesn't fall in any category where we might need to add more constants in future. Lets keep the HttpStatusCode for now and remove this one.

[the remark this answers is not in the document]
amerjusupovic: Oh that's fair, this situation is definitely different. Thanks!

tests/Tests.AzureAppConfiguration/Tests.cs
amerjusupovic: I haven't been able to find anything too helpful on this. I think it might just be a good idea to remove this test for now, since the only solutions I've seen that might work complicate things significantly.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -23,6 +25,19 @@ public AzureAppConfigurationSource(Action<AzureAppConfigurationOptions> optionsI
...
+            {
+                providerDisabled = Environment.GetEnvironmentVariable(DisableProviderEnvironmentVariable);
+            }
+            catch (SecurityException) { }
+            
+            // Disable the provider if the environment variable evaluates to true
+            if (bool.TryParse(providerDisabled, out bool disabled) ? disabled : false)
+            {
+                return new EmptyConfigurationProvider();
+            }
```
avanigupta, line 39: How about this?

```cs
try
{
    if (bool.TryParse(Environment.GetEnvironmentVariable(DisableProviderEnvironmentVariable), out bool disabled) ? disabled : false)
    {
        return new EmptyConfigurationProvider();
    }
}
catch (SecurityException) { }
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -2,13 +2,15 @@
...
 using System;
+using System.Security;
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     internal class AzureAppConfigurationSource : IConfigurationSource
     {
         private readonly bool _optional;
         private readonly Func<AzureAppConfigurationOptions> _optionsProvider;
+        public const string DisableProviderEnvironmentVariable = "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED";
```
avanigupta, line 13: private const

tests/Tests.AzureAppConfiguration/Tests.cs
avanigupta: Agreed, its not worth taking on the additional complexity.

zhenlan:
The code looks good. Before we merge it, I just want to make sure this works end to end. @amerjusupovic, you mentioned  launch.json. Can you please share the steps a user needs to do for a .NET app and an ASP.NET app?

amerjusupovic:
@zhenlan Here's what I have so far, I'm going to add this and the other scenarios we discussed for working with multiple environments to the original issue as well.

### Disable Azure App Configuration for specific hosting environment:

Set the new environment variable `AZURE_APP_CONFIGURATION_PROVIDER_DISABLED` equal to "true" to disable Azure App Configuration in your application. Multiple environments can be set up in a variety of ways, including [using a `launchSettings.json` file](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-7.0#development-and-launchsettingsjson) which we recommend for local machine development.

If you don't already have a file at `Properties/launchSettings.json` in your project, you can add one.

#### ASP.NET Core

Here is an example of a `launchSettings.json` file for an ASP.NET core app called `ConfigStoreDemo` where Azure App Configuration is disabled for development, but not in production:

```
{
  "profiles": {
    "ConfigStoreDemo": {
      "commandName": "Project",
      "launchBrowser": true,
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development",
        "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED": "true"
      },
      "applicationUrl": "https://localhost:62606;http://localhost:62607"
    },
    "ConfigStoreDemo-Production": {
      "commandName": "Project",
      "launchBrowser": true,
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Production",
        "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED": "false"
      },
      "applicationUrl": "https://localhost:62606;http://localhost:62607"
    }
  }
}
```

#### .NET Core

Here is a `launchSettings.json` file for a .NET Core app called `ConsoleApplication` where Azure App Configuration is disabled.

```
{
  "profiles": {
    "ConsoleApplication": {
      "commandName": "Project",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development",
        "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED": "true"
      }
    }
  }
}
```

In either example, you can edit this file directly and add the `AZURE_APP_CONFIGURATION_PROVIDER_DISABLED` environment variable under `environmentVariables`. Alternatively, if you're using Visual Studio 2022, you can edit the environment variables by going to the `Debug` tab in project properties and clicking `Open debug launch profiles UI`. This is explained further in the link above.

avanigupta:
> @zhenlan @avanigupta I was also looking at [this exception message](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/9a76d2a1b43d137e139e329a575eee09641e51d3/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs#L27) since this is what's hit if we return an `EmptyConfigurationProvider` like in this PR. I wanted to either change it to "If you did not disable Azure App Configuration, please ensure that it has been configured correctly." or do something different here to provide more clarity. Also, is it ok to throw in this case?

That's a good catch! We cannot throw exceptions if someone wants to disable our provider in their dev environment. Their app should keep working without our provider. We should just skip adding our middleware when UseAzureAppConfiguration is invoked. We can make both AddAzureAppConfiguration and UseAzureAppConfiguration no-op when our provider is disabled.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
```diff
@@ -70,8 +85,20 @@ public static IServiceCollection AddAzureAppConfiguration(this IServiceCollectio
                 throw new ArgumentNullException(nameof(services));
             }
 
-            services.AddLogging();
-            services.AddSingleton<IConfigurationRefresherProvider, AzureAppConfigurationRefresherProvider>();
+            bool providerDisabled = false;
+
+            try
+            {
+                providerDisabled = bool.TryParse(Environment.GetEnvironmentVariable(ConditionalProviderConstants.DisableProviderEnvironmentVariable), out bool disabled) ? disabled : false;
```
zhenlan, line 92: Can we avoid reading the environment variable repeatedly?

amerjusupovic: From what I understand it's only called during startup and could be repeated once in `UseAzureAppConfiguration`, and the only idea I had was storing it in the constants class I read from to get the environment variable name, but that didn't seem like the correct way to use the class. An exception is thrown as soon as `GetService` is called if the provider is disabled.

zhenlan: I hope the environment variable is only read once and the decision is only made once. This ensures a consistent state. Yes, we can save it as a static variable at the startup to avoid rereading the environment variable.

I hope we can avoid reading the environment variable in the middleware too... Should we update the `AddAzureAppConfiguration(this IServiceCollection services)` too? I'm wondering if we can do something special there so the middleware knows.

avanigupta: Its possible to keep all the environment variable logic in a single place by adding a new dummy refresher provider that implements `IConfigurationRefresherProvider`.

We only need to read the environment var once in the static constructor of `AzureAppConfigurationExtensions` class and store it in a private variable. Then we can use this variable to decide whether to add `AzureAppConfigurationSource` or skip it in `AddAzureAppConfiguration(this IConfigurationBuilder configurationBuilder,Action<AzureAppConfigurationOptions> action,bool optional = false)` method.
Additionally, we can add `EmptyRefresherProvider` instead of `AzureAppConfigurationRefresherProvider` in `AddAzureAppConfiguration(this IServiceCollection services)` method when provider is disabled. 

Then we dont need to read the variable in any other file and can eliminate the new constants file as well.

zhenlan: Like it! Just one thing... We can probably do a stack method in `AzureAppConfigurationSource` instead of a static constructor. Failure of a static constructor can cause weird results.

amerjusupovic: I'm not aware of what we could do with the `IServiceCollection` object, but I can look into it. I think the static variable makes sense.

avanigupta: > We can probably do a stack method in `AzureAppConfigurationSource` instead of a static constructor.

@zhenlan Did you mean static method inside `AzureAppConfigurationExtensions`?

zhenlan: Kinda weird to put it there, as it's not an extension method.

avanigupta: `AzureAppConfigurationSource` is not static and we dont need to add a new instance unless provider is enabled. That's why I thought it might be best to either use static constructor or static field initializer within `AzureAppConfigurationExtensions`.

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshExtensions.cs
```diff
@@ -30,7 +30,12 @@ public static IApplicationBuilder UseAzureAppConfiguration(this IApplicationBuil
                 throw new InvalidOperationException("Unable to find the required services. Please add all the required services by calling 'IServiceCollection.AddAzureAppConfiguration' inside the call to 'ConfigureServices(...)' in the application startup code.");
             }
 
-            return builder.UseMiddleware<AzureAppConfigurationRefreshMiddleware>();
+            if (builder.ApplicationServices.GetService(typeof(EmptyRefresherProvider)) == null)
```
zhenlan, line 33: Is it possible we don't make `EmptyRefresherProvider` a public class?

amerjusupovic: I can look into it but I wasn't sure how. Also side note, should it be `EmptyConfigurationRefresherProvider`?

avanigupta: We can make it internal and use `InternalsVisibleTo` attribute (like we expose the internals to test assemblies).
https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-7.0

avanigupta: Although it might be weird to expose all internals just for this one type.

zhenlan: We get an instance of `IConfigurationRefresherProvider`. Can we ignore if its `Refreshers` is empty?

amerjusupovic: I think it makes sense since if there are no Refreshers the middleware is basically a no-op anyway.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/EmptyConfigurationRefresherProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT license.
+//
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    /// <summary>
+    /// A dummy class used to show that <see cref="AzureAppConfigurationRefresherProvider"/> is not available.
+    /// </summary>
+    public class EmptyConfigurationRefresherProvider : IConfigurationRefresherProvider
```
avanigupta, line 11: Doesnt need to be public now

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/EmptyConfigurationRefresherProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        /// <summary>
+        /// A dummy variable to implement the interface <see cref="IConfigurationRefresherProvider"/>.
+        /// </summary>
+        public IEnumerable<IConfigurationRefresher> Refreshers => new List<IConfigurationRefresher>();
```
avanigupta, line 16: Can we use `Enumerable.Empty` to be more explicit?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationExtensions.cs
```diff
@@ -13,6 +14,20 @@ namespace Microsoft.Extensions.Configuration
     /// </summary>
     public static class AzureAppConfigurationExtensions
     {
+        private const string DisableProviderEnvironmentVariable = "AZURE_APP_CONFIGURATION_PROVIDER_DISABLED";
+        private static readonly bool _providerDisabled = IsProviderDisabled();
```
avanigupta, line 18: nit: `_isProviderDisabled `

avanigupta (approved):

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshExtensions.cs
```diff
@@ -23,14 +24,21 @@ public static IApplicationBuilder UseAzureAppConfiguration(this IApplicationBuil
...
             }
 
+            IConfigurationRefresherProvider refresherProvider = (IConfigurationRefresherProvider)builder.ApplicationServices.GetService(typeof(IConfigurationRefresherProvider));
+
             // Verify if AddAzureAppConfiguration was done before calling UseAzureAppConfiguration.
             // We use the IConfigurationRefresherProvider to make sure if the required services were added.
-            if (builder.ApplicationServices.GetService(typeof(IConfigurationRefresherProvider)) == null)
+            if (refresherProvider == null)
             {
                 throw new InvalidOperationException("Unable to find the required services. Please add all the required services by calling 'IServiceCollection.AddAzureAppConfiguration' inside the call to 'ConfigureServices(...)' in the application startup code.");
```
zhenlan, line 33: There is no `ConfigureServices` anymore since .NET 6. If you prefer, we can change it in a separate PR.

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshExtensions.cs
```diff
@@ -23,14 +24,21 @@ public static IApplicationBuilder UseAzureAppConfiguration(this IApplicationBuil
...
             }
 
-            return builder.UseMiddleware<AzureAppConfigurationRefreshMiddleware>();
+            if (refresherProvider.Refreshers.Count() != 0)
```
zhenlan, line 36: `refresherProvider.Refreshers` could be null.

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshExtensions.cs
amerjusupovic: I guess it is a pretty small change, would it just be removing the part saying `inside the call to 'ConfigureServices(...)'`?

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshExtensions.cs
```diff
@@ -23,14 +24,21 @@ public static IApplicationBuilder UseAzureAppConfiguration(this IApplicationBuil
...
+
             // Verify if AddAzureAppConfiguration was done before calling UseAzureAppConfiguration.
             // We use the IConfigurationRefresherProvider to make sure if the required services were added.
-            if (builder.ApplicationServices.GetService(typeof(IConfigurationRefresherProvider)) == null)
+            if (refresherProvider == null)
+            {
+                throw new InvalidOperationException("Unable to find the required services. Please add all the required services by calling 'IServiceCollection.AddAzureAppConfiguration' in the application startup code.");
+            }
+
+            if (refresherProvider.Refreshers?.Count() != 0)
```
avanigupta, line 36: If Refreshers is null, this condition will still be true (null != 0).
You could use `if (refresherProvider.Refreshers?.Count() > 0)`

src/Microsoft.Azure.AppConfiguration.Functions.Worker/AzureAppConfigurationRefreshExtensions.cs
```diff
@@ -23,14 +23,22 @@ public static class AzureAppConfigurationRefreshExtensions
...
         {
+            IServiceProvider serviceProvider = builder.Services.BuildServiceProvider();
+            IConfigurationRefresherProvider refresherProvider = serviceProvider.GetService<IConfigurationRefresherProvider>();
+
             // Verify if AddAzureAppConfiguration was done before calling UseAzureAppConfiguration.
             // We use the IConfigurationRefresherProvider to make sure if the required services were added.
-            if (!builder.Services.Any(service => service.ServiceType == typeof(IConfigurationRefresherProvider)))
+            if (refresherProvider == null)
+            {
+                throw new InvalidOperationException($"Unable to find the required services. Please add all the required services by calling '{nameof(IServiceCollection)}.{nameof(AzureAppConfigurationExtensions.AddAzureAppConfiguration)}()' in the application startup code.");
```
zhenlan, line 33: > '{nameof(IServiceCollection)}.{nameof(AzureAppConfigurationExtensions.AddAzureAppConfiguration)}()'

Any reason we do it differently in this message compared to the one used in ASP.NET middleware?

amerjusupovic: I tried to change it, but then some using errors came up that were preventing me from using the second `nameof()`. I added the same using statements as the functions worker file and it wouldn't recognize it still.

avanigupta: Unfortunately, the name of this class in AspNetCore package is same as the name of Extensions class with AddAzureAppConfiguration method. This class in AspNetCore package should've been called `AzureAppConfigurationRefreshExtensions`, but it would be a breaking change to modify the name now. 
Try using full class name along with namespace to explicitly point to the right class:
`nameof(Microsoft.Extensions.Configuration.AzureAppConfigurationExtensions.AddAzureAppConfiguration)`

amerjusupovic: Looks like it works, no error and it gets the method name

avanigupta (approved):

zhenlan (approved):

avanigupta (approved):

PR #431, opened 2023-06-20
Add warning level log for failover to new endpoint

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/LogHelper.cs
```diff
@@ -74,5 +74,10 @@ public static string BuildPushNotificationUnregisteredEndpointMessage(string res
         {
             return $"{LoggingConstants.PushNotificationUnregisteredEndpoint} '{resourceUri}'.";
         }
+
+        public static string BuildFailoverToDifferentEndpointMessage(string originalEndpoint, string currentEndpoint)
+        {
+            return $"{LoggingConstants.RefreshFailoverToDifferentEndpoint} Original Endpoint:'{originalEndpoint.TrimEnd('/')}' Current Endpoint:'{currentEndpoint.TrimEnd('/')}'";
```
amerjusupovic, line 80: Not sure about having both the original and current endpoint, if it seems unnecessary I can change it to being just the current one.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -827,6 +829,13 @@ private async Task<T> ExecuteWithFailOverPolicyAsync<T>(IEnumerable<Configuratio
                     T result = await funcToExecute(currentClient).ConfigureAwait(false);
                     success = true;
 
+                    Uri currentEndpoint = _configClientManager.GetEndpointForClient(currentClient);
+
+                    if (failureOccurred && originalEndpoint != currentEndpoint)
```
avanigupta, line 834: This would log only when (if) there is a successful operation. We should be logging all endpoints in the same order that they were tried in.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/LogHelper.cs
avanigupta: I think it makes sense to log previous and next endpoints. I'm imagining logs like this:

`Failed to get configuration settings from 'contoso.azconfig.io'. Trying to read from 'contoso-replica1.azconfig.io'`

`Failed to get configuration settings from 'contoso-replica1.azconfig.io'. Trying to read from 'contoso-replica2.azconfig.io'`

amerjusupovic: Makes sense, do you think I should remove the final log saying what endpoint succeeded then or keep it along with these? I think it just clarifies which endpoint succeeded, but I can also understand that the last log of the ones above could imply that the last endpoint worked.

avanigupta: We dont need warning level logs for successful refresh operation.

amerjusupovic: Oh didn't think of that way, sounds good

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -824,6 +826,15 @@ private async Task<T> ExecuteWithFailOverPolicyAsync<T>(IEnumerable<Configuratio
 
                 try
                 {
+                    Uri currentEndpoint = _configClientManager.GetEndpointForClient(currentClient);
+
+                    if (failureOccurred && previousEndpoint != currentEndpoint)
```
avanigupta, line 831: We dont need to maintain `failureOccurred` variable. If `previousEndpoint != currentEndpoint`, we know that failure must've occurred.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -849,6 +860,8 @@ private async Task<T> ExecuteWithFailOverPolicyAsync<T>(IEnumerable<Configuratio
                 }
                 finally
                 {
+                    failureOccurred = !success;
+
                     if (!success && backoffAllClients)
                     {
                         do
```
avanigupta, line 867: We need one more log statement when all endpoints have failed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs
```diff
@@ -29,5 +29,8 @@ internal class LoggingConstants
 
         // Other
         public const string RefreshSkippedNoClientAvailable = "Refresh skipped because no endpoint is accessible.";
+        public const string RefreshFailedToGetSettingsFromEndpoint = "Failed to get configuration settings from endpoint";
+        public const string RefreshTryingToReadFromEndpoint = "Trying to read from endpoint";
+        public const string RefreshAllEndpointsFailed = "Failed to read from all endpoints.";
```
avanigupta, line 34: This would be logged even if users set only one endpoint (no replicas). I think we can reuse RefreshFailedToGetSettingsFromEndpoint and log the previous endpoint that failed.

amerjusupovic: Oh I see, I understood it as a message to show that all endpoints failed, but makes sense that we're missing a log for the last failed endpoint.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -851,6 +861,8 @@ private async Task<T> ExecuteWithFailOverPolicyAsync<T>(IEnumerable<Configuratio
                 {
                     if (!success && backoffAllClients)
                     {
+                        _logger.LogWarning(LogHelper.BuildLastEndpointFailedMessage(_configClientManager.GetEndpointForClient(currentClient).ToString()));
```
avanigupta, line 864: We dont need to get endpoint again - we can reuse previousEndpoint  or currentEndpoint

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/LogHelper.cs
```diff
@@ -74,5 +74,15 @@ public static string BuildPushNotificationUnregisteredEndpointMessage(string res
         {
             return $"{LoggingConstants.PushNotificationUnregisteredEndpoint} '{resourceUri}'.";
         }
+
+        public static string BuildFailoverToDifferentEndpointMessage(string originalEndpoint, string currentEndpoint)
+        {
+            return $"{LoggingConstants.RefreshFailedToGetSettingsFromEndpoint} '{originalEndpoint.TrimEnd('/')}'. {LoggingConstants.RefreshTryingToReadFromEndpoint} '{currentEndpoint.TrimEnd('/')}'.";
```
jimmyca15, line 80: Is the trim necessary?

amerjusupovic: It was the pattern we followed with the original PR that added logging, since the endpoints always have a '/' at the end.

jimmyca15: Okay.

Parameters need validation for null.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -838,6 +839,15 @@ private async Task<T> ExecuteWithFailOverPolicyAsync<T>(IEnumerable<Configuratio
 
                 try
                 {
+                    Uri currentEndpoint = _configClientManager.GetEndpointForClient(currentClient);
+
+                    if (previousEndpoint != currentEndpoint)
+                    {
+                        _logger.LogWarning(LogHelper.BuildFailoverToDifferentEndpointMessage(previousEndpoint.ToString(), currentEndpoint.ToString()));
+                    }
```
jimmyca15, line 847: This seems like it should be after the finally block.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/LogHelper.cs
```diff
@@ -74,5 +74,15 @@ public static string BuildPushNotificationUnregisteredEndpointMessage(string res
         {
             return $"{LoggingConstants.PushNotificationUnregisteredEndpoint} '{resourceUri}'.";
         }
+
+        public static string BuildFailoverToDifferentEndpointMessage(string originalEndpoint, string currentEndpoint)
```
jimmyca15, line 78: How about just BuildFailoverMessage?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/LogHelper.cs
```diff
@@ -74,5 +74,15 @@ public static string BuildPushNotificationUnregisteredEndpointMessage(string res
...
+        {
```
jimmyca15, line 79: For the message it looks like it's currently

"Failed to get configuration settings from endpoint '{old endpoint}'. Trying to read from endpoint '{new endpoint}'."

I'm thinking to make the fall back clear:

"Failed to get configuration settings from endpoint '{old endpoint}'. Falling back to endpoint '{new endpoint}'."

avanigupta: That's a good idea, can we use "failover" instead of "fallback" though? That's the term we use in public docs too.

"Failed to get configuration settings from endpoint '{old endpoint}'. Initiating automatic failover to endpoint '{new endpoint}'."

jimmyca15: Yes. "Failing over to endpoint ..."

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/LogHelper.cs
avanigupta: Good catch! We need it throughout this helper class.

amerjusupovic: So just for the endpoint parameters because we call `TrimEnd` right?

jimmyca15: Yes. Without parameter validation it's a possible null-ref.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: I agree from how it looks, but then it seems like I need to either make an extra call to `GetEndpointforClient` or update `currentEndpoint` at the bottom of the while loop and I wasn't sure which made more sense.

PR #432, opened 2023-06-30
Prevent access to HttpContext throught HttpContextAccessor in TryRefreshAsync

amerjusupovic:


src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
```diff
@@ -24,9 +25,12 @@ public AzureAppConfigurationRefreshMiddleware(RequestDelegate next, IConfigurati
 
         public async Task InvokeAsync(HttpContext context)
         {
-            foreach (var refresher in Refreshers)
+            using (var flowControl = ExecutionContext.SuppressFlow())
```
jimmyca15, line 28: I think it needs a comment to explain the reasoning. A suggestion
```
//
// Configuration refresh is meant to execute as an isolated background task.
// To prevent access of request-based resources, such as HttpContext, we detach the execution context within the refresh operation.
```

jimmyca15: Did you get a chance to iterate on this?

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
```diff
@@ -24,9 +25,15 @@ public AzureAppConfigurationRefreshMiddleware(RequestDelegate next, IConfigurati
 
         public async Task InvokeAsync(HttpContext context)
         {
-            foreach (var refresher in Refreshers)
+            //
+            // Configuration refresh is meant to execute as an isolated background task.
+            // To prevent access of request-based resources, such as HttpContext, we suppress the execution context within the refresh operation.
```
jimmyca15, line 30: `using (var flowControl = ExecutionContext.SuppressFlow())`

Can you use an explicit type. It's hard to tell what type `flowControl` is from the right hand side.

jimmyca15:
Looks good, but need to figure out the destination branch.

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
```diff
@@ -24,9 +25,15 @@ public AzureAppConfigurationRefreshMiddleware(RequestDelegate next, IConfigurati
...
+            using (AsyncFlowControl flowControl = ExecutionContext.SuppressFlow())
             {
-                _ = refresher.TryRefreshAsync();
+                foreach (IConfigurationRefresher refresher in Refreshers)
+                {
+                    _ = Task.Run(() => refresher.TryRefreshAsync());
+                }
             }
```
Tratcher, line 37: This seems like an unnecessary amount of work to do on every request. Assume the server is running at 100k RPS. I assume the refreshers themselves are throttled? You still want a check in the middleware to call them at most every 60s(?).

jimmyca15: It is mostly no-op. Refresh only takes action on user-configured expiration interval (default 30s).

src/Microsoft.Azure.AppConfiguration.Functions.Worker/AzureAppConfigurationRefreshMiddleware.cs
```diff
@@ -24,9 +24,15 @@ public AzureAppConfigurationRefreshMiddleware(IConfigurationRefresherProvider re
...
-            foreach (IConfigurationRefresher refresher in Refreshers)
+            //
+            // Configuration refresh is meant to execute as an isolated background task.
+            // To prevent access of request-based resources, such as HttpContext, we suppress the execution context within the refresh operation.
+            using (AsyncFlowControl flowControl = ExecutionContext.SuppressFlow())
             {
-                _ = refresher.TryRefreshAsync();
+                foreach (IConfigurationRefresher refresher in Refreshers)
+                {
+                    _ = Task.Run(() => refresher.TryRefreshAsync());
```
jimmyca15, line 34: Can this be `_ = Task.Run(refresher.TryRefreshAsync);`

amerjusupovic: Did you mean something else? It `cannot convert from 'method group' to 'System.Action'`.

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
```diff
@@ -24,9 +27,20 @@ public AzureAppConfigurationRefreshMiddleware(RequestDelegate next, IConfigurati
...
+                // To prevent access of request-based resources, such as HttpContext, we suppress the execution context within the refresh operation.
+                using (AsyncFlowControl flowControl = ExecutionContext.SuppressFlow())
+                {
+                    foreach (IConfigurationRefresher refresher in Refreshers)
+                    {
+                        _ = Task.Run(() => refresher.TryRefreshAsync());
+                    }
+                }
+
+                refreshReadyTime = DateTimeOffset.UtcNow.Add(TimeSpan.FromSeconds(1));
```
Tratcher, line 43: Set this before doing the work to prevent other concurrent requests from doing duplicate work. I'd also suggest raising it to 15-60s, config doesn't change very often.

Also, store UtcNow in a variable to avoid calling it twice.

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
```diff
@@ -13,7 +15,10 @@ namespace Microsoft.Azure.AppConfiguration.AspNetCore
     /// </summary>
     internal class AzureAppConfigurationRefreshMiddleware
     {
+        // The minimum refresh interval on the configuration provider is 1 second, so refreshing more often is unnecessary
+        private static readonly long MinimumRefreshInterval = TimeSpan.FromSeconds(1).Ticks;
         private readonly RequestDelegate _next;
+        private long _refreshReadyTime = DateTimeOffset.UtcNow.Ticks;
         public IEnumerable<IConfigurationRefresher> Refreshers { get; }
```
jimmyca15, line 22: Can you add a new-line before `Refreshers` to separate properties vs fields.

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
amerjusupovic: Thans for the comment, added some logic with help from @jimmyca15 that should help prevent calling the logic in the if statement more than once per second. I don't think we can raise the interval though, since our current minimum refresh interval is 1 second and we wouldn't want to change that.

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshMiddleware.cs
```diff
@@ -24,9 +29,21 @@ public AzureAppConfigurationRefreshMiddleware(RequestDelegate next, IConfigurati
 
         public async Task InvokeAsync(HttpContext context)
         {
-            foreach (var refresher in Refreshers)
+            long refreshReadyTime = Interlocked.Read(ref _refreshReadyTime);
+
+            if (refreshReadyTime <= DateTimeOffset.UtcNow.Ticks &&
```
jimmyca15, line 34: DateTimeOffset.UtcNow.Ticks needs a variable, because it should be used in the new value of `_refreshReadyTime`

amerjusupovic: Missed that, thanks

jimmyca15 (approved):

avanigupta (approved):

PR #440, opened 2023-07-21
Catch OperationCanceledException in AggregateException when all refresh retries fail

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -420,7 +420,7 @@ public async Task<bool> TryRefreshAsync(CancellationToken cancellationToken)
 
                 return false;
             }
-            catch (AggregateException e) when (e?.InnerExceptions?.All(e => e is RequestFailedException) ?? false)
+            catch (AggregateException e) when (e?.InnerExceptions?.All(e => e is RequestFailedException || e is OperationCanceledException) ?? false)
```
avanigupta, line 423: We are handling AggregateException where all inner exceptions are OperationCanceledException. This might not be the intended behavior. I think we should handle AggregateException only when at least one inner exception is RequestFailedException. As for other inner exceptions, I'm conflicted about handling OperationCanceledException explicitly vs handling AggregateException with any inner exception as long as there is at least one RequestFailedException. 
We got a customer issue where they saw AggregateException with 2 RequestFailedException and 1 OperationCanceledException. I dont know if there are any other exceptions that SDK might throw inside AggregateException, but for future-proofing the code, I feel like we should handle every AggregateException which has at least one RequestFailedException. 

@jimmyca15 thoughts?

jimmyca15: My thoughts --

We handle RequestFailedException, OperationCanceledException, KeyVaultReferenceException.

AggregateException is a holder of exceptions that we have to account for. Usually the way I see it handled is to delegate to known inner exception type. So what that would mean here is that if AggregateException contains any of the exceptions we know (RequestFailedException, OperationCanceledException, KeyVaultReferenceException), which we are already swallowing, then we should swallow it as well.

jimmyca15: Also, I would intentionally order AggregateException as the last caught exception type.

jimmyca15: Something like

```
catch (AggregateException ae)
{
    if (ae.Any(e is OperationCanceledException))
    {
        _logger.LogWarning(LogHelper.BuildRefreshCanceledErrorMessage());
    }
    else if (ae.Any(e is RequestFailedException))
    {
        if (IsAuthenticationError(e))
        {
            _logger.LogWarning(LogHelper.BuildRefreshFailedDueToAuthenticationErrorMessage(e.Message));
        }
        else
        {
            _logger.LogWarning(LogHelper.BuildRefreshFailedErrorMessage(e.Message));
        }
    }
    else if (ae.Any(e is KeyVaultReferenceException))
    {
        _logger.LogWarning(LogHelper.BuildRefreshFailedDueToKeyVaultErrorMessage(e.Message));
    }
    else
    {
        throw;
    }
}
```

avanigupta: We dont need to worry about KeyVaultReferenceException here because that's our internal exception, so it would never be wrapped inside AggregateException. The logic looks good otherwise.

jimmyca15: In general that's how I would expect aggregate exception to look regardless of possibilities. It is merely possible holder and is delegated to exceptions we already handle.

avanigupta: When we catch an AggregateException, we usually just care about the RequestFailedException inside it to determine if its an auth error or if its a failover-able error. We dont really care about operation cancelled if its accompanied by another RequestFailedException (right?). So keeping the complete AggregateException handling pattern may not be necessary in our case. We can trim it down to something more concise:

```
catch (RequestFailedException rfe)
{
}
catch (KeyVaultReferenceException kvre)
{
}
catch (OperationCanceledException oce)
{
}
catch (AggregateException ae) when (ae?.InnerExceptions?.Any(e => e is RequestFailedException || e is OperationCanceledException) ?? false)
{
}
```

And the helper methods `IsAuthenticationError()` and `IsFailOverable()` can be updated to find the RequestFailedException inside AggregateException and log appropriately.

jimmyca15: Inside the catch you put, are you expecting to have 

```

            if (ae.InnerExceptions.Any(e => e is e is OperationCanceledException))
            {
                _logger.LogWarning(LogHelper.BuildRefreshCanceledErrorMessage());
            }
```

?

If so, I'd avoid using the `when` clause to avoid repetition.

avanigupta: I was not expecting the AggregateException catch block to have any changes. When we catch an AggregateException here, its either an authentication error or generic refresh failed error.

jimmyca15: Well that doesn't seem to add up for the case where AggregateException only has an OperationCanceledException in it. The handling wouldn't be the same as the catch block for OperationCanceledException.

avanigupta: Ah I see. Okay, so we can remove the when clause and handle inner exceptions inside the catch block. 
We should still make sure that RequestFailedException is given preference over OperationCancelledException when both are present as inner exceptions.

jimmyca15: I'm fine with it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -561,7 +572,7 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
             }
             catch (Exception exception) when (ignoreFailures &&
                                              (exception is RequestFailedException ||
-                                             ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
+                                             ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException || e is OperationCanceledException) ?? false) ||
```
avanigupta, line 575: There is a mismatch between the `Any` logic in TryRefreshAsync and `All` logic here. We should use the same exception handling logic in all places.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -881,9 +892,9 @@ private bool IsFailOverable(AggregateException ex)
         {
             IReadOnlyCollection<Exception> innerExceptions = ex.InnerExceptions;
```
avanigupta, line 893: Can you also change `ExecuteWithFailOverPolicyAsync` method to catch AggregateException after other exceptions?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -881,9 +892,9 @@ private bool IsFailOverable(AggregateException ex)
...
 
-            if (innerExceptions != null && innerExceptions.Any() && innerExceptions.All(ex => ex is RequestFailedException))
+            if (innerExceptions != null && innerExceptions.Any(ex => ex is RequestFailedException))
```
avanigupta, line 895: This method can be simplified to the following:
```
private bool IsFailOverable(AggregateException ex)
{
    RequestFailedException rfe = ex.InnerExceptions?.LastOrDefault(e => e is RequestFailedException) as RequestFailedException;

    return rfe != null ? IsFailOverable(rfe) : false;
}
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -879,14 +890,9 @@ private async Task ExecuteWithFailOverPolicyAsync(IEnumerable<ConfigurationClien
 
         private bool IsFailOverable(AggregateException ex)
         {
-            IReadOnlyCollection<Exception> innerExceptions = ex.InnerExceptions;
+            RequestFailedException rfe = ex.InnerExceptions?.LastOrDefault(e => e is RequestFailedException) as RequestFailedException;
```
jimmyca15, line 893: Prefer FirstOrDefault over LastOrDefault to avoid unnecessary enumeration.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -561,8 +572,8 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
             }
             catch (Exception exception) when (ignoreFailures &&
                                              (exception is RequestFailedException ||
-                                             ((exception as AggregateException)?.InnerExceptions?.All(e => e is RequestFailedException) ?? false) ||
-                                             exception is OperationCanceledException))
+                                             (exception is OperationCanceledException) ||
+                                             ((exception as AggregateException)?.InnerExceptions?.Any(e => e is RequestFailedException || e is OperationCanceledException) ?? false)))
             { }
```
jimmyca15, line 577: The indentation is a bit off

```
            catch (Exception exception) when (
                ignoreFailures &&
                (exception is RequestFailedException ||
                exception is OperationCanceledException ||
                (exception as AggregateException?).InnerExceptions?.Any(e => e is RequestFailedException || e is OperationCanceledException) ?? false))
            { }
```

jimmyca15 (approved):

avanigupta (approved):

PR #449, opened 2023-08-18
Update UseFeatureFlags for snapshot changes

amerjusupovic:


avanigupta (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -188,7 +188,7 @@ public AzureAppConfigurationOptions SelectSnapshot(string name)
         }
 
         /// <summary>
-        /// Enables Azure App Configuration feature flags to be parsed and transformed into feature management configuration.
+        /// Configures options for Azure App Configuration feature flags that will be parsed and transformed into feature management configuration.
```
jimmyca15, line 191: We're missing a bit here.

* Selects feature flags
* Allows filtering of feature flags
* Set's up individual feature flag watching for selected feature flags.

amerjusupovic: Should we mention the feature flag selecting/filtering even though it's summarized in `FeatureFlagOptions`? I think the watching isn't mentioned anywhere currently though.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -188,7 +188,9 @@ public AzureAppConfigurationOptions SelectSnapshot(string name)
...
+        /// If no feature flags are selected, all feature flags with the label set in options, or null label if not set, are included in the configuration.
```
jimmyca15, line 192:
```suggestion
        /// If no filtering is specified via the <cref="FeatureFlagOptions"> then all feature flags with no label are loaded.
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -188,7 +188,9 @@ public AzureAppConfigurationOptions SelectSnapshot(string name)
...
+        /// Sets up individual feature flag refresh for any feature flags included using the configured options.
```
jimmyca15, line 193:
```suggestion
        /// All loaded feature flags will be automatically registered for refresh on an individual flag level.
```

jimmyca15 (approved):

PR #451, opened 2023-08-23
Upgrade packages to 7.0.0-preview2

amerjusupovic:


src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
```diff
@@ -28,7 +28,7 @@
   <!-- Nuget Package Version Settings -->
 
   <PropertyGroup>
-    <OfficialVersion>7.0.0-preview</OfficialVersion>
+    <OfficialVersion>7.0.0-preview2</OfficialVersion>
```
avanigupta, line 31: I think we should call it `7.0.0-preview.2`. There's some guidance for naming pre-release versions with dot separated identifiers here: https://semver.org/#spec-item-9

cc @jimmyca15

avanigupta: And SDK follows a similar pattern: https://www.nuget.org/packages/Azure.Data.AppConfiguration/1.0.0-preview.6

amerjusupovic: Makes sense, looks like most other Microsoft packages follow that pattern.

avanigupta (approved):

PR #460, opened 2023-09-13
Add requirement type parsing and upgrade SDK version

amerjusupovic:
We can now use the changes (with slight adjustments) added in #418 with the new Azure .NET SDK [version 1.2.1](https://github.com/Azure/azure-sdk-for-net/releases/tag/Azure.Data.AppConfiguration_1.2.1).

### DIFF
--- a/tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
+++ b/tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
@@ -185,6 +185,21 @@ namespace Tests.AzureAppConfiguration
                 eTag: new ETag("0a76e3d7-7ec1-4e37-883c-9ea6d0d89e63"),
                 contentType: "text");
 
+        readonly ConfigurationSetting Feature_RequirementTypeAll = ConfigurationModelFactory.ConfigurationSetting(
+            key: FeatureManagementConstants.FeatureFlagMarker + "Feature_All",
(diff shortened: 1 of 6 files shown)


tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -185,6 +185,21 @@ public class FeatureManagementTests
...
                 contentType: "text");
 
+        readonly ConfigurationSetting Feature_RequirementTypeAll = ConfigurationModelFactory.ConfigurationSetting(
+            key: FeatureManagementConstants.FeatureFlagMarker + "Feature_All",
+                value: @"
+                        {
+                          ""id"": ""Feature_All"",
+                          ""enabled"": true,
+                          ""conditions"": {
+                            ""requirement_type"": ""All"",
```
avanigupta, line 195: Is `requirement_type` a property of `conditions`?

amerjusupovic: We originally had `requirement_type` as part of `conditions` in #418 but it wasn't being returned by the SDK if it was in `conditions` so I moved it. This variable isn't used anywhere so I didn't catch that I forgot to change this, thanks.

avanigupta (approved):

PR #461, opened 2023-09-14
Ensure setting is processed by one adapter at most

Eskibear:
Given an important detail about adapters is that only one adapter is allowed to process each setting, i.e. we don't want the same configuration setting to be processed by two different adapters.

This PR breaks the loop ensuring at most one adapter successfully processes the setting.

We haven't noticed this issue because there are no two adapters with overlap `CanProcess` conditions so far.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -756,19 +756,21 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             foreach (IKeyValueAdapter adapter in _options.Adapters)
             {
-                if (!adapter.CanProcess(setting))
+                if (adapter.CanProcess(setting))
                 {
-                    continue;
+                    IEnumerable<KeyValuePair<string, string>> kvs = await adapter.ProcessKeyValue(setting, _logger, cancellationToken).ConfigureAwait(false);
+
+                    if (kvs != null)
+                    {
+                        keyValues = keyValues ?? new List<KeyValuePair<string, string>>();
+
+                        keyValues.AddRange(kvs);
+                    }
+
+                    // Ensure a setting is processed by one adpater at most.
+                    break;
                 }
 
-                IEnumerable<KeyValuePair<string, string>> kvs = await adapter.ProcessKeyValue(setting, _logger, cancellationToken).ConfigureAwait(false);
-
-                if (kvs != null)
-                {
-                    keyValues = keyValues ?? new List<KeyValuePair<string, string>>();
-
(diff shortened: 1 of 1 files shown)


avanigupta (approved):
This looks good to me.
@jimmyca15 was there a reason why we didn't break after a setting had been processed by an adapter? Just want to make sure I'm not missing any scenarios where we might want more than one adapter to process a setting.

jimmyca15:
> Given an important detail about adapters is that only one adapter is allowed to process each setting, i.e. we don't want the same configuration setting to be processed by two different adapters.

I'm not sure where this comes from. That was not a part of the design. If an adapter can adapt it should.

avanigupta:
@jimmyca15 if we do allow one setting to be processed by multiple adapters, wouldn't that lead to indeterministic key-values in `IConfiguration`? Depending on how the keys are flattened, we could end up with some overwritten keys and a mix of values coming from multiple adapters. 
Since the concept of adapters is our internal implementation, do we see any scenario where one setting might need to be processed by multiple adapters?

jimmyca15:
> if we do allow one setting to be processed by multiple adapters, wouldn't that lead to indeterministic key-values in IConfiguration? Depending on how the keys are flattened, we could end up with some overwritten keys and a mix of values coming from multiple adapters.

Depends on if the adapters place the same keys.

I just don't see a problem with the current code.

avanigupta:
> I just don't see a problem with the current code.

There's no problem with the current code because all adapters process different type of key-values. So, no two adapters will ever return `true` for `CanProcess` on the same setting. So it feels like we are implicitly enforcing that only one adapter can process a setting. If we add another adapter in future which can process a setting that's also processed by one of the existing adapters, I'm concerned about the resulting configuration being indeterministic. Hence the proposal to explicitly break after any adapter has processed a setting.

jimmyca15:
> If we add another adapter in future which can process a setting that's also processed by one of the existing adapters, I'm concerned about the resulting configuration being indeterministic.

Always hard to solve for the unknown future. If we did introduce multiple adapters that could process the same setting in the future and we take the change to run a single matched adapter now adapter ordering is important. We have to define the ordering and perhaps a list of adapters doesn't even make sense anymore. I think if we hit that problem we might decide to redesign since it's internal anyways.

Eskibear:
I was thinking about applying multiple adapters (like a pipeline) during implementing JS provider, as we don't need the flattening in JS.

a) At most one adapter wins:
- The list of adapters implies "available/registered adapters", but the order can be a problem. The behavior would depend on the order, which is very **implicit** and might be easier to cause issues in the future.

b) Any adapter applies if it can:
- We have to make sure adapters don't overlap with each other, just like the JSON adapter we have today, excluding feature management and key vault adapters.

But yes, in either way, it's internal, as long as we provide the expected results to customers. Let's just keep it as it is today. 

I'm closing this PR, and thanks for all the information!

PR #464, opened 2023-09-19
Add data collection notice

zhenlan:


avanigupta (approved):

zhenlan:
@avanigupta do we still have the "AppConfiguration-DotnetProvider-Windows-Buddy-main" (see it under "show all checks")? I can remove it from the branch policy. It's blocking from merging the PR.

Eskibear:
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/3e4de83d71b12fcb952cd20d19712dd232522ced/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L115-L115

> You may turn off the telemetry by setting the environment variable `AZURE_APP_CONFIGURATION_TRACING_DISABLED` to `True` or `1`

For "True". I'm thinking what the result will be if the env is set to `true` (lowercase). According to the impl, as far as I understand it should be the same as `True`, because it's eventually parsed by `bool.TryParse`. Anyway, it doesn't matter much as it's impossible/unnecessary for us to list all possible strings that can be parsed as True.

For "1". Learning from the [examples of system.boolean.tryparse](https://learn.microsoft.com/en-us/dotnet/api/system.boolean.tryparse?view=net-7.0#system-boolean-tryparse(system-string-system-boolean@)), **I'm wondering whether setting it to `1` can really disable the request tracing.** Because environment variable is of string type, and the string "1" should fail the parsing, thus enables the request tracing. Maybe it's a stupid question as I lack expertise on c#, please let me know then.

amerjusupovic:
> do we still have the "AppConfiguration-DotnetProvider-Windows-Buddy-main" (see it under "show all checks")? If not, I can remove it from the branch policy. It's blocking from merging the PR.

@zhenlan I'm trying to fix some issues since the automatic triggers seemed to stop working on some PRs. I ran the pipeline for this branch for now.

avanigupta:
> https://github.com/Azure/AppConfiguration-DotnetProvider/blob/3e4de83d71b12fcb952cd20d19712dd232522ced/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L115-L115
> 
> > You may turn off the telemetry by setting the environment variable `AZURE_APP_CONFIGURATION_TRACING_DISABLED` to `True` or `1`
> 
> For "True". I'm thinking what the result will be if the env is set to `true` (lowercase). According to the impl, as far as I understand it should be the same as `True`, because it's eventually parsed by `bool.TryParse`. Anyway, it doesn't matter much as it's impossible/unnecessary for us to list all possible strings that can be parsed as True.
> 
> For "1". Learning from the [examples of system.boolean.tryparse](https://learn.microsoft.com/en-us/dotnet/api/system.boolean.tryparse?view=net-7.0#system-boolean-tryparse(system-string-system-boolean@)), **I'm wondering whether setting it to `1` can really disable the request tracing.** Because environment variable is of string type, and the string "1" should fail the parsing, thus enables the request tracing. Maybe it's a stupid question as I lack expertise on c#, please let me know then.

That is a good catch, request tracing will not be disabled of the environment variable is set to "1". It only works for true/false (case-insensitive).

avanigupta:
I think we should fix the code to support "0/1" values for disabling telemetry. Azure SDK also does the same ([code](https://github.com/Azure/azure-sdk-for-net/blob/d5ee8a0ff55293afa6252ece2395fb08b75815ba/sdk/core/Azure.Core/src/DiagnosticsOptions.cs#L140-L155)).

zhenlan:
Thanks for the feedback. I updated the text to just `TRUE`.

avanigupta:
@zhenlan do you think we need to support "0/1" values as well if SDK supports this?

zhenlan:
I think it's fine not to. We can't go back to fix released versions anyway.

avanigupta (approved):

PR #466, opened 2023-09-22
Fix user agent header test

amerjusupovic:
Fix test to match the naming convention used for 7.0.0-preview.2

### DIFF
--- a/tests/Tests.AzureAppConfiguration/Tests.cs
+++ b/tests/Tests.AzureAppConfiguration/Tests.cs
@@ -186,7 +186,7 @@ namespace Tests.AzureAppConfiguration
             // 4. Contains the name and version of the App Configuration SDK package
             // 5. Contains the runtime information (target framework, OS description etc.) in the format set by the SDK
             // 6. Does not contain any additional components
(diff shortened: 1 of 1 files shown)


tests/Tests.AzureAppConfiguration/Tests.cs
```diff
@@ -186,7 +186,7 @@ public void TestUserAgentHeader()
             // 4. Contains the name and version of the App Configuration SDK package
             // 5. Contains the runtime information (target framework, OS description etc.) in the format set by the SDK
             // 6. Does not contain any additional components
-            string userAgentRegex = @"^Microsoft\.Extensions\.Configuration\.AzureAppConfiguration/\d+\.\d+\.\d+(-preview)?(-\d+-\d+)?,azsdk-net-Data.AppConfiguration/[.+\w-]+ \([.;\w\s]+\)$";
+            string userAgentRegex = @"^Microsoft\.Extensions\.Configuration\.AzureAppConfiguration/\d+\.\d+\.\d+(-preview)?(\.\d+)?,azsdk-net-Data.AppConfiguration/[.+\w-]+ \([.;\w\s]+\)$";
```
avanigupta, line 189: does this regex pass the test if version is "7.0.0.3"?

amerjusupovic: I didn't realize that, but I moved the check to inside the check for `-preview` so should fix that

avanigupta (approved):

PR #467, opened 2023-09-22
Improve summaries for methods in AzureAppConfigurationOptions

amerjusupovic:
Add mention of default query in the summary for `AzureAppConfigurationOptions`. Fix tags in summary for `AzureAppConfigurationOptions.UseFeatureFlags`.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -15,7 +15,8 @@ using System.Threading.Tasks;
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     /// <summary>
-    /// Options used to configure the behavior of an Azure App Configuration provider.
+    /// Options used to configure the behavior of an Azure App Configuration provider.         
+    /// If neither <see cref="Select"/> nor <see cref="SelectSnapshot"/> is ever called, all key-values with null label are included in the configuration provider.
     /// </summary>
     public class AzureAppConfigurationOptions
     {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -118,6 +118,7 @@ internal IEnumerable<IKeyValueAdapter> Adapters
         /// <summary>
         /// Specify what key-values to include in the configuration provider.
         /// <see cref="Select"/> can be called multiple times to include multiple sets of key-values.
+        /// If neither <see cref="Select"/> nor <see cref="SelectSnapshot"/> is ever called, all key-values with null label are included in the configuration provider.
```
jimmyca15, line 121: The intent seems good, but it's a bit odd that it's on the Select method. I guess I would expect it on AppConfigurationOptions summary.

amerjusupovic: Yea I agree honestly, I wasn't sure if adding it to the options summary seemed random/too much information there. Probably a bit better than this

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -15,7 +15,8 @@
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     /// <summary>
-    /// Options used to configure the behavior of an Azure App Configuration provider.
+    /// Options used to configure the behavior of an Azure App Configuration provider.         
+    /// If neither <see cref="Select"/> nor <see cref="SelectSnapshot"/> is ever called, all key-values with null label are included in the configuration provider.
```
jimmyca15, line 19:
```suggestion
    /// If neither <see cref="Select"/> nor <see cref="SelectSnapshot"/> is ever called, all key-values with no label are included in the configuration provider.
```

PR #469, opened 2023-10-03
- Add updated watchers

TheGoldenDragon:
Within our solution the following scenario occurred.

Our setup is as follows:
- We provider our appconfiguration values with terraform for multiple applications. 
- There is multiple pipelines writing values to appconfiguration.
- These pipelines can not share a single sentinel key, so they can't update it whenever new values are added or old ones have changed. 

While our application is running the app configurations can be updated. After this update our application will receive a service bus message that there is a new available setting. Based on this setting our application has to do some actions. 

We need to refresh the settings between the time the service bus call is made and the key is updated/added to appconfiguration.

In this PR following has been added to make it work:
- KeyValuePushNotification
The key value notifcation a model.
- ProcessKeyValuePushNotification
When this notification is processed it will mark all the keys that have one of the registered labels or key filters as expired. The next refresh will then reload those keys that have been updated.
- RegisterRefreshOnly was added to make the user aware it is using updates instead of the cache refresh mechanism

Example: 

Configuring:
`.AddAzureAppConfiguration(options => options
                    .Connect(connStr)
                    .Select(KeyFilter.Any)
                    .Select(KeyFilter.Any, "anylabelspecifictotheapplication")
                    .ConfigureRefresh(refresh => refresh
                        .RegisterUpdatesOnly()
                        .SetCacheExpiration(TimeSpan.FromDays(1))
                    )`
Receiving the eventgrid notification and processing it:
`if (eventGridEvent.TryCreateKeyValuePushNotification(out KeyValuePushNotification keyValuePushNotification))
{
    foreach (var refresher in _refreshers)
    {
        // Prompt Configuration Refresh based on the PushNotification
        refresher.ProcessKeyValuePushNotification(keyValuePushNotification, TimeSpan.FromSeconds(0));
    }
}`

Within your application you can now do:
`IConfigurationRefresher.RefreshAsync()`

This will now have reloaded the keys that have been updated.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -24,10 +24,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         private List<KeyValueWatcher> _changeWatchers = new List<KeyValueWatcher>();
         private List<KeyValueWatcher> _multiKeyWatchers = new List<KeyValueWatcher>();
-        private List<IKeyValueAdapter> _adapters = new List<IKeyValueAdapter>() 
-        { 
+        private List<IKeyValueAdapter> _adapters = new List<IKeyValueAdapter>()
+        {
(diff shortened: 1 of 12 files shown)


TheGoldenDragon:
@microsoft-github-policy-service agree company="embracesbs"

PR #470, opened 2023-10-11
Emit FeatureManagement versions in telemetry

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -75,6 +76,21 @@ public static bool IsDevEnvironment()
...
         }
 
+        public static string GetAssemblyVersion(string assemblyName)
+        {
+            try
+            {
+                if (!string.IsNullOrEmpty(assemblyName))
+                {
+                    // Return the version using only the first 3 fields and remove additional characters
+                    return Assembly.Load(assemblyName).GetName().Version.ToString(3).Trim('{', '}');
```
zhenlan, line 86: I suppose the assembly is loaded already and we don't have to load it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -24,13 +24,18 @@ internal class RequestTracingConstants
         public const string HostTypeKey = "Host";
         public const string FilterTypeKey = "Filter";
         public const string EnvironmentKey = "Env";
+        public const string FeatureManagementVersionKey = "FeatureManagementVersion";
```
zhenlan, line 27: This goes on the wire. Can we have something shorter, for example, FMver and FMANver?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -75,6 +75,21 @@ public static bool IsDevEnvironment()
...
+        {
+            try
+            {
+                if (!string.IsNullOrEmpty(assemblyName))
+                {
+                    // Return the version using only the first 3 fields and remove additional characters
+                    return AppDomain.CurrentDomain.GetAssemblies()?.SingleOrDefault(assembly => assembly.GetName().Name == assemblyName)?.GetName().Version.ToString(3).Trim('{', '}');
+                }
+            }
+            catch (Exception exception) when (exception is AppDomainUnloadedException) { }
```
jimmyca15, line 88:
```suggestion
            catch (AppDomainUnloadedException)
            {
                // Why is it ignored
            }
```

amerjusupovic: Missed this from old changes, thanks

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -24,13 +24,18 @@ internal class RequestTracingConstants
...
         public const string KeyVaultConfiguredTag = "UsesKeyVault";
         public const string KeyVaultRefreshConfiguredTag = "RefreshesKeyVault";
-        public const string ReplicaCount = "ReplicaCount";
+        public const string ReplicaCountKey = "ReplicaCount";
 
         public const string DiagnosticHeaderActivityName = "Azure.CustomDiagnosticHeaders";
         public const string CorrelationContextHeader = "Correlation-Context";
         public const string UserAgentHeader = "User-Agent";
+
+        public const string FeatureManagementPackageName = "Microsoft.FeatureManagement";
```
jimmyca15, line 38:
```suggestion
        public const string FeatureManagementAssemblyName = "Microsoft.FeatureManagement";
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -789,7 +805,9 @@ private void SetRequestTracingOptions()
                 IsKeyVaultConfigured = _options.IsKeyVaultConfigured,
                 IsKeyVaultRefreshConfigured = _options.IsKeyVaultRefreshConfigured,
                 ReplicaCount = _options.Endpoints?.Count() - 1 ?? _options.ConnectionStrings?.Count() - 1 ?? 0,
-                FilterTelemetry = _options.FeatureFilterTelemetry
+                FilterTelemetry = _options.FeatureFilterTelemetry,
+                FeatureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementPackageName),
```
jimmyca15, line 809: Given we know what limitations we're working with, I think we should just set this up in refresh. And add a comment of why.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
amerjusupovic: Now that I looked into it more, I'm not really sure if this exception should be ignored, since it signifies a rare/critical error and sounds like it would affect other parts of the application. Maybe it makes sense to leave this catch out?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -178,6 +179,16 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
             {
                 try
                 {
+                    // FeatureManagement assemblies may not be loaded on provider startup, so version information is gathered upon first refresh
+                    if (!_featureManagementVersionsTelemetrySet)
```
jimmyca15, line 183: I'm thinking of organizing this a bit differently so it doesn't interrupt the refresh flow as much.

a private method `EnsureTelemetry` and the flag is changed to `_isTelemetrySet`

So the addition here just becomes `EnsureTelemetry()`;

amerjusupovic: Pulling it out to the private method would definitely make it cleaner, but since we already have other telemetry (like feature filters) is it unclear to just call the method `EnsureTelemetry`?

jimmyca15: Doesn't seem unclear to me.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -178,6 +179,16 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                    {
+                        _requestTracingOptions.FeatureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
+
+                        _requestTracingOptions.FeatureManagementAspNetCoreVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAspNetCoreAssemblyName);
+
+                        _featureManagementVersionsTelemetrySet = true;
```
zhenlan, line 189: I don't think this is something we should attempt repeatedly. I would move this line before the GetAssemblyVersion call... but I'm still concerned this may skip the first refresh.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -26,6 +26,7 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
     {
         private bool _optional;
         private bool _isInitialLoadComplete = false;
+        private bool _featureManagementVersionsTelemetrySet;
```
zhenlan, line 29: How about `_isFeatureManagementVersionInspected`?

amerjusupovic: I think that sounds fine, but @jimmyca15 would you still prefer `_isTelemetrySet` once I pull the logic out to an `EnsureTelemtry` method?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Why would this be attempted repeatedly? From what I understand, `GetAssemblyVersion` wouldn't prevent the code from reaching the line where we set the boolean flag.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: We should use "tracing" instead of "telemetry". While I was thinking about suggesting that, I figured we could avoid it altogether :)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: It will never throw?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -971,5 +974,18 @@ private IEnumerable<ConfigurationSetting> GetCurrentKeyValueCollection(string ke
 
             return currentKeyValues;
         }
+
+        private void EnsureTracing()
```
zhenlan, line 978: Why do we give it such a generic name while it's all about feature management version detection?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Name update sounds good.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -971,5 +974,18 @@ private IEnumerable<ConfigurationSetting> GetCurrentKeyValueCollection(string ke
...
+        {
+            if (!_isTracingSet)
+            {
+                // FeatureManagement assemblies may not be loaded on provider startup, so version information is gathered upon first refresh for telemetry
+                _requestTracingOptions.FeatureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
+
+                _requestTracingOptions.FeatureManagementAspNetCoreVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAspNetCoreAssemblyName);
+
+                _isTracingSet = true;
```
jimmyca15, line 987: Do you mind setting this as the first step? Usually a good idea for once only behavior.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: It can throw some rare exceptions like `AppDomainUnloadedException` but I think it makes sense to move the flag setting to the top anyway now

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Ok that's true, I'll fix this to match the variable name you suggested as well

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -178,6 +179,9 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
             {
                 try
                 {
+                    // FeatureManagement assemblies may not be loaded on provider startup, so version information is gathered upon first refresh for telemetry
+                    EnsureFeatureManagementVersionInspected();
```
zhenlan, line 183: ... or `InspectFeatureManagementVersion`? Things can happen. We don't really "ensure" anything.

zhenlan: "EnsureInitialized" does ensure initialization, so it's a bit different, but I see your point. In similar cases like ours, people sometimes name the API like DoSomething**Once**. This is a private API, so I'm okay with it as is. It's more about code readability for someone new in the future :)

jimmyca15 (approved):

PR #488, opened 2023-11-02
Add StartupOptions for time-based retries on startup

amerjusupovic:
### StartupOptions and the Timeout property

This PR adds the new `Startup` property to `AzureAppConfigurationOptions`. You can set the `Startup.Timeout` property equal to a `TimeSpan` that represents how long the provider should continue retrying the store and replicas passed to the provider before failing.

### Example usage:

```c#
config.AddAzureAppConfiguration(options =>
{
   // . . .
    options.ConfigureStartupOptions(startupOptions =>
    {
        startupOptions.Timeout = TimeSpan.FromMinutes(5);
    });
});
```


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -114,6 +114,11 @@ internal IEnumerable<IKeyValueAdapter> Adapters
         /// </summary>
         internal FeatureFilterTracing FeatureFilterTracing { get; set; } = new FeatureFilterTracing();
 
+        /// <summary>
+        /// Options used to configure provider startup.
+        /// </summary>
+        internal StartupOptions Startup { get; set; } = new StartupOptions();
```
avanigupta, line 120: public?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -79,7 +83,8 @@ public void UpdateClientStatus(ConfigurationClient client, bool successful)
             else
             {
                 clientWrapper.FailedAttempts++;
-                TimeSpan backoffDuration = FailOverConstants.MinBackoffDuration.CalculateBackoffDuration(FailOverConstants.MaxBackoffDuration, clientWrapper.FailedAttempts);
+                TimeSpan minBackoffDuration = _isStartup ? TimeSpan.FromSeconds(1) : FailOverConstants.MinBackoffDuration;
```
avanigupta, line 86: Can you make this timespan a constant?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     public class AzureAppConfigurationOptions
     {
         private const int MaxRetries = 2;
-        private static readonly TimeSpan MaxRetryDelay = TimeSpan.FromMinutes(1);
+        internal static readonly TimeSpan MaxRetryDelay = TimeSpan.FromMinutes(1);
```
avanigupta, line 23: Is this change necessary?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: Not sure when I changed it to internal, will change it back

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: Referenced it in `StartupOptions` to try and keep this value consistent for both files, but I could just use a constant there too.

avanigupta: Do we want to support a different MaxRetryDelay for startup vs refresh?
@jimmyca15

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -28,7 +28,8 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
         private bool _isInitialLoadComplete = false;
         private bool _isFeatureManagementVersionInspected;
         private readonly bool _requestTracingEnabled;
-        private readonly IConfigurationClientManager _configClientManager;
+        private readonly IConfigurationClientManager _refreshConfigClientManager;
+        private readonly IConfigurationClientManager _startupConfigClientManager;
```
avanigupta, line 32: We might not need two ConfigurationClientManagers since the only configurable startup option should be the overall timeout.

zhenlan:
>                     if (_mappedData == null)

Is this only triggered during refresh? If so, it means the startup is already done (though initial configuration load had failed). We shouldn't use the startup logic here, because if anything fails here it will be tried again in the next refresh.

---
Refers to: src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs:211 in c8ada35. [](commit_id = c8ada35504eb98d64cbb06ed008a2354f58eadf7, deletion_comment = False)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -554,25 +563,84 @@ private async Task InitializeAsync(bool ignoreFailures, IEnumerable<Configuratio
...
-                            client,
-                            data,
-                            cancellationToken)
-                            .ConfigureAwait(false);
-
-                        watchedSettings = UpdateWatchedKeyValueCollections(watchedSettings, data);
-                    },
-                    cancellationToken)
-                    .ConfigureAwait(false);
+                        await Task.Delay(1000).ConfigureAwait(false);
```
zhenlan, line 639: Can we make this 1s a const? This is the minimum delay between retries. Not sure if we should have a maximal delay and slowly grow the delay between retries. Thoughts?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: I'm thinking the same. Essentially, the startup is like triggering the refresh nonstop (well, 1s interval) until the allocated time runs out. It's the same clients for startup or refresh.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/StartupOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// The maximum delay before timing out when loading data from Azure App Configuration on startup.
+        /// </summary>
+        public TimeSpan Timeout { get; set; } = TimeSpan.FromMinutes(1);
+
+        /// <summary>
+        /// The maximum delay between retries when loading data from Azure App Configuration on startup.
+        /// </summary>
+        public TimeSpan MaxRetryDelay { get; set; } = AzureAppConfigurationOptions.MaxRetryDelay;
```
zhenlan, line 22: Is this to ignore what a user may set in ClientOptions? I don't think we need it. If a user set something, we can just honor it even if it may mess up the startup timeout.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -114,6 +114,11 @@ internal IEnumerable<IKeyValueAdapter> Adapters
         /// </summary>
         internal FeatureFilterTracing FeatureFilterTracing { get; set; } = new FeatureFilterTracing();
 
+        /// <summary>
+        /// Options used to configure provider startup.
+        /// </summary>
+        public StartupOptions Startup { get; set; } = new StartupOptions();
```
jimmyca15, line 120: I notice that ClientOptions is internal and we have `ConfigureClientOptions`. Seems we should stick with that pattern given we have established it?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/StartupOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Options used when initially loading data into the configuration provider.
+    /// </summary>
+    public class StartupOptions
+    {
+        /// <summary>
+        /// The maximum delay before timing out when loading data from Azure App Configuration on startup.
+        /// </summary>
+        public TimeSpan Timeout { get; set; } = TimeSpan.FromMinutes(1);
```
jimmyca15, line 17: Default timeout of one minute seems a bit low. I'd go with 10 minutes.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
jimmyca15: Looks like my second comment got swallowed. Although I expect we would use the same time, I don't think this should be shared in that manner. A separate field to reference where needed seems appropriate.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/StartupOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    /// <summary>
+    /// Options used when initially loading data into the configuration provider.
+    /// </summary>
+    public class StartupOptions
+    {
+        /// <summary>
+        /// The maximum delay before timing out when loading data from Azure App Configuration on startup.
```
jimmyca15, line 15:
```suggestion
        /// The amount of time allowed to load data from Azure App Configuration on startup.
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -131,13 +133,15 @@ public override void Load()
             var loadStartTime = DateTimeOffset.UtcNow;
 
             // Guaranteed to have atleast one available client since it is a application startup path.
-            IEnumerable<ConfigurationClient> availableClients = _configClientManager.GetAvailableClients(loadStartTime);
+            IEnumerable<ConfigurationClient> availableClients = _startupConfigClientManager.GetAvailableClients(loadStartTime);
 
             try
             {
+                CancellationTokenSource startupCancellationTokenSource = new CancellationTokenSource(_options.Startup.Timeout);
```
jimmyca15, line 140: disposable

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: `ConfigureStartupOptions` sounds like a good idea

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: The actual server calls are already retried with some exponential backoff strategy. I dont see why we would need to slow down the overall startup retries.

avanigupta: Unless the concern is that this can be misused by customers if they set SDK retries to 0. If SDK is not retrying, we'll keep retrying every 1 second which is not good. So having this second layer of retries with exponential backoff might be good.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Since it looks like we want to remove the startup MaxRetryDelay then I think there's no reason for the 2 managers

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: Imagining there is only one client, we don't want to beat that client in a tight loop.

zhenlan, line 639: >Task.Delay(1000) [](http://example.com/codeflow?start=30&length=16)

Should we pass the cancelationTokey to Task.Delay?

jimmyca15: I think Amer will be reworking this a bit. I agree that we should backoff on the top level. I suggested a flow like below.

```
public async Task LoadAsync()
{
    var cts = new CancellationTokenSource(_startupOptions.Timeout);

    var errors = new List<Exception>();

    int attempts = 0;

    while (true)
    {
        attempts++;

        IEnumerable<ConfigurationClient> clients = _clientManager.GetClients(ignoreBackoff: true);

        if (await TryInitializeAsync(clients, errors, cts.Token))
        {
            break;
        }

        try
        {
            await Task.Delay(Timespan.FromSeconds(1).Backoff(attempts), cts.Token);
        }
        catch (OperationCanceledException)
        {
            throw new TimeoutException(
                new AggregateException(errors));
        }
    }
}

//
// TryInitializeAsync
// It swallows (returns false for) operation canceled exception
// It swallows failoverable exceptions
// It propagates non-failoverable exceptions
// It fills the 'errors' list with swallowed exceptions
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationClientManager.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     internal interface IConfigurationClientManager
     {
-        IEnumerable<ConfigurationClient> GetAvailableClients(DateTimeOffset time);
+        IEnumerable<ConfigurationClient> GetAvailableClients(DateTimeOffset time, bool ignoreBackoff = false);
```
jimmyca15, line 13: It doesn't make sense to pass the time and the ignore backoff parameter.

Perhaps a separate method is warranted. GetClients, which returns all clients.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -128,16 +130,13 @@ public override void Load()
...
             var watch = Stopwatch.StartNew();
 
-            var loadStartTime = DateTimeOffset.UtcNow;
-
-            // Guaranteed to have atleast one available client since it is a application startup path.
-            IEnumerable<ConfigurationClient> availableClients = _configClientManager.GetAvailableClients(loadStartTime);
-
             try
             {
+                using CancellationTokenSource startupCancellationTokenSource = new CancellationTokenSource(_options.Startup.Timeout);
```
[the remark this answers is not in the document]
jimmyca15: Line is a bit long, var can help since we have type on the right side.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -209,7 +208,8 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                         if (InitializationCacheExpires < utcNow)
                         {
                             InitializationCacheExpires = utcNow.Add(MinCacheExpirationInterval);
-                            await InitializeAsync(ignoreFailures: false, availableClients, cancellationToken).ConfigureAwait(false);
+
+                            await InitializeAsync(_configClientManager.GetAvailableClients(utcNow), cancellationToken).ConfigureAwait(false);
```
jimmyca15, line 212: Why not use `availableClients` defined above?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +547,116 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                if (IsFailOverable(exception))
+                {
+                    startupExceptions.Add(exception);
+
+                    return false;
+                }
+
+                throw;
+            }
+            catch (Exception exception) when ((exception as AggregateException)?.InnerExceptions?.Any(e =>
```
jimmyca15, line 620: catch aggregate exception

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +547,116 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                e is RequestFailedException) ?? false)
+            {
+                if (IsFailOverable(exception as AggregateException))
```
jimmyca15, line 623: as implies it can be null. If it's known to be a certain type, use a cast instead.

amerjusupovic: Makes sense, for now I think changing it to catch `AggregateException` as you mentioned should fix this too

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +547,116 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                        break;
+                    }
+
+                    try
+                    {
+                        await Task.Delay(FailOverConstants.MinStartupBackoffDuration.CalculateBackoffDuration(FailOverConstants.MaxBackoffDuration, attempts), cancellationToken).ConfigureAwait(false);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        throw new TimeoutException($"The provider timed out while attempting to load data on startup.",
```
jimmyca15, line 575:
```suggestion
                        throw new TimeoutException($"The provider timed out while attempting to load.",
```

I would leave out "on startup", technically nothing prevents someone from instantiating our provider during the lifetime of the application.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +547,116 @@ private void SetDirty(TimeSpan? maxDelay)
             return applicationData;
         }
 
-        private async Task InitializeAsync(bool ignoreFailures, IEnumerable<ConfigurationClient> availableClients, CancellationToken cancellationToken = default)
+        public async Task LoadAsync(bool ignoreFailures, CancellationToken cancellationToken)
```
jimmyca15, line 550: why is this public?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +547,116 @@ private void SetDirty(TimeSpan? maxDelay)
...
+        }
+
+        private async Task<bool> TryInitializeAsync(IEnumerable<ConfigurationClient> clients, List<Exception> startupExceptions, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                await InitializeAsync(clients, cancellationToken).ConfigureAwait(false);
+            }
+            catch (Exception exception) when (exception is OperationCanceledException ||
+                ((exception as AggregateException)?.InnerExceptions?.Any(e =>
```
jimmyca15, line 599: aggregate exception is handled below. Seems this should just be catch operationcanceledexception.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: My mistake, will fix that

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: We've had customer issues where the SDK returns an `AggregateException` that contains only `OperationCanceledException`, so I wanted to check for that in case it was thrown because of our startup cancellation token. I thought it might avoid throwing an `AggregateException` when it was actually cancelled by the timeout and should be a `TimeoutException`.

jimmyca15: That's fine, but should be handled in the catch block for aggregate exception.

jimmyca15:
pr example usage needs an update.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -407,6 +412,16 @@ public AzureAppConfigurationOptions Map(Func<ConfigurationSetting, ValueTask<Con
             return this;
         }
 
+        /// <summary>
+        /// Configure the client(s) used to communicate with Azure App Configuration during startup.
```
jimmyca15, line 416: that summary wrong?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -28,26 +28,26 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
             try
             {
                 AzureAppConfigurationOptions options = _optionsProvider();
-                IConfigurationClientManager clientManager;
+                IConfigurationClientManager refreshClientManager;
```
jimmyca15, line 32: The changes in this file seem unneeded.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -3,12 +3,14 @@
 //
 using Azure;
 using Azure.Data.AppConfiguration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.Constants;
```
jimmyca15, line 6: We shouldn't have a Constants namespace.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +546,124 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                {
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        return false;
+                    }
+                }
+
+                if (exception.InnerExceptions?.Any(e => e is RequestFailedException) ?? false)
+                {
+                    if (IsFailOverable(exception))
```
jimmyca15, line 629: the ifs can be combined with &&

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +546,124 @@ private void SetDirty(TimeSpan? maxDelay)
...
+            {
+                if (exception.InnerExceptions?.Any(e => e is OperationCanceledException) ?? false)
+                {
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        return false;
+                    }
+                }
+
+                if (exception.InnerExceptions?.Any(e => e is RequestFailedException) ?? false)
```
jimmyca15, line 627:
```suggestion
                if (exception.InnerExceptions != null &&
                    exception.InnerExceptions.Any(e => e is RequestFailedException))
```

A suggestion, I feel the ?? false is a bit weird to read.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +546,118 @@ private void SetDirty(TimeSpan? maxDelay)
...
+            }
+            catch (OperationCanceledException)
+            {
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    return false;
+                }
+
+                throw;
+            }
```
avanigupta, line 605:
```suggestion
            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
            {
                return false;
            }
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +546,118 @@ private void SetDirty(TimeSpan? maxDelay)
...
             try
             {
-                await ExecuteWithFailOverPolicyAsync(
-                    availableClients,
-                    async (client) =>
+                while (true)
+                {
+                    attempts++;
+
+                    IEnumerable<ConfigurationClient> clients = _configClientManager.GetClients();
```
avanigupta, line 561: We can call `GetClients` outside the while loop

tests/Tests.AzureAppConfiguration/ClientOptionsTests.cs
```diff
@@ -19,40 +16,42 @@ public void ClientOptionsTests_OverridesDefaultClientOptions()
         {
             // Arrange
             var requestCountPolicy = new HttpRequestCountPipelinePolicy();
-            int defaultMaxRetries = 0;
```
avanigupta, line 22: Could you add a new test? The current one we have is still valuable.

tests/Tests.AzureAppConfiguration/FailoverTests.cs
```diff
@@ -123,7 +127,7 @@ public void FailOverTests_ReturnsAllClientsIfAllBackedOff()
                 });
 
             // Throws last exception when all clients fail.
-            Assert.Throws<RequestFailedException>(configBuilder.Build);
+            Assert.Throws<TimeoutException>(configBuilder.Build);
```
avanigupta, line 130: We can also assert if inner exceptions are as expected.

tests/Tests.AzureAppConfiguration/MockedConfigurationClientManager.cs
```diff
@@ -20,12 +20,17 @@ internal class MockedConfigurationClientManager : IConfigurationClientManager
 
         public MockedConfigurationClientManager(IEnumerable<ConfigurationClientWrapper> clients)
         {
-            this._clients = clients.ToList();
+            _clients = clients.ToList();
         }
 
         public IEnumerable<ConfigurationClient> GetAvailableClients(DateTimeOffset time)
```
avanigupta, line 26: `public IEnumerable<ConfigurationClient> GetAvailableClients(DateTimeOffset time) => GetClients();`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +546,113 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                    startupExceptions.Add(exception);
+
+                    return false;
+                }
+
+                throw;
+            }
+            catch (AggregateException exception)
+            {
+                if ((exception.InnerExceptions?.Any(e => e is OperationCanceledException) ?? false) && cancellationToken.IsCancellationRequested)
```
avanigupta, line 614: I think we still need to add the exception to startup exceptions if the operation was cancelled by something other than our startup cancellation token

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +546,113 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                {
+                    return false;
+                }
+
+                if ((exception.InnerExceptions?.Any(e => e is RequestFailedException) ?? false) && IsFailOverable(exception))
```
avanigupta, line 619: This can be simplified if we just check `IsFailOverable(exception)`. The first condition is not needed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Maybe something like this?

```
if (exception.InnerExceptions?.Any(e => e is OperationCanceledException) ?? false)
{
    if (!cancellationToken.IsCancellationRequested)
    {
        startupExceptions.Add(exception);
    }

    return false;
}
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationClientManager.cs
zhenlan: How about `GetAllClients()` to better differentiate from `GetAvailableClients`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +546,113 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                    return false;
+                }
+                
+                throw;
+            }
+
+            return true;
+        }
+
+        private async Task InitializeAsync(IEnumerable<ConfigurationClient> availableClients, CancellationToken cancellationToken = default)
```
zhenlan, line 632: NIT: can we rename `availableClients` to `clients`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/StartupOptions.cs
zhenlan: I'm a bit concerned that we may get complaints like [https://github.com/Azure/AppConfiguration/issues/255](https://github.com/Azure/AppConfiguration/issues/255) all over again if we set the default startup timeout to anything more than a minute.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +546,113 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                    attempts++;
+
+                    if (await TryInitializeAsync(clients, startupExceptions, cancellationToken).ConfigureAwait(false))
+                    {
+                        break;
+                    }
+
+                    try
+                    {
+                        await Task.Delay(FailOverConstants.MinStartupBackoffDuration.CalculateBackoffDuration(FailOverConstants.MaxBackoffDuration, attempts), cancellationToken).ConfigureAwait(false);
```
zhenlan, line 570: The current backoff logic uses a factor of 2. I'm afraid it's not effective enough for the startup scenario. Let's assume all appconfig requests fail immediately. If the startup timeout is 1 minute, we will retry 5 times at the second 1, 3, 7, 15 and 31. We almost won't do anything for the second half of the 1-minute timeout. Even if we extend the timeout to 10 minutes, we will only retry an additional 4 times (at the second 63, 127, 155, and 511). I'm wondering if we should pick a smaller backoff factor.

avanigupta: Exponential backoff strategy commonly indicates that we are using factors of 2. If we are concerned about not getting enough retries within 1 minute timeout, should we consider using a fixed backoff policy? We can more easily control how many times we want to retry on startup and we can always add jitter to avoid retrying at the same interval every time. For example, we can backoff for 5 seconds + jitter every time we fail to load on startup.

zhenlan: The downside of linear backoff is that we may beat the service too much if someone sets a very long startup timeout, for example, a day. The exponential backoff (y=f^x) helps balance the short-term vs. long-term problem. Here is a comparison of f=2 (blue) vs. f=1.5 (red). A smaller f will retry more times for the short-term but will still increase the interval for the longer term.
image

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: Do we expect the clients to change between retries, for example, with replica auto-discovery?

avanigupta: Oh that's a good point, it could change.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: I didn't think of the longer timeouts. Backoff with 1.5 factor sounds good in that case.

zhenlan: I used 1.5 as an example. Amer can help to see which works best for our scenarios.

amerjusupovic: Good question, I can try and see what looks best with a 1 minute timeout by default, does that sound good?

tests/Tests.AzureAppConfiguration/ClientOptionsTests.cs
amerjusupovic: I'm not sure exactly how to keep this test since the number of requests during startup depends on the time elapsed, which is why I changed it. I'm not sure if it would consistently give the same number of requests given the same timeout.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhenlan: Yes, that will be great. Thanks, Amer.

amerjusupovic: Overall, I think the takeaway is that the time isn't perfectly predictable. There are a lot of variables that can be different: configuration client retries can delay up to ~5.4 seconds per call with default options, the `Random.Double` value in the backoff calculation, or how long a request to app config takes to complete. Assuming retries have 6 seconds delay and an average of 0.5 for `Random.Double`, a backoff factor of around 1.9 would fit 5 retries in 1 minute with time to spare and a factor of 1.5 would fit in 6 retries fairly well.

I think it makes sense to change the backoff to 1.5 (for startup specifically) to fit an extra retry and have less wasted time on average in the first minute. That's what I gathered from running an application locally and checking the times for each request, let me know if I missed anything.

zhenlan: Thanks for investigating this, Amer. Sounds good to me.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +545,130 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                    {
+                        if (startupStopwatch.Elapsed < FailOverConstants.StartupFixedBackoffDuration)
+                        {
+                            await Task.Delay(startupStopwatch.Elapsed.CalculateFixedStartupBackoffDuration(), cancellationToken).ConfigureAwait(false);
+                        }
+                        else
+                        {
+                            exponentialBackoffAttempts++;
+
+                            await Task.Delay(FailOverConstants.MinBackoffDuration.CalculateExponentialStartupBackoffDuration(
```
jimmyca15, line 577: I'd like to see the delay calculated outside the try, and a single place where we invoke Task.Delay

jimmyca15:
@amerjusupovic will you be updating the target branch to preview?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -547,41 +545,130 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                    if (await TryInitializeAsync(clients, startupExceptions, cancellationToken).ConfigureAwait(false))
+                    {
+                        break;
+                    }
+
+                    try
+                    {
+                        if (startupStopwatch.Elapsed < FailOverConstants.StartupFixedBackoffDuration)
+                        {
+                            await Task.Delay(startupStopwatch.Elapsed.CalculateFixedStartupBackoffDuration(), cancellationToken).ConfigureAwait(false);
```
jimmyca15, line 571: the timespan you are passing to the method has a different value than the one you have checked in the condition.

jimmyca15: To avoid the problem, you could cache elapsed in a variable, but I'd recommend an approach like

```
TimeSpan delay;
if (startupStopwatch.TryGetFixedBackoff(out Timespan backoff))
{
    delay = backoff;
}
else
{
    postFixedWindowAttempts++;

    delay = FailoverConstants.MaxFixedStartupBackoff.CalculateBackoff(
        attempts: postFixedWindowAttempts,
        maxDuration: FailOverConstants.MaxBackoffDuration
    );
}

await Task.Delay(delay, ct);

```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -72,5 +80,98 @@ public static TimeSpan CalculateBackoffDuration(this TimeSpan minDuration, TimeS
...
+
+            var rand = new Random();
+
+            double interval = Math.Abs(timeSpan.TotalMilliseconds * ratio);
+
+            double lowest = timeSpan.TotalMilliseconds - interval * 0.5;
+
+            double offset = interval * rand.NextDouble();
+
+            return TimeSpan.FromMilliseconds(lowest + offset);
```
zhenlan, line 170: For a ratio of 0.5, do we expect a jitted time of 10s falls in [5, 15]? If so, above implementation is incorrect as it generates the range of [7.5, 12.5]. I hope something like below is easier to understand:

```cs
double jitter = ratio * (rand.NextDouble() * 2 - 1);
double interval = timeSpan.TotalMilliseconds * (1 + jitter);
return TimeSpan.FromMilliseconds(interval);
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -548,42 +546,135 @@ private void SetDirty(TimeSpan? maxDelay)
...
-            Dictionary<string, ConfigurationSetting> data = null;
-            Dictionary<KeyValueIdentifier, ConfigurationSetting> watchedSettings = null;
-            
+            var startupStopwatch = Stopwatch.StartNew();
+
+            int postFixedWindowAttempts = 0;
+
+            var startupExceptions = new List<Exception>();
+
+            IEnumerable<ConfigurationClient> clients = _configClientManager.GetAllClients();
```
zhenlan, line 557: Move this into the loop so you get new clients when replica auto discovery is enabled.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Didn't see this initially, will move it out

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
jimmyca15: jittering a 10s duration by a ratio of .5 introduces a 5 second (10 x .5) window. So the expectation is [7.5, 12.5]

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/FailOverConstants.cs
```diff
@@ -4,12 +4,14 @@
...
-namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Constants
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     internal class FailOverConstants
     {
         // Timeouts to retry requests to config stores and their replicas after failure.
         public static readonly TimeSpan MinBackoffDuration = TimeSpan.FromSeconds(30);
         public static readonly TimeSpan MaxBackoffDuration = TimeSpan.FromMinutes(10);
+        public static readonly TimeSpan StartupFixedBackoffDuration = TimeSpan.FromMinutes(10);
+        public static readonly TimeSpan MaxFixedStartupBackoff = TimeSpan.FromSeconds(30);
```
zhenlan, line 15: MinStartupBackoffDuration with comments

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -548,42 +546,135 @@ private void SetDirty(TimeSpan? maxDelay)
...
-                            client,
-                            data,
-                            cancellationToken)
-                            .ConfigureAwait(false);
-
-                        watchedSettings = UpdateWatchedKeyValueCollections(watchedSettings, data);
-                    },
-                    cancellationToken)
-                    .ConfigureAwait(false);
+                        TimeSpan delay;
```
jimmyca15, line 570: Only the Task.Delay should be in the try.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -548,42 +546,135 @@ private void SetDirty(TimeSpan? maxDelay)
...
+                                FailOverConstants.MaxBackoffDuration,
+                                postFixedWindowAttempts,
+                                startupStopwatch.Elapsed);
+                        }
+
+                        await Task.Delay(delay, cancellationToken).ConfigureAwait(false);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        throw new TimeoutException($"The provider timed out while attempting to load.",
```
jimmyca15, line 590:
```suggestion
                        throw new TimeoutException(
                            $"The provider timed out while attempting to load.",
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -72,5 +80,107 @@ public static TimeSpan CalculateBackoffDuration(this TimeSpan minDuration, TimeS
...
+            // IMPORTANT: This can overflow
+            double calculatedMilliseconds = Math.Max(1, minDuration.TotalMilliseconds) * ((long)1 << Math.Min(attempts, MaxAttempts));
+
+            if (calculatedMilliseconds > maxDuration.TotalMilliseconds ||
+                    calculatedMilliseconds <= 0 /*overflow*/)
+            {
+                calculatedMilliseconds = maxDuration.TotalMilliseconds;
+            }
+
+            TimeSpan maxBackoffDuration = CalculateMaxStartupBackoffDuration(startupTimeElapsed);
```
jimmyca15, line 118:
```suggestion
            TimeSpan maxBackoffDuration = FailOverConstants.MaxBackoffDuration;
```

jimmyca15: Actually, the variable shouldn't even be needed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -72,5 +80,107 @@ public static TimeSpan CalculateBackoffDuration(this TimeSpan minDuration, TimeS
...
+        /// </summary>
+        /// <param name="minDuration">The minimum duration to retry after.</param>
+        /// <param name="maxDuration">The maximum duration to retry after.</param>
+        /// <param name="attempts">The number of attempts made to the configuration store after the fixed retry period.</param>
+        /// <param name="startupTimeElapsed">The time elapsed since the current startup began.</param>
+        /// <returns>The backoff duration before retrying a request to the configuration store or replica again.</returns>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// An exception is thrown when <paramref name="attempts"/> is less than 1.
+        /// </exception>
+        public static TimeSpan CalculateExponentialStartupBackoffDuration(this TimeSpan minDuration, TimeSpan maxDuration, int attempts, TimeSpan startupTimeElapsed)
```
jimmyca15, line 96:
```suggestion
        public static TimeSpan CalculateExponentialStartupBackoffDuration(this TimeSpan duration, TimeSpan maxDuration, int attempts, TimeSpan startupTimeElapsed)
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
jimmyca15, line 118: The maxBackoffDuration variable shouldn't be needed. The max backoff in this case should always be our 10 min max backoff.

jimmyca15: What's the point of the maxDuration parameter then?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
jimmyca15: This method really isn't tied to startup in any way.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
zhenlan: The formula can be updated easily to achieve the desired results, but that's not intuitive to me. The 50% jitter doesn't give me 50% more (or less) than what I had. So if I want 100% more (or less) than what I had, I actually need a 200% jitter. Isn't that weird?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -72,5 +80,107 @@ public static TimeSpan CalculateBackoffDuration(this TimeSpan minDuration, TimeS
...
+            double interval = Math.Abs(timeSpan.TotalMilliseconds * ratio);
+
+            double lowest = timeSpan.TotalMilliseconds - interval * 0.5;
+
+            double offset = interval * rand.NextDouble();
+
+            return TimeSpan.FromMilliseconds(lowest + offset);
+        }
+
+        private static TimeSpan CalculateMaxStartupBackoffDuration(TimeSpan startupTimeElapsed)
```
jimmyca15, line 173: I really only see a need for this logic in TryGetFixedBackoff, and in that case the default return isn't needed since TryGetFixedBackoff would return false. I recommend moving it in there.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
amerjusupovic: I agree, I missed some things here. Zhenlan suggested just using `CalculateBackoffDuration` and adjusting that method to use jitter instead of the randomizing it does in the last line. At this point we're using the same backoff duration values, so unless there's a reason refresh is different from startup retries then I think it makes sense.

jimmyca15: A single method will be better. That's your suggestion right?

amerjusupovic: Right, just one method and removing the `CalculateExponentialStartupBackoffDuration` method

jimmyca15: Perfect, let's do it.

jimmyca15: Just update that implementation with this one should be good right?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
jimmyca15: Discussed separately with @zhenlan. Seems more people interpret the ratio here as the delta from the extreme end of the possible duration window to the original duration. In that case, we would benefit to change it.

So jitter(10, .5) yields [5, 15] rather than [7.5, 12.5]

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -1,14 +1,22 @@
...
-using Microsoft.Extensions.Configuration.AzureAppConfiguration.Constants;
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 {
     internal static class TimeSpanExtensions
     {
         private const int MaxAttempts = 63;
+        private const double JitterRatio = 0.5;
```
zhenlan, line 12: I believe @jimmyca15 prefers this to be 0.25.

zhenlan (approved):
:shipit:

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -948,18 +1038,24 @@ private bool IsFailOverable(AggregateException ex)
 
         private bool IsFailOverable(RequestFailedException rfe)
```
amerjusupovic, line 1039: For net framework 4.8 only, tests failed because the structure of the `RequestFailedException` returned was different, with the inner exceptions not being wrapped in an `HttpRequestException`. I've adjusted the `IsFailoverable` method to account for that. Wanted to call it out here. 

Actual:   typeof(System.AggregateException): Retry failed after 3 tries. Retry settings can be adjusted in ClientOptions.Retry or by configuring a custom retry policy in ClientOptions.RetryPolicy.
---- System.AggregateException : Retry failed after 3 tries. Retry settings can be adjusted in ClientOptions.Retry or by configuring a custom retry policy in ClientOptions.RetryPolicy.
-------- Azure.RequestFailedException : The remote name could not be resolved: 'azure.azconfig.io'
------------ System.Net.WebException : The remote name could not be resolved: 'azure.azconfig.io'
-------- Azure.RequestFailedException : The remote name could not be resolved: 'azure.azconfig.io'
------------ System.Net.WebException : The remote name could not be resolved: 'azure.azconfig.io'
-------- Azure.RequestFailedException : The remote name could not be resolved: 'azure.azconfig.io'
------------ System.Net.WebException : The remote name could not be resolved: 'azure.azconfig.io'

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -948,18 +1038,24 @@ private bool IsFailOverable(AggregateException ex)
...
-                       hre.InnerException is IOException;
-            }
-
-            return rfe.Status == HttpStatusCodes.TooManyRequests ||
+                return rfe.Status == HttpStatusCodes.TooManyRequests ||
                    rfe.Status == (int)HttpStatusCode.RequestTimeout ||
                    rfe.Status >= (int)HttpStatusCode.InternalServerError;
+            }
+
+            if (innerException is HttpRequestException hre && hre.InnerException != null)
```
jimmyca15, line 1050: Seems solid, but the multiassignment to innerException threw me off a bit. I think it would help if the assignment to inner exception only happened once because it's really behaving in a branched manner.

I'd suggest

```
            if (rfe.Status == HttpStatusCodes.TooManyRequests ||
                rfe.Status == (int)HttpStatusCode.RequestTimeout ||
                rfe.Status >= (int)HttpStatusCode.InternalServerError)
            {
                return true;
            }

            Exception innerException;

            if (rfe.InnerException is HttpRequestException hre)
            {
                innerException = hre.InnerException;
            }
            else
            {
                innerException = rfe.InnerException;
            }

            // The InnerException could be SocketException or WebException when an endpoint is invalid and IOException if it's a network issue.
            return innerException is WebException ||
                   innerException is SocketException ||
                   innerException is IOException;

```

jimmyca15 (approved):

jimmyca15 (approved):

PR #489, opened 2023-11-03
Validate domain of Srv endpoints

RichardChen820:
Validate domain of Srv endpoints @jimmyca15

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
@@ -36,9 +36,9 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private readonly TokenCredential _credential;
         private readonly ConfigurationClientOptions _clientOptions;
         private readonly bool _replicaDiscoveryEnabled;
+        private readonly SrvLookupClient _srvLookupClient;
 
         private IList<ConfigurationClientWrapper> _dynamicClients;
-        private SrvLookupClient _srvLookupClient;
         private IEnumerable<string> _validDomains;
         private DateTimeOffset _lastFallbackClientRefresh = default;
         private DateTimeOffset _lastFallbackClientRefreshAttempt = default;
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
@@ -351,7 +351,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             {
                 try
                 {
-                    _validDomains = await DomainVerifier.GetValidDomain(_endpoint, hostName).ConfigureAwait(false);
+                    _validDomains = await DomainVerifier.GetValidDomains(_endpoint, hostName).ConfigureAwait(false);
                 }
                 catch (SocketException)
                 {
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
@@ -36,7 +36,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private readonly bool _replicaDiscoveryEnabled;
 
         private IList<ConfigurationClientWrapper> _dynamicClients;
-        private Lazy<SrvLookupClient> _srvLookupClient;
+        private SrvLookupClient _srvLookupClient;
+        private string _validDomain = string.Empty;
         private DateTimeOffset _lastFallbackClientRefresh = default;
         private DateTimeOffset _lastFallbackClientRefreshAttempt = default;
         private Logger _logger = new Logger();
@@ -61,7 +62,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             _clientOptions = clientOptions;
             _replicaDiscoveryEnabled = replicaDiscoveryEnabled;
 
-            _srvLookupClient = new Lazy<SrvLookupClient>();
+            _srvLookupClient = new SrvLookupClient();
 
             _clients = connectionStrings
                 .Select(cs =>
@@ -93,7 +94,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             _clientOptions = clientOptions;
             _replicaDiscoveryEnabled = replicaDiscoveryEnabled;
 
-            _srvLookupClient = new Lazy<SrvLookupClient>();
+            _srvLookupClient = new SrvLookupClient();
 
             _clients = endpoints
                 .Select(endpoint => new ConfigurationClientWrapper(endpoint, new ConfigurationClient(endpoint, _credential, _clientOptions)))
@@ -173,11 +174,14 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 await RefreshFallbackClients(cancellationToken).ConfigureAwait(false);
             }
 
-            foreach (ConfigurationClientWrapper clientWrapper in _dynamicClients)
+            if (_dynamicClients != null)
             {
-                if (clientWrapper.BackoffEndTime <= now)
+                foreach (ConfigurationClientWrapper clientWrapper in _dynamicClients)
                 {
-                    yield return clientWrapper.Client;
+                    if (clientWrapper.BackoffEndTime <= now)
+                    {
+                        yield return clientWrapper.Client;
+                    }
                 }
             }
         }
@@ -275,7 +279,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             try
             {
-                results = await _srvLookupClient.Value.QueryAsync(_endpoint.DnsSafeHost, cancellationToken).ConfigureAwait(false);
+                results = await _srvLookupClient.QueryAsync(_endpoint.DnsSafeHost, cancellationToken).ConfigureAwait(false);
             }
             // Catch and log all exceptions thrown by srv lookup client to avoid new possible exceptions on app startup.
             catch (SocketException ex)
@@ -320,7 +324,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             foreach (string host in srvTargetHosts)
             {
                 if (!_clients.Any(c => c.Endpoint.Host.Equals(host, StringComparison.OrdinalIgnoreCase)) &&
-                    !_dynamicClients.Any(c => c.Endpoint.Host.Equals(host, StringComparison.OrdinalIgnoreCase)))
+                    !_dynamicClients.Any(c => c.Endpoint.Host.Equals(host, StringComparison.OrdinalIgnoreCase)) &&
+                    await IsValidEndpoint(host).ConfigureAwait(false))
                 {
                     var targetEndpoint = new Uri($"https://{host}");
 
@@ -333,6 +338,28 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             }
 
             _lastFallbackClientRefresh = DateTime.UtcNow;
+        }
+
+        private async Task<bool> IsValidEndpoint(string hostName)
+        {
+            if (string.IsNullOrWhiteSpace(hostName))
+            {
+                return false;
+            }
+
+            if (string.IsNullOrEmpty(_validDomain))
+            {
+                try
+                {
+                    _validDomain = await DomainVerifier.GetValidDomain(_endpoint, hostName).ConfigureAwait(false);
+                }
+                catch (SocketException)
+                {
+                    return false;
+                }
+            }
+            
+            return hostName.EndsWith(_validDomain, StringComparison.OrdinalIgnoreCase);
         }
     }
 }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
@@ -10,9 +10,11 @@ using Microsoft.Extensions.Configuration.AzureAppConfiguration.Constants;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Net.Sockets;
 using System.Runtime.CompilerServices;
+using System.Security.Authentication;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -37,7 +39,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         private IList<ConfigurationClientWrapper> _dynamicClients;
         private SrvLookupClient _srvLookupClient;
-        private string _validDomain = string.Empty;
+        private IEnumerable<string> _validDomains;
         private DateTimeOffset _lastFallbackClientRefresh = default;
         private DateTimeOffset _lastFallbackClientRefreshAttempt = default;
         private Logger _logger = new Logger();
@@ -259,7 +261,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             {
                 currentClient = _dynamicClients.FirstOrDefault(c => c.Client == client);
             }
-            
+
             return currentClient?.Endpoint;
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+    internal static class DomainVerifier
+    {
+        private const int TlsPort = 443;
+        private const string SubjectAltNameOid = "2.5.29.17";
+        private const string CommonName = "CN=";
+        private const string MultiDomainWildcard = "*.";
+
+        public static async Task<string> GetValidDomain(Uri originEndpoint, string srvHostName)
+        {
+            if (string.IsNullOrEmpty(srvHostName))
```
jimmyca15, line 27: That seems like argument exception. Why would empty or null be passed?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            if (sanExtension != null)
+            {
+                IEnumerable<string> formattedSanExtensions = sanExtension
+                    .Format(true)
+                    .Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+
+                foreach (string formattedExtension in formattedSanExtensions)
+                {
+                    // Valid pattern should be "DNS Name=*.domain.com"
```
jimmyca15, line 86: Then can you confirm starts with "dns name=" ?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+
+                    if (parts.Length != 2)
+                    {
+                        continue;
+                    }
+
+                    string value = parts[1];
+
+                    // Skip non-multi domain
+                    if (!value.StartsWith(MultiDomainWildcard))
```
jimmyca15, line 97: It's a single char, perhaps value[0] == foo is fine?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+                    {
+                        continue;
+                    }
+
+                    // .domain.com
+                    string domain = value.Substring(1);
```
jimmyca15, line 103: Should it be value.Substring(MultiDomainWildcard.Length); ?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+
+                    if (domain.Length > 1 && !validDomains.Contains(domain))
+                    {
+                        validDomains.Add(domain);
+                    }
+                }
+            }
+
+            // Get the Common Name (CN) from the Subject Name if Subject Alternative Name is not available
```
jimmyca15, line 112: All of our certs use SAN right?

jimmyca15: They should.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+                if (originEndpoint.DnsSafeHost.EndsWith(domain, StringComparison.OrdinalIgnoreCase))
+                {
+                    return domain;
+                }
+            }
+
+            return string.Empty;
+        }
+
+        private static async Task<List<string>> GetValidDomainsFromTlsCert(string endpoint)
```
jimmyca15, line 45:
```suggestion
        private static async Task<IEnumerable<string>> GetSubjectAlternativeNames(string endpoint)
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+        public static async Task<string> GetValidDomain(Uri originEndpoint, string srvHostName)
+        {
+            if (string.IsNullOrEmpty(srvHostName))
+            {
+                return string.Empty;
+            }
+
+            IEnumerable<string> validDomains = await GetValidDomainsFromTlsCert(srvHostName).ConfigureAwait(false);
+            
+            foreach (var domain in validDomains)
```
jimmyca15, line 34:
```suggestion
            foreach (string domain in validDomains)
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// Verify the domain name of the endpoint matches the certificate.
+    /// </summary>
+    internal static class DomainVerifier
+    {
+        private const int TlsPort = 443;
+        private const string SubjectAltNameOid = "2.5.29.17";
+        private const string CommonName = "CN=";
+        private const string MultiDomainWildcard = "*.";
+
+        public static async Task<string> GetValidDomain(Uri originEndpoint, string srvHostName)
```
jimmyca15, line 25: This should be able to return multiple valid domains if there are multiple matches.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -334,5 +339,27 @@ private async Task RefreshFallbackClients(CancellationToken cancellationToken)
 
             _lastFallbackClientRefresh = DateTime.UtcNow;
         }
+
+        private async Task<bool> IsValidEndpoint(string hostName)
+        {
+            if (string.IsNullOrWhiteSpace(hostName))
```
jimmyca15, line 345: Why would we have null or whitespace here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -334,5 +339,27 @@ private async Task RefreshFallbackClients(CancellationToken cancellationToken)
...
         }
+
+        private async Task<bool> IsValidEndpoint(string hostName)
+        {
+            if (string.IsNullOrWhiteSpace(hostName))
+            {
+                return false;
+            }
+
+            if (string.IsNullOrEmpty(_validDomain))
```
jimmyca15, line 350: I was expecting valid domains to be an enumerable.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
jimmyca15: then we can remove this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+            Debug.Assert(!string.IsNullOrEmpty(endpoint));
+
+            var validDomains = new List<string>();
+
+            X509Certificate2 cert = null;
+
+            using (var client = new TcpClient(endpoint, TlsPort))
+            using (var sslStream = new SslStream(client.GetStream(), leaveInnerStreamOpen: false))
+            {
+
```
jimmyca15, line 56: Looks like an unnecessary blank line here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+                // Initiate the connection, so it will download the server certificate
+                await sslStream.AuthenticateAsClientAsync(endpoint).ConfigureAwait(false);
+
+                var serverCertificate = sslStream.RemoteCertificate;
```
jimmyca15, line 60: This will throw for untrusted cert right?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+            var validDomains = new List<string>();
+
+            X509Certificate2 cert = null;
+
+            using (var client = new TcpClient(endpoint, TlsPort))
+            using (var sslStream = new SslStream(client.GetStream(), leaveInnerStreamOpen: false))
+            {
+
+                // Initiate the connection, so it will download the server certificate
+                await sslStream.AuthenticateAsClientAsync(endpoint).ConfigureAwait(false);
```
jimmyca15, line 58: Seems it can throw AuthenticationException. I don't see it handled.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: `.` is a valid targe host in SRV DNS, should if we can't trust the SRV DNS lookup result, we probably can get such a host name, and after trimming the tailing '.', we get an empty host name

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: This is a way recommend by a RFC doc, but given we ensure we use SAN, we can remove this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: We use `EndWith()` to validate, don't want to pass the validation for domains like ".xxxazconfig.io".

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
RichardChen820: Explained in another comment

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: It should always start with "dns name=" for DNS names. Update accordingly.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: Make sense, updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
RichardChen820: updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: Handled, as well as `InvalidOperationException` that may also thrown by this method

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: I think not, this doesn't verify whether it's a trusted, it won't verify the certificate chain, the expiration time, the issuer or the root cert is trusted or not. And seems doing such verification is overkill here because we just aim to verify the domain of srv endpoint, not really need to check the legality of that certificate. We just leverage the cert to ensure the acquired SRV endpoints are in the same domain as the original endpoint would be fine.

jimmyca15: No way ! We shouldn't be using any info from a certificate we haven't validated.

jimmyca15: In that case, someone controlling a bad endpoint could mint a certificate with whatever they want.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -36,7 +38,8 @@ internal class ConfigurationClientManager : IConfigurationClientManager
         private readonly bool _replicaDiscoveryEnabled;
 
         private IList<ConfigurationClientWrapper> _dynamicClients;
-        private Lazy<SrvLookupClient> _srvLookupClient;
+        private SrvLookupClient _srvLookupClient;
```
jimmyca15, line 41:
```suggestion
        private readonly SrvLookupClient _srvLookupClient;
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                var serverCertificate = sslStream.RemoteCertificate;
+
+                if (serverCertificate == null)
+                {
+                    return validDomains;
+                }
+
+                cert = new X509Certificate2(serverCertificate);
```
jimmyca15, line 62: That's disposable.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
jimmyca15: I put this comment because hardcoding the '1' here seems out of place given that the check for startswith uses MultiDomainWildCard. I think it would be a bit easier to follow if "*." was hardcoded into the StartsWith check to be consistent with the `1` being hardcoded here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            if (cert == null)
+            {
+                return validDomains;
+            }
+
+            // Get the Subject Alternative Name (SAN) extension
+            var sanExtension = cert.Extensions[SubjectAltNameOid];
+
```
jimmyca15, line 72: explicit type please.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: Then we should validate the certificate, what kind of validation should be done against the certificate? This is what we can get which can help to do the validation
https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslpolicyerrors?view=net-7.0

jimmyca15: Can we make an anonymous http call to the endpoint? I think http call has built in validation.

RichardChen820: https://learn.microsoft.com/en-us/dotnet/api/system.net.security.remotecertificatevalidationcallback?view=net-7.0
Such Example

RichardChen820: It does throw for untrusted cert as expected.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: Updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: Not bad :), updated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -0,0 +1,110 @@
...
+                cert = new X509Certificate2(serverCertificate);
+            }
+
+            if (cert == null)
+            {
+                return validDomains;
+            }
+
+            using (cert)
```
jimmyca15, line 69: It's a bit unfortunate the `using` is separated from the insantiation of the disposable object.

Could you use a private method to avoid separating the two?

```
using (X509Certificate2 cert = new X509Certificate2(serverCertificate))
{
    return GetValidDomains(cert);
}
```

jimmyca15 (approved):

PR #494, opened 2023-11-17
dedup same selectors with original precedence

### LINKED ISSUE #490: Should keep the original precedence after dedup selectors with same filters
After the dedup of selectors with exact same key filter and label filter, the precedence of selectors should not change. 

Consider case:
```cs
        options.Select("app.*", "dev");
        options.Select("app.*", "prod");
        options.Select("app.*", "dev");
```

Without dedup, "dev" wins. Dedup logic below will remove last occurrence, "prod" will win eventually. The dedup causes the change of precedence.

https://github.com/Azure/AppConfiguration-DotnetProvider/blob/75f3d6c186bbfb8c43e62f37207b1193e66e39e6/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs#L154

Eskibear:
fix #490


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -165,8 +165,14 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
             {
                 throw new ArgumentException("The characters '*' and ',' are not supported in label filters.", nameof(labelFilter));
             }
-
-            if (!_kvSelectors.Any(s => string.Equals(s.KeyFilter, keyFilter) && string.Equals(s.LabelFilter, labelFilter)))
+            var foundExactSameSelector = _kvSelectors.Find(s => string.Equals(s.KeyFilter, keyFilter) && string.Equals(s.LabelFilter, labelFilter));
```
avanigupta, line 168: We generally use `FirstOrDefault` instead of `Find`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -165,8 +165,14 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
...
+            if (foundExactSameSelector != null)
```
avanigupta, line 169: nit: can you leave an empty line before and after `if` blocks

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta, line 168: Since the type is not very obvious here, we should use explicit type instead of `var` for readability.

avanigupta, line 168: How about renaming the variable to something like `existingKvSelector`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -165,8 +165,14 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
...
+            {
+                // Move to the end, keeping precedence.
+                _kvSelectors.Remove(foundExactSameSelector);
+                _kvSelectors.Add(foundExactSameSelector);
+            }
+            else
```
avanigupta, line 175: Looks like we have the same problem while selecting feature flags or snapshots.

avanigupta: Also, same problem when adding to `_multiKeyWatchers` list. None of them are preserving precedence order.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
Eskibear: Searching the web a little bit, looks like `Find` has better performance (though a little bit). 
About the usage, here both methods return null if not found (if I don't specify a default value in FirstOrDefault). Just curious about the benefit of using `FirstOrDefault` here? Consistency?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
Eskibear: Are you suggesting me to fix them together in this PR, or in a separate PR?
Or we extract a utility to perform deduplication of selectors/watchers?

avanigupta: I think we can fix it in the same PR since the idea is the same. A common utility function might be good idea.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: Consistency, yes. Also, Find is only implemented for List type. FirstOrDefault is more generic for IEnumerable type.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
Eskibear: OK

Eskibear: MultiKeyWatchers has slightly different process while dedup, so I simply extracted a util for `List<KeyValueSelector>` instead of for both selectors and watchers. Let me know if it's ok...Otherwise I would have to add a generic utility, accepting detailed callbacks for existing/non-existing cases, which might make it a little bit complicated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -266,6 +248,10 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
                 {
                     // If UseFeatureFlags is called multiple times for the same key and label filters, last cache expiration time wins
                     multiKeyWatcher.CacheExpirationInterval = options.CacheExpirationInterval;
+
+                    // Move to the end, keeping precedence
+                    _multiKeyWatchers.Remove(multiKeyWatcher);
```
avanigupta, line 253: Can you remove it before modifying the CacheExpirationInterval ?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -238,10 +230,8 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
                 var featureFlagFilter = featureFlagSelector.KeyFilter;
                 var labelFilter = featureFlagSelector.LabelFilter;
 
-                if (!_kvSelectors.Any(selector => selector.KeyFilter == featureFlagFilter && selector.LabelFilter == labelFilter))
-                {
-                    Select(featureFlagFilter, labelFilter);
-                }
+                // For FeatureFlags, FeatureFlagOptions.Select should have higher precedence than AzureAppConfigurationOptions.Select
```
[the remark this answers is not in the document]
Eskibear: Do we encourage users to use `AzureAppConfigurationOptions.Select` with FeatureFlagMarker to select feature flags?
```c#
options.useFeatureFlags(ffOptions => {
  ffOptions.Select("Feature", "labelA");
});
...
options.Select(".appconfig.featureflag/Feature", "labelB");
```

Just to confirm that in this case, labelB wins, right?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
Eskibear: Sure, it would make more sense.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
avanigupta: > Do we encourage users to use AzureAppConfigurationOptions.Select with FeatureFlagMarker to select feature flags?

We dont encourage it, but we dont block them from doing it either.

> Just to confirm that in this case, labelB wins, right?

Yes that is correct. We load key-values and feature flags based on the insertion order in `AzureAppConfigurationOptions.KeyValueSelectors`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -166,15 +166,7 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
...
-                _kvSelectors.Add(new KeyValueSelector
-                {
-                    KeyFilter = keyFilter,
-                    LabelFilter = labelFilter
-                });
-            }
-
+            _kvSelectors.AppendUnique(keyFilter, labelFilter);
             return this;
         }
```
avanigupta, line 172: The problem still exists in `SelectSnapshot` API. We are not preserving precedence order for snapshots.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ListKeyValueSelectorExtensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.Models;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+    internal static class ListKeyValueSelectorExtensions
```
avanigupta, line 10: I think we may be able to generalize the concept of preserving precedence order while maintaining unique elements in the IEnumerable collection. What do you think about making this `IEnumerableExtensions` with the following extension method:

`public static void AppendUnique<T>(this IEnumerable<T> selectors, T selector)`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
Eskibear: Then I would simply remove this line of comments. Next line works because the correct dedup logic is already in `AzureAppConfigurationOptions.Select`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ListKeyValueSelectorExtensions.cs
Eskibear: > public static void AppendUnique<T>(this IEnumerable<T> selectors, T selector)

I was considering that before. then we might also need to specify a `Equals` method for T? BTW Cons for above util:
- Calling this function, we would first construct a new instance of the selector, regardless of whether it's duplicate. Not sure if the slight increase of memory consumption is important or not.

avanigupta: We dont necessarily need equals method because we can still check all 3 properties manually. But I'm also not against adding Equals override for `KeyValueSelector` since we have precedence for doing the same in `KeyValueWatcher` and `KeyValueIdentifier`.

About the memory consumption, I think KeyValueSelector is a lightweight object with just 2 (or 3) string values. Since we are not persisting these instances of `KeyValueSelector`, memory consumption doesn't sound like a big concern to me.

Eskibear: Then we might use `List` instead of `IEnumerable` as it doesn't have `Add` or `Remove`...

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
Eskibear: fixed. but next step is considering whether to use a generic utility for it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ListKeyValueSelectorExtensions.cs
Eskibear: updated using `public static void AppendUnique<T>(this List<T> items, T item)`

Note, since we have the generic util, I'm also using it to dedup multiKeyWatchers.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ListExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT license.
+//
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+    internal static class ListExtensions
+    {
+
```
avanigupta, line 11: nit: extra line

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
```diff
@@ -23,5 +23,31 @@ public class KeyValueSelector
...
+            return false;
+        }
+
+        /// <summary>
+        /// Serves as the hash function.
+        /// </summary>
+        /// <returns>A hash code for the current object.</returns>
+        public override int GetHashCode()
+        {
+            return SnapshotName != null ? SnapshotName.GetHashCode() : KeyFilter.GetHashCode() ^ LabelFilter.GetHashCode();
```
avanigupta, line 50: `LabelFilter.GetHashCode()` can throw if `LabelFilter` is null

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ListExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+        public static void AppendUnique<T>(this List<T> items, T item)
```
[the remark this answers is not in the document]
Eskibear: I assumed it's an internally used util that we make sure the params are not null.

Sure let's add null check, throw ArgumentNullException

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ListExtensions.cs
Eskibear: just curious about is there any Linting tools performing emtpy line check before I commit code?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
amerjusupovic: I also think it makes sense to check each property for null, since `GetHashCode` shouldn't necessarily need to know the valid cases.

This is sort of related, but the `KeyValueWatcher` can throw a null-ref if you pass in `null` as the key when registering a key-value for refresh. In that scenario, `Key` being null is unexpected since there shouldn't be a null key in the configuration. It seems like we should throw `ArgumentNullException` in `Register` before the `KeyValueWatcher` throws a null-ref, since it we already do this in the `Select` method.

avanigupta: > It seems like we should throw `ArgumentNullException` in `Register` before the `KeyValueWatcher` throws a null-ref, since we already do this in the `Select` method.

Good catch! @amerjusupovic could you help to create a tracking issue or PR for this bug?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ListExtensions.cs
avanigupta: We dont have any linting tools in this repo yet

Eskibear: https://learn.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2022#supported-settings

Looks like VS natively supports [EditConfig](https://editorconfig.org/) which is popular among other languages. But not sure if it supports rules like extra line around if blocks... Anyway, we might add some linting tools some day, if necessary.  :)

PR #495, opened 2023-11-17
Robust checking subject alternative name

RichardChen820:
Robust checking subject alternative name

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     {
         private const int TlsPort = 443;
         private const string SubjectAltNameOid = "2.5.29.17";
-        private const string DnsName = "DNS Name=";
+        private readonly static char[] DnsNameDelimiters = new char[2] { ':', '=' };
 
         public static async Task<IEnumerable<string>> GetValidDomains(Uri originEndpoint, string srvHostName)
         {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -66,36 +73,40 @@ private static IEnumerable<string> GetDomainsFromSanExtension(X509Certificate2 c
 
             X509Extension sanExtension = cert.Extensions[SubjectAltNameOid];
 
-            if (sanExtension != null)
+            // For the DNS name, the tag class is ContextSpecific and tag value is 2 according to RFC 5280
+            var dnsNameTag = new Asn1Tag(TagClass.ContextSpecific, tagValue: 2, isConstructed: false);
+
+            AsnReader reader = new AsnReader(sanExtension.RawData, AsnEncodingRules.BER);
```
jimmyca15, line 79: Is any of this disposable?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
```diff
@@ -19,7 +20,6 @@ internal static class DomainVerifier
     {
         private const int TlsPort = 443;
         private const string SubjectAltNameOid = "2.5.29.17";
-        private const string DnsName = "DNS Name=";
 
         public static async Task<IEnumerable<string>> GetValidDomains(Uri originEndpoint, string srvHostName)
```
jimmyca15, line 24: I think we need to add a test for this that calls a fake App Config endpoint like azure.aconfig.io and tries to get the domains.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: I had checked,  it's not disposable

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/DomainVerifier.cs
RichardChen820: Then we expect it fails, get no valid domain?

PR #499, opened 2023-11-21
Validate endpoint by domain name

RichardChen820:
Validate endpoint by domain name


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -343,39 +345,35 @@ await IsValidEndpoint(host).ConfigureAwait(false))
...
-        private async Task<bool> IsValidEndpoint(string hostName)
+        private string GetValidDomain(Uri endpoint)
         {
-            Debug.Assert(!string.IsNullOrEmpty(hostName));
+            Debug.Assert(endpoint != null);
+
+            string hostName = endpoint.Host;
 
-            if (_validDomains == null || !_validDomains.Any())
+            foreach (string domainPrefix in TrustedDomainPrefixes)
```
jimmyca15, line 354: The search needs to be from right to left.

jimmyca15: Seems to be fine since you look for segment prefixed with label separator.

jimmyca15: I'd still feel better if we went from right to left and found exact label match.

jimmyca15: Perhaps `.LastIndexOf($".{label}.")`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -37,15 +36,16 @@ internal class ConfigurationClientManager : IConfigurationClientManager
...
 
         private IList<ConfigurationClientWrapper> _dynamicClients;
-        private IEnumerable<string> _validDomains;
         private DateTimeOffset _lastFallbackClientRefresh = default;
         private DateTimeOffset _lastFallbackClientRefreshAttempt = default;
         private Logger _logger = new Logger();
 
         private static readonly TimeSpan FallbackClientRefreshExpireInterval = TimeSpan.FromHours(1);
         private static readonly TimeSpan MinimalClientRefreshInterval = TimeSpan.FromSeconds(30);
+        private static readonly string[] TrustedDomainPrefixes = new[] { ".azconfig", ".appconfig" };
```
jimmyca15, line 48: It's not a prefix. `azconfig` and `appconfig` are both "label"s in the domain name.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
RichardChen820: Got you, I will update

RichardChen820: But we have a domain in such a pattern right? `azconfig-test.io`  Can we make exact label match?  Or we list all labels we have (or potentially have in the future) and traverse them?

jimmyca15: We'll have to think about azconfig-test.io. Most likely, we'll have to phase it out and adopt a proper .azconfig.xyz domain in ppe.

RichardChen820: Updated

jimmyca15 (approved):

jimmyca15 (changes requested):
we should not have azconfig-test, appconfig-test.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -37,15 +36,16 @@ internal class ConfigurationClientManager : IConfigurationClientManager
...
 
         private IList<ConfigurationClientWrapper> _dynamicClients;
-        private IEnumerable<string> _validDomains;
         private DateTimeOffset _lastFallbackClientRefresh = default;
         private DateTimeOffset _lastFallbackClientRefreshAttempt = default;
         private Logger _logger = new Logger();
 
         private static readonly TimeSpan FallbackClientRefreshExpireInterval = TimeSpan.FromHours(1);
         private static readonly TimeSpan MinimalClientRefreshInterval = TimeSpan.FromSeconds(30);
+        private static readonly string[] TrustedDomainLabels = new[] { "azconfig", "azconfig-test", "appconfig", "appconfig-test" };
```
jimmyca15, line 48: We need to avoid azconfig-test and appconfig-test. We don't use that domain across all clouds.

I know it's to get PPE to work, but we need to solve the root problem which is the use of a domain other than "azconfig".

jimmyca15 (approved):

PR #503, opened 2023-12-12
Add null check for key parameter in AzureAppConfigurationRefreshOptions.Register

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefreshOptions.cs
```diff
@@ -35,6 +35,11 @@ public AzureAppConfigurationRefreshOptions Register(string key, bool refreshAll)
         /// <param name="refreshAll">If true, a change in the value of this key refreshes all key-values being used by the configuration provider.</param>
         public AzureAppConfigurationRefreshOptions Register(string key, string label = LabelFilter.Null, bool refreshAll = false)
         {
+            if (string.IsNullOrEmpty(key))
```
zhenlan, line 38: What's the behavior today if a null is passed in? No errors but nothing is monitored?

amerjusupovic: I could've added some context, but `KeyValueWatcher.GetHashCode()` throws a null-ref if `key` is null. I mentioned it in [this comment](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/494#discussion_r1417979414).

zhenlan: Cool. Then it's not a breaking change.

zhenlan (approved):

Eskibear (approved):

avanigupta (approved):

PR #507, opened 2024-01-04
Remove MaxAttempts for CalculateBackoffTime

zhiyuanliang-ms:
The variable name `MaxAttempts` is misleading. This variable is used for preventing the case that the attempts is larger than the bit number of long int. For example, `1 << 65 ` will give the same result as `1 << 1`.
 I think the name `MaxSafeShiftBit` or something similar will be more informative. The name `MaxAttempts` is quite misleading because it feels like if the attempts reach the `MaxAttempts`, the retry will be ended. 
However, in the code,
```C#
//
// IMPORTANT: This can overflow
double calculatedMilliseconds = Math.Max(1, minDuration.TotalMilliseconds) * ((long)1 << Math.Min(attempts, MaxAttempts));

if (calculatedMilliseconds > maxDuration.TotalMilliseconds ||
        calculatedMilliseconds <= 0 /*overflow*/)
{
    calculatedMilliseconds = maxDuration.TotalMilliseconds;
}
```
In fact, attempts will never reach `MaxAttempts`, because `maxDuration` is usually very small.

In [Python Provider](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py#L411), the implementation is the same as dotnet provider. 
The `max_attempts = 63` makes no sense in python provider to keep the same behavior since there is no overflow in [python](https://docs.python.org/3/c-api/long.html#integer-objects).

In this PR, I suggested using Math.Pow instead of << operation.
I tested the performance difference, it is only about 0.01 ms. 
Double type will not have the overflow issue. The "overflow" will return double.PositiveInfinity which is larger than any double number. We can remove the code of handling overflow.

Also, I found the [discussion](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/488#discussion_r1387429187) about the exponential strategy which called out the possibility to adjust the factor.

Also, I feels `Math.Max(1, minDuration.TotalMilliseconds)` weird. I suggest adding some out range check for all TimeSpan parameters.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 {
     internal static class TimeSpanExtensions
     {
-        private const double ExponentialBackoffFactor = 2;
+        private const int MaxAttempts = 63;
         private const double JitterRatio = 0.25;
 
         private static readonly IList<KeyValuePair<TimeSpan, TimeSpan>> StartupMaxBackoffDurationIntervals = new List<KeyValuePair<TimeSpan, TimeSpan>>
@@ -93,9 +93,12 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
                 return minDuration;
             }
 
-            double calculatedMilliseconds = minDuration.TotalMilliseconds * Math.Pow(ExponentialBackoffFactor, attempts);
+            //
+            // IMPORTANT: This can overflow
+            double calculatedMilliseconds = minDuration.TotalMilliseconds * ((long)1 << Math.Min(attempts, MaxAttempts));
 
-            if (calculatedMilliseconds > maxDuration.TotalMilliseconds)
+            if (calculatedMilliseconds > maxDuration.TotalMilliseconds ||
+                    calculatedMilliseconds <= 0 /*overflow*/)
             {
                 calculatedMilliseconds = maxDuration.TotalMilliseconds;
             }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 {
     internal static class TimeSpanExtensions
     {
-        private const int MaxAttempts = 63;
+        private const int MaxSafeShiftBit = 63;
         private const double JitterRatio = 0.25;
 
         private static readonly IList<KeyValuePair<TimeSpan, TimeSpan>> StartupMaxBackoffDurationIntervals = new List<KeyValuePair<TimeSpan, TimeSpan>>
@@ -95,7 +95,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 
             //
             // IMPORTANT: This can overflow
-            double calculatedMilliseconds = minDuration.TotalMilliseconds * ((long)1 << Math.Min(attempts, MaxAttempts));
+            double calculatedMilliseconds = minDuration.TotalMilliseconds * (1L << Math.Min(attempts, MaxSafeShiftBit));
 
             if (calculatedMilliseconds > maxDuration.TotalMilliseconds ||
                     calculatedMilliseconds <= 0 /*overflow*/)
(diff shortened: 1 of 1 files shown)


Eskibear:
I agree with most of above points. 

Btw I just glanced at the original discussion mentioned above, where a good analysis of factor was done and 1.5 was decided to be the value. Just wondering why it  eventually used 2 (with bitwise operation)? Performance or any other concern.

jimmyca15:
> Just wondering why it eventually used 2 (with bitwise operation)? Performance or any other concern?

Because 2 is the de-facto approach for exponential backoff. Using 1.5 instead was proposed as an attempt to increase the amount of retry attempts performed on startup. An offline discussion was had about this. Changing the exponential growth factor isn't a real solution to the problem that was trying to be solved. After discussion to pin down what was actually desired when the original suggestion was made, it revealed that the desire was to have a max time between attempts on startup. So the final design was to keep an exponential growth factor of 2 over the long term, but design certain cut off points on startup where the max time between attempts is determined.

avanigupta:
> In this PR, I suggested using Math.Pow instead of << operation.
> I tested the performance difference, it is only about 0.01 ms.

Bitwise operations are much faster than arithmetic operations because bit manipulation is optimized at the hardware level. `Math.Pow` is more generic and designed to handle floating point calculations better. Hence, it's not optimized for the special case that is calculating powers of 2. I think we should keep that optimization until we decide to change the exponential factor of 2.


> The name MaxAttempts is quite misleading because it feels like if the attempts reach the MaxAttempts, the retry will be ended.

I agree with this, we can change the name of the variable to clarify the intended behavior.

avanigupta:
> The max_attempts = 63 makes no sense in [python](https://docs.python.org/3/c-api/long.html#integer-objects) provider to keep the same behavior since there is no overflow in python.

@mrm9084 and I discussed this a while ago. Matt mentioned that `1 << 14284` can overflow in Python. Realistically, we will never reach that limit but it's still a good practice to have some upper bound for safeguarding. That's why we chose 63 to be consistent with other providers.

mrm9084:
> > The max_attempts = 63 makes no sense in [python](https://docs.python.org/3/c-api/long.html#integer-objects) provider to keep the same behavior since there is no overflow in python.
> 
> @mrm9084 and I discussed this a while ago. Matt mentioned that `1 << 14284` can overflow in Python. Realistically, we will never reach that limit but it's still a good practice to have some upper bound for safeguarding. That's why we chose 63 to be consistent with other providers.

I'll still update the variable name as I agree it isn't a good one.

zhiyuanliang-ms:
@RichardChen820 The name "MaxSafeShiftBit" is more informative and indicate this variable is only for safeguarding the shift operation. When I was reading the code of dotnet provider, I was confused about the "MaxAttempts", I expected there would be some logic to end refresh when Attempts reachs MaxAttempts.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 {
     internal static class TimeSpanExtensions
     {
-        private const int MaxAttempts = 63;
+        private const int MaxSafeShiftBit = 63;
```
avanigupta, line 11: The new name is also a bit odd. 63 is not a "Bit" as the name suggests.
How about calling it `MaxShiftCount`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -70,7 +95,7 @@ public static TimeSpan CalculateBackoffDuration(this TimeSpan minDuration, TimeS
 
             //
             // IMPORTANT: This can overflow
-            double calculatedMilliseconds = Math.Max(1, minDuration.TotalMilliseconds) * ((long)1 << Math.Min(attempts, MaxAttempts));
+            double calculatedMilliseconds = minDuration.TotalMilliseconds * (1L << Math.Min(attempts, MaxSafeShiftBit));
```
avanigupta, line 98: Let's keep the original `Math.Max` calculation because `minDuration.TotalMilliseconds` can still be less than 1 ms. We dont want to reduce the calculated exponential by multiplying it with a number less than 1.

zhiyuanliang-ms: Sounds good.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
zhiyuanliang-ms: How about MaxShiftLimit?

zhiyuanliang-ms: Discussed offline. We will use "SafeShiftLimit"

avanigupta (approved):

PR #508, opened 2024-01-05
Refactor and move backoff logic to AzureAppConfigurationProvider

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -193,11 +200,31 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                        Uri endpoint = _configClientManager.GetEndpointForClient(client);
+
+                        if (!_configClientBackoffs.TryGetValue(endpoint, out ConfigurationClientBackoffStatus clientBackoffStatus))
+                        {
+                            UpdateClientBackoffStatus(endpoint, true);
+
+                            clientBackoffStatus = _configClientBackoffs[endpoint];
+                        }
+
+                        return clientBackoffStatus?.BackoffEndTime <= DateTimeOffset.UtcNow;
```
jimmyca15, line 218: Regarding the null check (`?`). I see that we either get existing backoff status or create a new one. How could it be null?

jimmyca15: `DateTimeOffset.UtcNow`, should calculate once outside of the loop.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -193,11 +200,31 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+
+                    //
+                    // Filter clients based on their backoff status
+                    clients = clients.Where(client => 
+                    {
+                        Uri endpoint = _configClientManager.GetEndpointForClient(client);
+
+                        if (!_configClientBackoffs.TryGetValue(endpoint, out ConfigurationClientBackoffStatus clientBackoffStatus))
+                        {
+                            UpdateClientBackoffStatus(endpoint, true);
```
jimmyca15, line 213: There's no definition of success/fail here. I think it might be a bit better to just initialize it inline here and then add it into the dictionary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -319,12 +282,11 @@ private async Task RefreshFallbackClients(CancellationToken cancellationToken)
             _lastFallbackClientRefresh = DateTime.UtcNow;
         }
 
-        private bool IsFallbackClientDiscoveryDue(DateTimeOffset dateTime)
+        private bool IsFallbackClientDiscoveryDue(DateTimeOffset now)
```
jimmyca15, line 285: Unused.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Right it shouldn't be null, leftover from old change

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1112,6 +1145,31 @@ private IEnumerable<ConfigurationSetting> GetCurrentKeyValueCollection(string ke
...
+            {
+                clientBackoffStatus.BackoffEndTime = DateTimeOffset.UtcNow;
+
+                clientBackoffStatus.FailedAttempts = 0;
+            }
+            else
+            {
+                clientBackoffStatus.FailedAttempts++;
+
+                TimeSpan backoffDuration = _options.MinBackoffDuration.CalculateBackoffDuration(FailOverConstants.MaxBackoffDuration, clientBackoffStatus.FailedAttempts);
```
Eskibear, line 1165: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/e9a3aa5b49958bd6190fb91c7a16d10ac46e3967/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs#L73

Here basically the backoff time = minDuration * (1<<attemps) when attempts >=2.
attemps: 1, 2, 3, 4...
backoffTime( x minDuration): 1x, **4x**, 8x, 16x...

Should we change it to `1<<(attemps -1)`, so the backoff time becomes 1x,2x,4x,8x, ...?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -101,6 +101,12 @@ internal IEnumerable<IKeyValueAdapter> Adapters
         /// <remarks>This property is used only for unit testing.</remarks>
         internal IConfigurationClientManager ClientManager { get; set; }
 
+        /// <summary>
+        /// An optional timespan value to set the minimum backoff duration to a value other than the default.
+        /// </summary>
+        /// <remarks>This property is used only for unit testing.</remarks>
```
jimmyca15, line 107: This property is not only used for unit testing. It's referenced in this file for backoff. It would be appropriate to say this property is internal for availability in unit tests. But even that seems unnecessary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: That is a good idea. I think this happens because we optimized to return when attempts=1, so we never get the 2x backoff duration.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: That makes sense, I'll remove the remarks.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Makes sense, just gonna subtract one from attempts in that code then.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -193,11 +200,35 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
...
+                    // Filter clients based on their backoff status
+                    clients = clients.Where(client => 
                     {
+                        Uri endpoint = _configClientManager.GetEndpointForClient(client);
+
+                        if (!_configClientBackoffs.TryGetValue(endpoint, out ConfigurationClientBackoffStatus clientBackoffStatus))
+                        {
+                            clientBackoffStatus = new ConfigurationClientBackoffStatus();
+
+                            clientBackoffStatus.BackoffEndTime = utcNow;
```
RichardChen820, line 215: Seems  line 215-217 are not necessary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -48,6 +48,9 @@ internal class ConfigurationClientManager : IConfigurationClientManager, IDispos
 
         private readonly CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
 
+        // Only used for unit testing
+        public int RefreshClientsCalled { get; set; } = 0;
```
RichardChen820, line 52: internal is good enough? Not necessary to be a public?

RichardChen820 (changes requested):

RichardChen820 (approved):

PR #517, opened 2024-02-10
Populate feature flag telemetry metadata when feature flag telemetry is enabled

amerjusupovic:
Adding 3 values to feature flag telemetry metadata when a flag has telemetry enabled: `ETag`, `FeatureFlagId`, and `FeatureFlagReference`.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -75,6 +75,14 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     catch (FormatException) { }
                 }
 
+                // This is used only when a test passes in a ConfigurationClientManager but no endpoint or connection string is provided.
+                ConfigurationClient client = _configClientManager.GetClients().FirstOrDefault();
+
+                if (client != null)
+                {
+                    return _configClientManager.GetEndpointForClient(client);
+                }
+
                 return null;
             }
         }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -75,14 +75,6 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     catch (FormatException) { }
                 }
 
-                // This is used only when a test passes in a ConfigurationClientManager but no endpoint or connection string is provided.
-                ConfigurationClient client = _configClientManager.GetClients().FirstOrDefault();
-
-                if (client != null)
-                {
-                    return _configClientManager.GetEndpointForClient(client);
-                }
-
                 return null;
             }
         }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -916,7 +916,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     continue;
                 }
 
-                IEnumerable<KeyValuePair<string, string>> kvs = await adapter.ProcessKeyValue(setting, _logger, cancellationToken).ConfigureAwait(false);
+                IEnumerable<KeyValuePair<string, string>> kvs = await adapter.ProcessKeyValue(setting, AppConfigurationEndpoint, _logger, cancellationToken).ConfigureAwait(false);
 
                 if (kvs != null)
                 {
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
@@ -10,6 +10,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
     {
         /// <summary>
         /// Converts a byte array to Base64URL string and removes trailing '=' characters.
+        /// Base64 description: https://datatracker.ietf.org/doc/html/rfc4648.html#section-4
         /// </summary>
         public static string ToBase64Url(this byte[] bytes)
         {
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
@@ -8,6 +8,9 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 {
     internal static class BytesExtensions
     {
+        /// <summary>
+        /// Converts a byte array to Base64URL string and removes trailing '=' characters.
+        /// </summary>
         public static string ToBase64Url(this byte[] bytes)
         {
             string bytesBase64 = Convert.ToBase64String(bytes);
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using System.Text;
+using System;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+    internal static class BytesExtensions
+    {
+        public static string ToBase64Url(this byte[] bytes)
+        {
+            string featureFlagIdBase64 = Convert.ToBase64String(bytes);
+
+            int indexOfEquals = featureFlagIdBase64.IndexOf("=");
+
+            int stringBuilderCapacity = indexOfEquals != -1 ? indexOfEquals : featureFlagIdBase64.Length;
+
+            StringBuilder featureFlagIdBuilder = new StringBuilder(stringBuilderCapacity);
+
+            for (int i = 0; i < stringBuilderCapacity; i++)
+            {
+                if (featureFlagIdBase64[i] == '+')
+                {
+                    featureFlagIdBuilder.Append('-');
+                }
+                else if (featureFlagIdBase64[i] == '/')
+                {
+                    featureFlagIdBuilder.Append('_');
+                }
+                else
+                {
+                    featureFlagIdBuilder.Append(featureFlagIdBase64[i]);
+                }
+            }
+
+            return featureFlagIdBuilder.ToString();
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
@@ -14,10 +14,12 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
 
             int indexOfEquals = bytesBase64.IndexOf("=");
 
+            // Remove all instances of "=" at the end of the string that were added as padding
             int stringBuilderCapacity = indexOfEquals != -1 ? indexOfEquals : bytesBase64.Length;
 
             StringBuilder stringBuilder = new StringBuilder(stringBuilderCapacity);
 
+            // Construct Base64URL string by replacing characters in Base64 conversion that are not URL safe
             for (int i = 0; i < stringBuilderCapacity; i++)
             {
                 if (bytesBase64[i] == '+')
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
     internal static class BytesExtensions
     {
         /// <summary>
-        /// Converts a byte array to Base64URL string and removes trailing '=' characters.
+        /// Converts a byte array to Base64URL string with optional padding ('=') characters removed.
         /// Base64 description: https://datatracker.ietf.org/doc/html/rfc4648.html#section-4
         /// </summary>
         public static string ToBase64Url(this byte[] bytes)
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
@@ -10,31 +10,31 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
     {
         public static string ToBase64Url(this byte[] bytes)
         {
-            string featureFlagIdBase64 = Convert.ToBase64String(bytes);
+            string bytesBase64 = Convert.ToBase64String(bytes);
 
-            int indexOfEquals = featureFlagIdBase64.IndexOf("=");
+            int indexOfEquals = bytesBase64.IndexOf("=");
 
-            int stringBuilderCapacity = indexOfEquals != -1 ? indexOfEquals : featureFlagIdBase64.Length;
+            int stringBuilderCapacity = indexOfEquals != -1 ? indexOfEquals : bytesBase64.Length;
 
-            StringBuilder featureFlagIdBuilder = new StringBuilder(stringBuilderCapacity);
+            StringBuilder stringBuilder = new StringBuilder(stringBuilderCapacity);
 
             for (int i = 0; i < stringBuilderCapacity; i++)
             {
-                if (featureFlagIdBase64[i] == '+')
+                if (bytesBase64[i] == '+')
                 {
-                    featureFlagIdBuilder.Append('-');
+                    stringBuilder.Append('-');
                 }
-                else if (featureFlagIdBase64[i] == '/')
+                else if (bytesBase64[i] == '/')
                 {
-                    featureFlagIdBuilder.Append('_');
+                    stringBuilder.Append('_');
                 }
                 else
                 {
-                    featureFlagIdBuilder.Append(featureFlagIdBase64[i]);
+                    stringBuilder.Append(bytesBase64[i]);
                 }
             }
 
-            return featureFlagIdBuilder.ToString();
+            return stringBuilder.ToString();
         }
     }
 }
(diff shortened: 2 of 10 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -75,6 +75,14 @@ public Uri AppConfigurationEndpoint
                     catch (FormatException) { }
                 }
 
+                // This is used only when a test passes in a ConfigurationClientManager but no endpoint or connection string is provided.
```
rossgrambo, line 78: I know this test-code-within-the-library pattern wasn't introduced in this PR, but I don't like that it's happening. I feel like we should be able to avoid it with a mocked dependency.

amerjusupovic: Yea I agree it's weird, I think there might be a way that I just didn't see before so I'll try it out.

amerjusupovic: I just changed the test to call `Connect` with an endpoint too but it isn't used, and it's stored in the endpoint value in the provider.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -195,18 +198,38 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
+                if (telemetry.Enabled)
+                {
+                    byte[] featureFlagIdHash;
+
+                    using (HashAlgorithm hashAlgorithm = SHA256.Create())
+                    {
+                        featureFlagIdHash = hashAlgorithm.ComputeHash(Encoding.UTF8.GetBytes($"{setting.Key}\n{setting.Label}"));
+                    }
+
+                    string featureFlagId = WebUtility.UrlEncode(Convert.ToBase64String(featureFlagIdHash));
```
rossgrambo, line 218: This isn't quite accurate for what Jimmy purposed. He suggested Base64Url encoding- which is different than Base64 + Url encoding (see https://stackoverflow.com/a/55389212/5937135). 

I see an AspNetCore method for it here:  https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webutilities.webencoders.base64urlencode?view=aspnetcore-8.0

You could achieve base64url encoding with
```csharp
string base64UrlStr = Convert.ToBase64String(Encoding.UTF8.GetBytes(target))
    .TrimEnd('=') // Remove padding (optional)
    .Replace('+', '-') // Replace '+' with '-'
    .Replace('/', '_'); // Replace '/' with '_'
```

But using a built-in method is probably preferred.

amerjusupovic: Oh I see, I think I misunderstood how the URL conversion works. I did see that AspNetCore method, but it's not compatible with the target frameworks in the main project, so I can't find another built in method that exists. I can change it to this for now.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -195,18 +197,41 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
                     }
                 }
+
+                if (telemetry.Enabled)
+                {
+                    byte[] featureFlagIdHash;
+
+                    using (HashAlgorithm hashAlgorithm = SHA256.Create())
+                    {
+                        featureFlagIdHash = hashAlgorithm.ComputeHash(Encoding.UTF8.GetBytes($"{setting.Key}\n{setting.Label}"));
```
rossgrambo, line 214: I don't recall what the plan was- but if label is null- are we omitting the `\n` and null? I think currently this would give the string: 

```text
Key
null
```

rossgrambo (approved):
1 remaining comment but otherwise lgtm

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: It should evaluate to "Key\n", but I'm not sure if we clarified whether we want to keep the newline character as part of the value even if there's a null label? I think it's fine as it is but just wanted to call it out.

zhenlan: Do we differentiate null vs. empty label?

amerjusupovic: I assumed we wouldn't, when I tried on portal I couldn't pass an empty string, and we also treat whitespace/null as the same for label filters in calls to `Select`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -195,18 +197,56 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
+                    byte[] featureFlagIdHash;
+
+                    using (HashAlgorithm hashAlgorithm = SHA256.Create())
+                    {
+                        featureFlagIdHash = hashAlgorithm.ComputeHash(Encoding.UTF8.GetBytes($"{setting.Key}\n{(string.IsNullOrWhiteSpace(setting.Label) ? null : setting.Label)}"));
+                    }
+
+                    string featureFlagIdBase64 = Convert.ToBase64String(featureFlagIdHash);
+
+                    StringBuilder featureFlagIdBuilder = new StringBuilder(featureFlagIdBase64[featureFlagIdBase64.Length - 1] == '=' ? featureFlagIdBase64.Length - 1 : featureFlagIdBase64.Length);
```
jimmyca15, line 219: Can't b64 have two equal signs at the end?

also, seems like you should be using `IndexOf('=')` and judge based off that.

rossgrambo: Good catch. There's an appendix about implementing it we could follow https://datatracker.ietf.org/doc/html/rfc7515#appendix-C, they split and remove all "=", but index would work too (and probably better)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -216,7 +216,9 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
 
                     string featureFlagIdBase64 = Convert.ToBase64String(featureFlagIdHash);
 
-                    StringBuilder featureFlagIdBuilder = new StringBuilder(featureFlagIdBase64[featureFlagIdBase64.Length - 1] == '=' ? featureFlagIdBase64.Length - 1 : featureFlagIdBase64.Length);
+                    int indexOfEquals = featureFlagIdBase64.IndexOf("=");
+
+                    StringBuilder featureFlagIdBuilder = new StringBuilder(indexOfEquals != -1 ? indexOfEquals : featureFlagIdBase64.Length);
 
                     for (int i = 0; i < featureFlagIdBuilder.Capacity; i++)
```
jimmyca15, line 223: Given string builder capacity varies based off characters appended to it (auto-expansion) I would stay away from using it in a for loop. Instead, prefer a clearly static value.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -195,18 +197,60 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
+                        {
+                            featureFlagIdBuilder.Append(featureFlagIdBase64[i]);
+                        }
+                    }
+
+                    keyValues.Add(new KeyValuePair<string, string>($"{telemetryPath}:{FeatureManagementConstants.Metadata}:{FeatureManagementConstants.FeatureFlagId}", featureFlagIdBuilder.ToString()));
+
+                    if (endpoint != null)
+                    {
+                        string featureFlagReference = $"{endpoint.AbsoluteUri}kv/{setting.Key}{(setting.Label != null ? $"?label={setting.Label}" : "")}";
```
zhenlan, line 245: NIT: instead of null check, should we do string.IsNullOrWhiteSpace(setting.Label) to be consistent with what you have above?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -195,18 +197,60 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
 
                 string telemetryPath = $"{featureFlagPath}:{FeatureManagementConstants.Telemetry}";
 
-                if (telemetry.Enabled)
-                {
-                    keyValues.Add(new KeyValuePair<string, string>($"{telemetryPath}:{FeatureManagementConstants.Enabled}", telemetry.Enabled.ToString()));
-                }
-
                 if (telemetry.Metadata != null)
```
zhenlan, line 200: Should we move this inside the `if (telemetry.Enabled)`? No one cares about it if telemetry is not enabled.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -195,18 +197,60 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
+                        }
+                        else if (featureFlagIdBase64[i] == '/')
+                        {
+                            featureFlagIdBuilder.Append('_');
+                        }
+                        else
+                        {
+                            featureFlagIdBuilder.Append(featureFlagIdBase64[i]);
+                        }
+                    }
```
zhenlan, line 239: Can we add some comments to explain how the featureFlagId is supposed to be calculated? It's not clear why we are doing some of the operations we are doing here.

jimmyca15: I think it would also add clarity if we move the b64url into an extension method.

ToBase64UrlString(this bytes[]);

At that point the code will be more self explanatory.

zhenlan: Right, won't mind if we move the whole featureFlagId calculation to a separate method. This function is getting way too long.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -195,18 +197,60 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
+                    byte[] featureFlagIdHash;
+
+                    using (HashAlgorithm hashAlgorithm = SHA256.Create())
+                    {
+                        featureFlagIdHash = hashAlgorithm.ComputeHash(Encoding.UTF8.GetBytes($"{setting.Key}\n{(string.IsNullOrWhiteSpace(setting.Label) ? null : setting.Label)}"));
+                    }
+
+                    string featureFlagIdBase64 = Convert.ToBase64String(featureFlagIdHash);
+
+                    int indexOfEquals = featureFlagIdBase64.IndexOf("=");
```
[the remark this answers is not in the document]
jimmyca15: comment will be obsoleted by ToBase64Url extension method which takes unknown bytes as input.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: I see now, I thought there was a scenario in feature management where it could still be used but I don't see how. I can move it inside the enabled block too.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: I think I'll go with the extension for now since it makes sense to me conceptually, but let me know if you have any other concerns

zhenlan: I just want to point out that the featureFlagId calculation is "by convention". It's not persisted anywhere. Portal and all our other providers will do the same calculation on-fly. Since the .NET provider is the first one doing this, it will make everyone's life easier if we can encapsulate it in one function with detailed comments so others can review/reference.

jimmyca15: @amerjusupovic Can we please go with b64url extension method and private static method for `CalculateFeatureFlagId(string key, string label)`

amerjusupovic: sounds good to me

amerjusupovic: @zhenlan do you think the extension method is unnecessary? I thought it made sense to be an extension just because it's a known conversion, but it's true that we're only using it here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the MIT license.
+//
+using System.Text;
+using System;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+    internal static class BytesExtensions
+    {
+        public static string ToBase64Url(this byte[] bytes)
```
jimmyca15, line 11: Can you put a summary that converts byte array to b64 URL. Trims trailing `=` characters and also link to b64 spec.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
jimmyca15: I would recommend keeping that extension method. b64url is a spec and encapsulates well as utility method.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
jimmyca15: The link to the spec: https://datatracker.ietf.org/doc/html/rfc4648#section-5

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// Base64 description: https://datatracker.ietf.org/doc/html/rfc4648.html#section-4
+        /// </summary>
+        public static string ToBase64Url(this byte[] bytes)
+        {
+            string bytesBase64 = Convert.ToBase64String(bytes);
+
+            int indexOfEquals = bytesBase64.IndexOf("=");
+
+            // Remove all instances of "=" at the end of the string that were added as padding
+            int stringBuilderCapacity = indexOfEquals != -1 ? indexOfEquals : bytesBase64.Length;
```
zhenlan, line 22: This code trims everything after the first occurrence of the `=`. It doesn't match your comment. You assume the `=` is at the end of the string. I'm wondering why we don't simply use `TrimEnd`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            StringBuilder stringBuilder = new StringBuilder(stringBuilderCapacity);
+
+            // Construct Base64URL string by replacing characters in Base64 conversion that are not URL safe
```
zhenlan, line 26: Is there a reason the featureFlagId needs to be URL safe?

jimmyca15: For use in REST resource IDs (/experiment/{id})

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
amerjusupovic: Base64 doesn't use equals as any value other than padding at the end, so it should work effectively the same. I had it as `TrimEnd` initially but I got lost with this other implementation that was suggested and I didn't reconsider it. I think it makes more sense to be `TrimEnd` to save iteration and be more clear. Jumped the gun on the merge, so I can make another PR.

jimmyca15: I don't see why TrimEnd would be used, it's unnecessary allocation. Otherwise it would just be `Convert.ToBase64String(bytes).Replace('+', '-').Replace('/', '_').TrimEnd('=')`. But that creates copies a bit carelessly.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
zhenlan: So, sounds like we should have a comment in CalculateFeatureFlagId to explain why we need base64url encoding.

jimmyca15: That seems more like a design doc/specification territory rather than a code comment. The code just implements the spec of a feature flag id, which uses b64url encoding.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/BytesExtensions.cs
zhenlan: Implementation is fine. The comment can be updated to clarify that we choose to skip the padding (if any) based on the spec.

PR #525, opened 2024-02-27
Fix missing feature flag status and enabled logic

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -44,9 +44,17 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
                 //if (featureFlag.Conditions?.ClientFilters == null)
                 if (featureFlag.Conditions?.ClientFilters == null || !featureFlag.Conditions.ClientFilters.Any()) // workaround since we are not yet setting client filters to null
                 {
-                    //
-                    // Always on
-                    keyValues.Add(new KeyValuePair<string, string>(featureFlagPath, true.ToString()));
+                    // Add the AlwaysOn filter instead of setting the flag to "true" so feature management doesn't skip evaluating variants and telemetry when present
+                    if (featureFlag.Variants != null || featureFlag.Telemetry != null)
```
amerjusupovic, line 48: Just want to mention that this condition could be removed so that we just always send `AlwaysOn` instead of setting the flag to `true`. It simplifies the logic and would better handle any future additions to the feature definition, but it would basically remove [the optimization](https://github.com/microsoft/FeatureManagement-Dotnet/blob/9a7f2ffcf6648426a66f3983ca15b5948ada154b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L216) in the feature management library to skip evaluating the rest of the flag if it's set to true.

rossgrambo: I prefer removing the optimization for simplicity / if we add more fields in the future, we don't have to update this.

jimmyca15: I agree on simplicity.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -44,9 +44,17 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
+                    {
+                        keyValues.Add(new KeyValuePair<string, string>($"{featureFlagPath}:{FeatureManagementConstants.EnabledFor}:{0}:{FeatureManagementConstants.Name}", FeatureManagementConstants.AlwaysOnFilter));
+                    }
+                    else
+                    {
+                        //
+                        // Always on
```
rossgrambo, line 55: Could this comment be updated to call out "featureFlagPath: true" since "Always on" sounds like the filter, which is now handled differently.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -77,10 +85,18 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
                             featureFlag.Conditions.RequirementType));
                     }
                 }
+
+                keyValues.Add(new KeyValuePair<string, string>($"{featureFlagPath}:{FeatureManagementConstants.Status}", FeatureManagementConstants.Conditional));
```
rossgrambo, line 89: Assuming we remove the "FF: true" shortcut, I prefer moving this to the top of this block since it's the most important "enabled" field.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -77,10 +85,18 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
             }
             else
             {
-                keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.SectionName}:{featureFlag.Id}", false.ToString()));
+                // Only explicitly set the flag to false if there are no variants to override the enabled state
+                if (featureFlag.Variants == null)
```
rossgrambo, line 94: I think this would need the same "Variants == null" && "Telemetry == null" logic as above.

But more importantly, I think we just remove this shortcut as well. Old FM will see no Filters and treat it as false, New FM will respect the status of Disabled

jimmyca15 (approved):

rossgrambo (approved):

PR #526, opened 2024-02-28
Merge main to preview for 8.0.0-preview release

amerjusupovic:


avanigupta:
Are we ready for v8 preview release? If so, we might want to include the changes needed to mark the existing "cache expiration" APIs as obsolete and add the new renamed APIs.

amerjusupovic:
> Are we ready for v8 preview release? If so, we might want to include the changes needed to mark the existing "cache expiration" APIs as obsolete and add the new renamed APIs.

@avanigupta That's a good point - I thought we'd deprecate those in 8.0.0 stable since we made a similar change last time in 5.0.0, but I don't think I clarified that with anyone yet.

jimmyca15 (approved):

jimmyca15:
Let's get it in the next 8.0.0 release.

avanigupta (approved):

PR #529, opened 2024-03-04
Target .NET 8 and update test packages

amerjusupovic:


amerjusupovic:
[This test](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/fd3e5387ff2ff201fccef582777b1221d260899f/tests/Tests.AzureAppConfiguration/JsonContentTypeTests.cs#L22-L46) raises a question, since it will now fail with newer versions of `Microsoft.Extensions.Configuration.Json` starting with 6.0.0. `AddJsonFile` now adds key-values that have a value of empty array or empty object as `[key]: null` instead of simply not adding the key, like app config does now. We could change this behavior in our `JsonKeyValueAdapter`, but it would break either the older/newer scenario unless we check for the package version somehow.

tests/Tests.AzureAppConfiguration.AspNetCore/Tests.AzureAppConfiguration.AspNetCore.csproj
```diff
@@ -10,14 +10,28 @@
...
+	<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net7.0'">
+	<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="7.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
+	<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0" />
+  </ItemGroup>
```
avanigupta, line 33: We could use [conditional package references](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#adding-a-packagereference-condition) to be more concise.

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
```diff
@@ -11,21 +11,35 @@
...
 
-  <ItemGroup>
-    <ProjectReference Include="..\..\src\Microsoft.Extensions.Configuration.AzureAppConfiguration\Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj" />
+  <ItemGroup Condition="'$(TargetFramework)' == 'net48'">
+	<PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.32" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+	<PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="6.0.0" />
   </ItemGroup>
```
avanigupta, line 31: We did not have this distinction of net48 vs net6 earlier. Can these frameworks still reference the same package version?

amerjusupovic: They can, I could just update the version to 8.0.0 where it is currently. I only did this because I was looking at the [target .NET 8 PR](https://github.com/microsoft/FeatureManagement-Dotnet/pull/365) in the feature management repo, but it's not necessary.

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
```diff
@@ -11,21 +11,35 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
-    <PackageReference Include="xunit" Version="2.4.1" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
+	<PackageReference Include="Moq" Version="4.20.70" />
+	<PackageReference Include="xunit" Version="2.7.0" />
+	<PackageReference Include="xunit.runner.visualstudio" Version="2.5.7">
+	  <PrivateAssets>all</PrivateAssets>
+	  <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+	</PackageReference>
```
avanigupta, line 21: Extra whitespace

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
avanigupta: what is the latest version that works for all frameworks? If the split isn't necessary, I'd vote for keeping it simple and use the same version everywhere. Its only the test project anyway.

tests/Tests.AzureAppConfiguration/JsonContentTypeTests.cs
```diff
@@ -28,20 +28,15 @@ public void JsonContentTypeTests_CompareJsonSettingsBetweenAppConfigAndJsonFile(
...
-
-            Assert.Equal(jsonSettings.Count(), appconfigSettings.Count());
+                .Build();
 
-            foreach (KeyValuePair<string, string> jsonSetting in jsonSettings)
+            foreach (KeyValuePair<string, string> jsonSetting in jsonSettings.AsEnumerable())
             {
-                KeyValuePair<string, string> appconfigSetting = appconfigSettings.SingleOrDefault(x => x.Key == jsonSetting.Key);
-                Assert.Equal(jsonSetting, appconfigSetting);
+                Assert.Equal(jsonSettings.GetSection(jsonSetting.Key).Value, appconfigSettings.GetSection(jsonSetting.Key).Value);
```
avanigupta, line 39: `GetSection` can return null.

amerjusupovic: I thought it couldn't based on [this](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.iconfiguration.getsection?view=dotnet-plat-ext-8.0#remarks).

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
amerjusupovic: 8.0.0 works when I run locally

tests/Tests.AzureAppConfiguration/JsonContentTypeTests.cs
avanigupta: You're right, I accidentally opened the wrong "[GetSection](https://learn.microsoft.com/en-us/dotnet/api/system.configuration.configuration.getsection?view=dotnet-plat-ext-8.0#returns)" :)

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
```diff
@@ -11,23 +11,22 @@
...
     <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" />
   </ItemGroup>
-
+    
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.Extensions.Configuration.AzureAppConfiguration\Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net7.0'">
-	<PackageReference Include="Microsoft.Extensions.Configuration" Version="7.0.0" />
```
avanigupta, line 28: We don't need this reference anymore?

tests/Tests.AzureAppConfiguration.Functions.Worker/Tests.AzureAppConfiguration.Functions.Worker.csproj
```diff
@@ -10,11 +10,14 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
-    <PackageReference Include="Moq" Version="4.16.0" />
-    <PackageReference Include="xunit" Version="2.4.1" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
+	<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0" />
+	<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
```
avanigupta, line 14: Extra whitespace

tests/Tests.AzureAppConfiguration.AspNetCore/Tests.AzureAppConfiguration.AspNetCore.csproj
```diff
@@ -10,11 +10,14 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
-    <PackageReference Include="Moq" Version="4.16.0" />
-    <PackageReference Include="xunit" Version="2.4.1" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
+	<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0" />
+	<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
```
avanigupta, line 14: Whitespace

examples/ConfigStoreDemo/ConfigStoreDemo.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
     <IsPackable>false</IsPackable>
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
avanigupta, line 4: We have other sample apps on .net 7. Do you mind updating everything to .net 8?

amerjusupovic: Yes missed it, will update

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
amerjusupovic: I should look into why this is happening, but the test passes now without it. It's related to these changes, since if I remove it on another branch it fails.

avanigupta: Could be because we are updating `Microsoft.Extensions.Configuration.Json` from 5.x to 8.x. That brings in `Microsoft.Extensions.Configuration` 8.x as a dependency, which works for .net 7 framework.

avanigupta (approved):

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
amerjusupovic: that was it, thanks!

avanigupta (approved):

PR #531, opened 2024-03-05
Fix bug with variant configuration value for feature management

amerjusupovic:
Currently, the provider incorrectly expects the `string` type for a variant's configuration value within a feature flag. This PR allows the provider to correctly handle all possible values for `configuration_value`.

Expected allowed values:

```JSON
{
	"id": "ExampleFlag",
	"description": "",
	"enabled": true,
	"conditions": {
		"client_filters": []
	},
	"variants": [
		{
			"name": "NumberValue",
			"configuration_value": 1
		},
		{
			"name": "StringValue",
			"configuration_value": "1"
		},
		{
			"name": "ObjectValue",
			"configuration_value": {
                                "Key1": "Value1"
                        }
		},
		{
			"name": "ArrayValue",
			"configuration_value": ["1"]
		},
		{
			"name": "BooleanValue",
			"configuration_value": true
		},
		{
			"name": "NullValue",
			"configuration_value": null
		}
	]
}
```

Currently allowed values:

```JSON
{
	"id": "ExampleFlag",
	"description": "",
	"enabled": true,
	"conditions": {
		"client_filters": []
	},
	"variants": [
		{
			"name": "StringValue",
			"configuration_value": "1"
		},
		{
			"name": "NullValue",
			"configuration_value": null
		}
	]
}
```

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
@@ -93,10 +93,17 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManage
 
                     keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.Name}", featureVariant.Name));
 
-                    foreach (KeyValuePair<string, string> kvp in new JsonFlattener().FlattenJson(featureVariant.ConfigurationValue))
+                    if (featureVariant.ConfigurationValue.ValueKind == JsonValueKind.Null)
                     {
-                        keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.ConfigurationValue}" + 
-                            (string.IsNullOrEmpty(kvp.Key) ? "" : $":{kvp.Key}"), kvp.Value));
+                        keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.ConfigurationValue}", null));
+                    }
+                    else
+                    {
+                        foreach (KeyValuePair<string, string> kvp in new JsonFlattener().FlattenJson(featureVariant.ConfigurationValue))
+                        {
+                            keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.ConfigurationValue}" +
+                                (string.IsNullOrEmpty(kvp.Key) ? "" : $":{kvp.Key}"), kvp.Value));
+                        }
                     }
 
                     if (featureVariant.ConfigurationReference != null)
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
@@ -93,17 +93,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManage
 
                     keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.Name}", featureVariant.Name));
 
-                    if (featureVariant.ConfigurationValue.ValueKind == JsonValueKind.Null)
+                    foreach (KeyValuePair<string, string> kvp in new JsonFlattener().FlattenJson(featureVariant.ConfigurationValue))
                     {
-                        keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.ConfigurationValue}", null));
-                    }
-                    else
-                    {
-                        foreach (KeyValuePair<string, string> kvp in new JsonFlattener().FlattenJson(featureVariant.ConfigurationValue))
-                        {
-                            keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.ConfigurationValue}" +
-                                (string.IsNullOrEmpty(kvp.Key) ? "" : $":{kvp.Key}"), kvp.Value));
-                        }
+                        keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.ConfigurationValue}" +
+                            (string.IsNullOrEmpty(kvp.Key) ? "" : $":{kvp.Key}"), kvp.Value));
                     }
 
                     if (featureVariant.ConfigurationReference != null)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -93,9 +93,10 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
 
                     keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.Name}", featureVariant.Name));
 
-                    if (featureVariant.ConfigurationValue != null)
+                    foreach (KeyValuePair<string, string> kvp in new JsonFlattener().FlattenJson(featureVariant.ConfigurationValue))
```
rossgrambo, line 96: I assume this works with `null`? We'll just actually get a key value pair with null as the value?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -93,9 +93,10 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
                     {
-                        keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.ConfigurationValue}", featureVariant.ConfigurationValue));
+                        keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.ConfigurationValue}" + 
+                            (string.IsNullOrEmpty(kvp.Key) ? "" : $":{kvp.Key}"), kvp.Value));
```
rossgrambo, line 99: Does this logic work with a primative value? It will as long as 
```
new JsonFlattener().FlattenJson(featureVariant.ConfigurationValue)
```
returns a kvp of `"": primative`, but that isn't obvious to me.

amerjusupovic: In the `WithVariants` unit test it does work with string, I tested with int and bool as well but didn't add it to the test. I can add it to be thorough.

jimmyca15:
What's the bug, no PR description no issue linked

amerjusupovic:
> What's the bug, no PR description no issue linked

My fault, let me add a description

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: At the moment it would have no key value pairs and just skip the loop. If you evaluate it with feature management `Variant.Configuration` would just be null either if the key exists as null or is missing entirely. It seems like having the key present set to null would be clearer though?

amerjusupovic: The way we handle json key values in the provider currently is to just exclude the key though, so could also be a question of keeping that consistent.

amerjusupovic: Kidding - it does just add the key with null value

rossgrambo: I guess `null` could mean a lot of things. The ones I was thinking about are:

null value
```json
"variants": [
  "name": "test",
  "configuration_value": null
]
```
Should have a kv pair "configuration_value" => null

Missing config_value
```json
"variants": [
  "name": "test",
  "configuration_reference": "SomethingWeHaventDoneYet"
]
```
Should not have a kv pair "configuration_value" => null

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
rossgrambo: Oh, I didn't see the test. Yeah adding the last option might not be a bad idea.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: I think that makes sense, would just need to alter the first scenario since that's where it returns empty string now. This scenario is a bit different from how we use `FlattenJson` for filter parameters since portal doesn't allow parameters to be null or omitted, so it should be ok to change that.

amerjusupovic: Yea, I can change the first scenario to return null instead of empty string. That would mean they get treated as effectively the same, where trying to get the value for that key will return null.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -93,9 +93,17 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
 
                     keyValues.Add(new KeyValuePair<string, string>($"{variantsPath}:{FeatureManagementConstants.Name}", featureVariant.Name));
 
-                    if (featureVariant.ConfigurationValue != null)
+                    if (featureVariant.ConfigurationValue.ValueKind == JsonValueKind.Null)
```
jimmyca15, line 96: Is this special case needed? We don't do it for filter parameters.

amerjusupovic: Related to what I mentioned above, but I think this is what makes more sense for this situation. Right now if `configuration_value` is `null` then it just sets the key to empty string, and feature management sets the value of `Variant.Configuration` to an empty string. I think it makes more sense to expect `null` as the value, and I didn't want to change anything with the json flattener.

jimmyca15: Discussed offline. Configuration does not differentiate null and empty string.

jimmyca15 (approved):

Issue #350, opened 2022-08-29
Proposal to rename refresh APIs 


avanigupta:
Due to some confusion around the use of the word "cache" in our API names, we could change the following APIs to better reflect their functionality:
AzureAppConfigurationRefreshOptions
- SetCacheExpiration -> `SetRefreshInterval`

FeatureFlagOptions
- CacheExpirationInterval  property -> Deprecate the property in favor of a new `SetFeatureFlagRefreshInterval` method

This would help clarify that the provided TimeSpan is only used to trigger a new refresh operation. If nothing has changed on the server, configuration values being used by the application continue to be valid even after this time has passed - values don't "expire".

The proposed name changes also align with our existing [AzureAppConfigurationKeyVaultOptions.SetSecretRefreshInterval](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.azureappconfiguration.azureappconfigurationkeyvaultoptions.setsecretrefreshinterval?view=azure-dotnet) API. 

cc @drago-draganov @zhenlan

amerjusupovic:
Completed and will be part of the next preview release

PR #532, opened 2024-03-13
Deprecate and rename refresh APIs for clarity

amerjusupovic:
See #350

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private RequestTracingOptions _requestTracingOptions;
         private Dictionary<Uri, ConfigurationClientBackoffStatus> _configClientBackoffs = new Dictionary<Uri, ConfigurationClientBackoffStatus>();
 
-        private readonly TimeSpan MinCacheExpirationInterval;
+        private readonly TimeSpan MinRefreshInterval;
 
         // The most-recent time when the refresh operation attempted to load the initial configuration
         private DateTimeOffset InitializationCacheExpires = default;
@@ -111,11 +111,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             if (watchers.Any())
             {
-                MinCacheExpirationInterval = watchers.Min(w => w.CacheExpirationInterval);
+                MinRefreshInterval = watchers.Min(w => w.RefreshInterval);
             }
             else
             {
-                MinCacheExpirationInterval = RefreshConstants.DefaultCacheExpirationInterval;
+                MinRefreshInterval = RefreshConstants.DefaultRefreshInterval;
             }
 
             // Enable request tracing if not opt-out
@@ -237,7 +237,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     {
                         if (InitializationCacheExpires < utcNow)
                         {
-                            InitializationCacheExpires = utcNow.Add(MinCacheExpirationInterval);
+                            InitializationCacheExpires = utcNow.Add(MinRefreshInterval);
 
                             await InitializeAsync(clients, cancellationToken).ConfigureAwait(false);
                         }
@@ -982,8 +982,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         private void UpdateCacheExpirationTime(KeyValueWatcher changeWatcher)
         {
-            TimeSpan cacheExpirationTime = changeWatcher.CacheExpirationInterval;
-            changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(cacheExpirationTime);
+            changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.RefreshInterval);
         }
 
         private async Task<T> ExecuteWithFailOverPolicyAsync<T>(
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -192,13 +192,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     EnsureFeatureManagementVersionInspected();
 
                     var utcNow = DateTimeOffset.UtcNow;
-                    IEnumerable<KeyValueWatcher> cacheExpiredWatchers = _options.ChangeWatchers.Where(changeWatcher => utcNow >= changeWatcher.CacheExpires);
-                    IEnumerable<KeyValueWatcher> cacheExpiredMultiKeyWatchers = _options.MultiKeyWatchers.Where(changeWatcher => utcNow >= changeWatcher.CacheExpires);
+                    IEnumerable<KeyValueWatcher> refreshableWatchers = _options.ChangeWatchers.Where(changeWatcher => utcNow >= changeWatcher.NextRefreshTime);
+                    IEnumerable<KeyValueWatcher> refreshableMultiKeyWatchers = _options.MultiKeyWatchers.Where(changeWatcher => utcNow >= changeWatcher.NextRefreshTime);
 
-                    // Skip refresh if mappedData is loaded, but none of the watchers or adapters cache is expired.
+                    // Skip refresh if mappedData is loaded, but none of the watchers or adapters are refreshable.
                     if (_mappedData != null &&
-                        !cacheExpiredWatchers.Any() &&
-                        !cacheExpiredMultiKeyWatchers.Any() &&
+                        !refreshableWatchers.Any() &&
+                        !refreshableMultiKeyWatchers.Any() &&
                         !_options.Adapters.Any(adapter => adapter.NeedsRefresh()))
                     {
                         return;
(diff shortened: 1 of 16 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -982,8 +982,7 @@ private bool IsAuthenticationError(Exception ex)
 
         private void UpdateCacheExpirationTime(KeyValueWatcher changeWatcher)
         {
-            TimeSpan cacheExpirationTime = changeWatcher.CacheExpirationInterval;
-            changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(cacheExpirationTime);
+            changeWatcher.CacheExpires = DateTimeOffset.UtcNow.Add(changeWatcher.RefreshInterval);
```
Eskibear, line 985: The name of `KeyValueWatcher.CacheExpires` remains as is. Just want to make sure if it is as expected.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -27,7 +38,26 @@ public class FeatureFlagOptions
...
+            set { _refreshInterval = value; }
+        }
+
+        /// <summary>
+        /// Sets the time after which feature flags can be refreshed.
+        /// </summary>
+        /// <param name="refreshInterval">
+        /// The minimum time after which feature flags can be refreshed. Must be greater than or equal to 1 second.
+        /// </param>
+        public FeatureFlagOptions SetFeatureFlagRefreshInterval(TimeSpan refreshInterval)
```
Eskibear, line 55: AFAIU here besides the renaming, it also removes the public getter. Is it by design?

Eskibear: Just wondering are there any existing test cases setting/getting `CacheExpirationInterval`? And do we want to also add cases for this new API?

zhenlan: I think we can just use `SetRefreshInterval`. It's under the `FeatureFlagOptions`, so it shouldn't cause any confusion.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Yea I was trying to think if it should be renamed too, but wasn't sure what made sense and CacheExpires is still accurate for our use I guess. If there's a good way to rename it while referring to refresh I think we could do that though.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
amerjusupovic: > Just wondering are there any existing test cases setting/getting `CacheExpirationInterval`? And do we want to also add cases for this new API?

Good point actually, I didn't re-update the tests when I made some changes, but it would make sense to have at least some reference the old getter and make sure the rest use the API instead of the internal property.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: How about `KeyValueWatcher.NextRefreshTime`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefreshOptions.cs
```diff
@@ -55,15 +55,27 @@ public AzureAppConfigurationRefreshOptions Register(string key, string label = L
...
+        [Obsolete("The SetCacheExpiration method will be deprecated in a future release. Please use the `SetRefreshInterval` method instead. " +
+            "Note that only the name of the method has changed, and the functionality remains the same.")]
         public AzureAppConfigurationRefreshOptions SetCacheExpiration(TimeSpan cacheExpiration)
         {
-            if (cacheExpiration < RefreshConstants.MinimumCacheExpirationInterval)
+            return SetRefreshInterval(cacheExpiration);
+        }
+
+        /// <summary>
+        /// Sets the refresh interval time for the key-values registered for refresh. Default value is 30 seconds. Must be greater than 1 second.
```
avanigupta, line 66:
```suggestion
        /// Sets the minimum time interval between consecutive refresh operations for the registered key-values. Default value is 30 seconds. Must be greater than 1 second.
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RefreshConstants.cs
```diff
@@ -8,12 +8,12 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
...
-        public static readonly TimeSpan DefaultCacheExpirationInterval = TimeSpan.FromSeconds(30);
-        public static readonly TimeSpan MinimumCacheExpirationInterval = TimeSpan.FromSeconds(1);
+        public static readonly TimeSpan DefaultRefreshInterval = TimeSpan.FromSeconds(30);
+        public static readonly TimeSpan MinimumRefreshInterval = TimeSpan.FromSeconds(1);
 
         // Feature flags
-        public static readonly TimeSpan DefaultFeatureFlagsCacheExpirationInterval = TimeSpan.FromSeconds(30);
-        public static readonly TimeSpan MinimumFeatureFlagsCacheExpirationInterval = TimeSpan.FromSeconds(1);
+        public static readonly TimeSpan DefaultFeatureFlagsRefreshInterval = TimeSpan.FromSeconds(30);
+        public static readonly TimeSpan MinimumFeatureFlagsRefreshInterval = TimeSpan.FromSeconds(1);
```
avanigupta, line 16: Can we use `FeatureFlag` (singular) in the name to match other places?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -27,7 +38,26 @@ public class FeatureFlagOptions
...
+        {
+            get { return _refreshInterval; }
+            set { _refreshInterval = value; }
+        }
+
+        /// <summary>
+        /// Sets the time after which feature flags can be refreshed.
+        /// </summary>
+        /// <param name="refreshInterval">
+        /// The minimum time after which feature flags can be refreshed. Must be greater than or equal to 1 second.
```
avanigupta, line 53: Can you mention the default too?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Sounds good to me

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RefreshConstants.cs
amerjusupovic: I was trying to think why this had an "s" anyway, is it just to show that it applies to all flags?

avanigupta: Might've been an oversight earlier. 'Secret' refresh interval is also singular.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefreshOptions.cs
```diff
@@ -55,15 +55,27 @@ public AzureAppConfigurationRefreshOptions Register(string key, string label = L
...
+            "Note that only the name of the method has changed, and the functionality remains the same.")]
         public AzureAppConfigurationRefreshOptions SetCacheExpiration(TimeSpan cacheExpiration)
         {
-            if (cacheExpiration < RefreshConstants.MinimumCacheExpirationInterval)
+            return SetRefreshInterval(cacheExpiration);
+        }
+
+        /// <summary>
+        /// Sets the minimum time interval between consecutive refresh operations for the registered key-values. Default value is 30 seconds. Must be greater than 1 second.
+        /// Any refresh operation triggered using <see cref="IConfigurationRefresher"/> will not update the value for a key unless the refresh interval has elapsed since the key was last refreshed.
```
avanigupta, line 67:
```suggestion
        /// Refresh operation triggered using <see cref="IConfigurationRefresher"/> will not make any server requests unless the refresh interval has elapsed since the key was last refreshed.
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -27,7 +38,26 @@ public class FeatureFlagOptions
         /// <summary>
         /// The time after which the cached values of the feature flags expire.  Must be greater than or equal to 1 second.
         /// </summary>
-        public TimeSpan CacheExpirationInterval { get; set; } = RefreshConstants.DefaultFeatureFlagsCacheExpirationInterval;
+        [Obsolete("The CacheExpirationInterval property will be deprecated in a future release. Please use the new `SetFeatureFlagRefreshInterval` method instead. " +
```
avanigupta, line 41: Method has been renamed to `SetRefreshInterval`

avanigupta: Can we use `nameof(SetRefreshInterval)` to avoid errors like this in future?

zhenlan: > property will be deprecated in a future release

Probably just terminology, but it sounds like the API is still not deprecated. How do you like this?

... property is deprecated and will be removed in a future release...

avanigupta (approved):

PR #538, opened 2024-03-27
Process json key-values without reflection

amerjusupovic:


amerjusupovic:
One thing to call out, we can disable reflection-based serialization in our unit tests like I added in this PR, but I don't think we can fully test if an application is compatible with AOT without a sample project that has the `<PublishAot>true</PublishAot>` MSBuild property. I could add that to an existing project or create a new one solely for that purpose. I've tested it locally as AOT and the current examples worked.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -22,58 +21,172 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
+                    string id = null;
 
-                        foreach (KeyValuePair<string, string> kvp in new JsonFlattener().FlattenJson(clientFilter.Parameters))
+                    if (root.TryGetProperty(FeatureManagementConstants.IdJsonPropertyName, out JsonElement idElement))
+                    {
+                        if (idElement.ValueKind == JsonValueKind.String)
+                        {
+                            id = idElement.GetString();
+                        }
+                        else if (idElement.ValueKind != JsonValueKind.Null)
```
jimmyca15, line 62: Is this saying that it's okay for id to be null?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -22,58 +21,172 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
-            }
-            catch (JsonException e)
-            {
-                throw new FormatException(setting.Key, e);
-            }
-
             var keyValues = new List<KeyValuePair<string, string>>();
 
-            if (featureFlag.Enabled)
+            try
```
jimmyca15, line 26: Mixing up the json deserialization with the configuration authoring is hard to follow.

I'd like to see a flow where we go from JSON to a feature flag object, and then use the feature flag object to author the configuration.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -22,58 +21,172 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
 
         public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, Logger logger, CancellationToken cancellationToken)
         {
-            FeatureFlag featureFlag;
-            try
-            {
-                 featureFlag = JsonSerializer.Deserialize<FeatureFlag>(setting.Value);
```
jimmyca15, line 28: I'd like to see this replaced with a private method `FeatureFlag GetFeatureFlag(setting.Value);`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -22,58 +21,172 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
             {
-                //if (featureFlag.Conditions?.ClientFilters == null)
-                if (featureFlag.Conditions?.ClientFilters == null || !featureFlag.Conditions.ClientFilters.Any()) // workaround since we are not yet setting client filters to null
-                {
-                    //
-                    // Always on
-                    keyValues.Add(new KeyValuePair<string, string>($"{FeatureManagementConstants.SectionName}:{featureFlag.Id}", true.ToString()));
-                }
-                else
+                using (JsonDocument document = JsonDocument.Parse(setting.Value))
```
jimmyca15, line 28: In the thread to move away from JsonSerializer the suggestion was to move to Utf8JsonReader

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: I was trying to avoid the extra steps needed, but I can see how it makes this easier to follow after it got more complex than I expected. I can just add back the files I deleted that model the feature flag.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: Makes sense, it's already a bit long and will have more in the preview branch

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: When I first was looking at how to use utf8jsonreader, it seemed like the logic would get pretty verbose for the amount of properties we have to add, so I thought it'd be easier to read/follow by using JsonDocument. It adds a little bit of abstraction but uses utf8jsonreader in its implementation. I wasn't sure what the benefits were, but if it's just for consistency reasons since we use it elsewhere then I can change it.

jimmyca15: It's to have a consistent approach to serialization/deserialization across the system.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: I was trying to model exactly what JsonSerializer was doing previously, and from what I tested it would allow you to pass null for a string property even if that was unlikely to actually happen in practice.

jimmyca15: Got it. So this is for backward compatibility.

I think we should check id first and skip the feature flag if id is null.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlag.cs
```diff
@@ -16,4 +16,4 @@ internal class FeatureFlag
         [JsonPropertyName("conditions")]
         public FeatureConditions Conditions { get; set; }
     }
-}
+}
```
jimmyca15, line 19: JsonPropertyName is unused now right?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,236 @@ public bool NeedsRefresh()
...
+                        case FeatureManagementConstants.IdJsonPropertyName:
+                            {
+                                if (reader.Read() && reader.TokenType == JsonTokenType.String)
+                                {
+                                    featureFlag.Id = reader.GetString();
+                                }
+                                else if (reader.TokenType == JsonTokenType.Null)
+                                {
+                                    // Effectively skips the feature flag if the Id is null
+                                    return null;
```
jimmyca15, line 135: This is putting business logic in the deserialization code path. I'd rather see the flag returned and the caller check if id is null or not.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,236 @@ public bool NeedsRefresh()
...
+                                {
+                                    featureFlag.Enabled = reader.GetBoolean();
+                                }
+                                else
+                                {
+                                    throw CreateFeatureFlagFormatException(
+                                        FeatureManagementConstants.EnabledJsonPropertyName,
+                                        settingKey,
+                                        reader.TokenType.ToString(),
+                                        $"{JsonTokenType.True}' or '{JsonTokenType.False}");
```
jimmyca15, line 161: The single quotes don't have matching pieces.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,236 @@ public bool NeedsRefresh()
...
+                                        FeatureManagementConstants.RequirementTypeJsonPropertyName,
+                                        settingKey,
+                                        reader.TokenType.ToString(),
+                                        JsonTokenType.String.ToString());
+                                }
+
+                                break;
+                            }
+
+                        default:
```
jimmyca15, line 312: What happens if there's unknown properties? I'm expecting a `Skip` of some kind is needed if an unknown property is hit.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,236 @@ public bool NeedsRefresh()
...
+            return featureFlag;
+        }
+
+        private FeatureConditions ParseFeatureConditions(ref Utf8JsonReader reader, string settingKey)
+        {
+            var featureConditions = new FeatureConditions();
+
+            while (reader.Read() && reader.TokenType != JsonTokenType.EndObject)
+            {
+                if (reader.TokenType == JsonTokenType.PropertyName)
```
jimmyca15, line 208: What's the reason to not use the pattern you have in the above method?

```

                    if (reader.TokenType != JsonTokenType.PropertyName)
                    {
                        continue;
                    }
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,236 @@ public bool NeedsRefresh()
...
+                                    break;
+                                }
+                                else if (reader.TokenType == JsonTokenType.StartArray)
+                                {
+                                    while (reader.Read() && reader.TokenType != JsonTokenType.EndArray)
+                                    {
+                                        if (reader.TokenType == JsonTokenType.StartObject)
+                                        {
+                                            ClientFilter clientFilter = new ClientFilter();
+
```
jimmyca15, line 227: Looks like it's worth having a `ParseClientFilter` method.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,236 @@ public bool NeedsRefresh()
...
+                                        settingKey,
+                                        reader.TokenType.ToString(),
+                                        JsonTokenType.StartObject.ToString());
+                                }
+
+                                break;
+                            }
+
+                        default:
+                            break;
```
jimmyca15, line 190: Same comment as before about unknown property names and possibly needing a skip.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -102,5 +92,34 @@ public bool NeedsRefresh()
...
             return _secretProvider.ShouldRefreshKeyVaultSecrets();
         }
+
+        private string ParseSecretReferenceUri(ConfigurationSetting setting)
+        {
+            try
+            {
+                var reader = new Utf8JsonReader(System.Text.Encoding.UTF8.GetBytes(setting.Value));
+
+                if (reader.Read() && reader.TokenType == JsonTokenType.StartObject)
```
jimmyca15, line 102: What if the read occurs and it isn't a start object? Seems there should be an exception?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -102,5 +92,34 @@ public bool NeedsRefresh()
...
+                {
+                    if (reader.Read() && reader.TokenType == JsonTokenType.PropertyName && reader.GetString() == KeyVaultConstants.SecretReferenceUriJsonPropertyName)
```
jimmyca15, line 104: Extra properties shouldn't break deserialization.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: With the error message it ends up looking correct because it has single quotes already around this string, but I can see how it looks odd here. Open to any suggestions for how to change this to be more clear.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
amerjusupovic: Right, had it in feature management adapter and forgot to match that logic here

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
amerjusupovic: Got it, so read all property names until we see the uri if it exists

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,236 @@ public bool NeedsRefresh()
...
+
+                        case FeatureManagementConstants.EnabledJsonPropertyName:
+                            {
+                                if (reader.Read() && (reader.TokenType == JsonTokenType.False || reader.TokenType == JsonTokenType.True))
+                                {
+                                    featureFlag.Enabled = reader.GetBoolean();
+                                }
+                                else
+                                {
+                                    throw CreateFeatureFlagFormatException(
```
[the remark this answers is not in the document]
amerjusupovic: Missed this while copying over from the JsonDocument logic I had, will add it back. Also seems like there should be at least one unit test for that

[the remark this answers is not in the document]
amerjusupovic: Also, we don't actually allow it on portal but the old serializer would convert it to boolean anyway so makes sense to do it here too (for backwards compatibility)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: Makes more sense that way, will move it out

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
jimmyca15: yes

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
jimmyca15: I see. Resolving.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,263 @@ public bool NeedsRefresh()
...
+                                        JsonTokenType.StartObject.ToString());
+                                }
+
+                                break;
+                            }
+
+                        default:
+                            reader.Skip();
+
+                            if (reader.TokenType == JsonTokenType.StartObject || reader.TokenType == JsonTokenType.StartArray)
```
jimmyca15, line 189: I'm a little bit confused why this is needed. Skip will position to the end of object or array at worst, and the `reader.Read()` that follows in the `while` clause should pass it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -102,5 +92,53 @@ public bool NeedsRefresh()
...
+                {
+                    if (reader.TokenType != JsonTokenType.PropertyName)
+                    {
+                        continue;
+                    }
+
+                    if (reader.GetString() == KeyVaultConstants.SecretReferenceUriJsonPropertyName)
+                    {
+                        if (reader.Read() && reader.TokenType == JsonTokenType.String)
+                        {
```
jimmyca15, line 117: Would we throw before for null?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -102,5 +92,53 @@ public bool NeedsRefresh()
...
+
+        private string ParseSecretReferenceUri(ConfigurationSetting setting)
+        {
+            try
+            {
+                var reader = new Utf8JsonReader(System.Text.Encoding.UTF8.GetBytes(setting.Value));
+
+                if (reader.Read() && reader.TokenType != JsonTokenType.StartObject)
+                {
+                    throw new FormatException(string.Format(ErrorMessages.KeyVaultSecretReferenceInvalidFormat, setting.Key));
```
avanigupta, line 104: Where are we handling FormatException?

jimmyca15: We handle it on startup [here](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/4ce9ec53988dc27d3045406601ee9961c5e1cda8/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs#L56)

But in the [TryRefreshAsync error handling code path](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/4ce9ec53988dc27d3045406601ee9961c5e1cda8/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L443) it looks like it is unhandled.

avanigupta: The first one is caught during build (while creating ConfigurationClient), so we can still fail with FormatException during startup. This would break the "optional" behavior [here](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/4ce9ec53988dc27d3045406601ee9961c5e1cda8/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L631).

jimmyca15: Ah interesting. It seems invalid feature flags break the optional promise, even in released versions.

avanigupta: You're right. Didn't realize we were always throwing FormatException for invalid feature flags.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
avanigupta: Yes, even if JsonSerializer didn't throw for null uri, we would eventually throw here:
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/4ce9ec53988dc27d3045406601ee9961c5e1cda8/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs#L45

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -102,5 +92,53 @@ public bool NeedsRefresh()
...
+                    if (reader.TokenType != JsonTokenType.PropertyName)
+                    {
+                        continue;
+                    }
+
+                    if (reader.GetString() == KeyVaultConstants.SecretReferenceUriJsonPropertyName)
+                    {
+                        if (reader.Read() && reader.TokenType == JsonTokenType.String)
+                        {
+                            return reader.GetString();
```
avanigupta, line 118: If we return here, we're not validating if the secret reference was a valid JSON object. For example, we should throw if end object token is missing.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementConstants.cs
```diff
@@ -10,5 +10,15 @@ internal class FeatureManagementConstants
         public const string SectionName = "FeatureManagement";
         public const string EnabledFor = "EnabledFor";
         public const string RequirementType = "RequirementType";
+        public const string Name = "Name";
+        public const string Parameters = "Parameters";
```
avanigupta, line 14: Are these being used anywhere?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
```diff
@@ -30,18 +30,23 @@ internal class JsonKeyValueAdapter : IKeyValueAdapter
...
 
             string rootJson = $"{{\"{setting.Key}\":{setting.Value}}}";
-            JsonElement jsonData;
+
+            List<KeyValuePair<string, string>> keyValuePairs = new List<KeyValuePair<string, string>>();
+
             try
             {
-                jsonData = JsonSerializer.Deserialize<JsonElement>(rootJson);
+                using (JsonDocument document = JsonDocument.Parse(rootJson))
```
avanigupta, line 38: Should we use `Utf8JsonReader` here instead of `JsonDocument`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,263 @@ public bool NeedsRefresh()
...
+                            {
+                                if (reader.Read() && reader.TokenType == JsonTokenType.StartObject)
+                                {
+                                    featureFlag.Conditions = ParseFeatureConditions(ref reader, settingKey);
+                                }
+                                else if (reader.TokenType == JsonTokenType.Null)
+                                {
+                                    break;
+                                }
+                                else
```
avanigupta, line 174: Can we use the same pattern as above?

```
 else if (reader.TokenType != JsonTokenType.Null)
{
    CreateFeatureFlagFormatException
}
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
avanigupta: Since the current token is a PropertyName, Skip will move to the property value.

amerjusupovic: Oh I didn't see the (originally, or after advancing) part, so I guess it already does that skip. I can remove this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
amerjusupovic: So it sounds like I can add the `FormatException` handling to `TryRefreshAsync` then

jimmyca15: I would open an issue and solve it in a separate PR. The exception propagation/handling isn't changing in this PR.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: Oh I tried to go back and use the same pattern everywhere but I missed this, I'll do that

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementConstants.cs
amerjusupovic: Nope! Removing them

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -85,6 +84,111 @@ public class FeatureManagementTests
             contentType: FeatureManagementConstants.ContentType + ";charset=utf-8",
             eTag: new ETag("c3c231fd-39a0-4cb6-3237-4614474b92c1"));
 
+        List<ConfigurationSetting> _nullOrMissingConditionsFeatureFlagCollection = new List<ConfigurationSetting>
```
avanigupta, line 87: Could you add more tests for invalid feature flags/Key vault ref/any malformed JSON values?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
amerjusupovic: ok, will do

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
amerjusupovic: Yea I just added them as I encountered things with the adapters, but I can try to be more thorough with all the scenarios I can think of

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
avanigupta: `JsonDocument` is indeed much simpler to use than `Utf8JsonReader`. Since `JsonDocument` already uses `Utf8JsonReader` in its implementation, why don't we use `JsonDocument` everywhere?

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
avanigupta: Earlier, we could rely on `JsonSerializer` to handle invalid JSON data appropriately. But now that we're going to own all that token parsing logic, we should add our tests for validating JSON structure.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
jimmyca15: JsonDocument loads and completely parses an entire JSON object. Even properties that one might not care about. 

Reader allows precise reading, skipping is available for tokens that are unnecessary. 

For feature filters parameters, their type is unknown. In this case, JsonDocument is the correct approach. We gain no benefits from reader.

avanigupta: Makes sense, thanks.

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
amerjusupovic: Added a few tests to cover some of the json scenarios mentioned in the PR comments

avanigupta:
This also looks like a good candidate for migrating to `Utf8JsonReader`:
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/a81ed763f983b910d95172efe9e56a91c5a5ba32/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/EventGridEventExtensions.cs#L38

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,244 @@ public bool NeedsRefresh()
...
+
+                    case FeatureManagementConstants.RequirementTypeJsonPropertyName:
+                        {
+                            if (reader.Read() && reader.TokenType == JsonTokenType.String)
+                            {
+                                featureConditions.RequirementType = reader.GetString();
+                            }
+                            else
+                            {
+                                throw CreateFeatureFlagFormatException(
```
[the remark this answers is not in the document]
amerjusupovic: Must've read something wrong, I'll change it

tests/Tests.AzureAppConfiguration/Tests.AzureAppConfiguration.csproj
```diff
@@ -7,6 +7,7 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\build\AzureAppConfiguration.snk</AssemblyOriginatorKeyFile>
     <IsPackable>false</IsPackable>
+    <JsonSerializerIsReflectionEnabledByDefault>false</JsonSerializerIsReflectionEnabledByDefault>
```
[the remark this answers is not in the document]
amerjusupovic: Oh that's interesting, I added it just as an automatic check for reflection. I didn't know it could prevent you from compiling though, does it mention why? Anyone could still test it locally using this or the `PublishAot` property in an example, but if it's an issue then I could remove it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -96,5 +86,244 @@ public bool NeedsRefresh()
...
+                            {
+                                break;
+                            }
+                            else if (reader.TokenType == JsonTokenType.StartArray)
+                            {
+                                while (reader.Read() && reader.TokenType != JsonTokenType.EndArray)
+                                {
+                                    if (reader.TokenType == JsonTokenType.StartObject)
+                                    {
+                                        featureConditions.ClientFilters.Add(ParseClientFilter(ref reader, settingKey));
```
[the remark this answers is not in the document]
amerjusupovic: Ok thanks, seems like maybe the latter makes more sense? Leave the parsing to the parse method and just handle the logic here

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -102,5 +92,50 @@ public bool NeedsRefresh()
...
+                            throw CreateKeyVaultReferenceException("Invalid Key Vault reference.", setting, null, null);
+                        }
+                    }
+                    else
+                    {
+                        reader.Skip();
+                    }
+                }
+            }
+            catch (JsonException e)
```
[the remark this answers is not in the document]
amerjusupovic: So probably just catch and throw `FormatException` like we do with `JsonException`?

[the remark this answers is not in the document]
avanigupta: That makes sense for feature flags. For KeyVault refs, we dont throw any FormatException, right? So just throw KeyVaultReferenceException here.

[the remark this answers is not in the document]
amerjusupovic: Oops didn't see the file - not formatexception, thanks

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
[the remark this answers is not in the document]
avanigupta: The latter could get complicated. Imagine this json:
```json
{
    ...
    ...

    "client_filters": [
        { }, 
        {   
          "Name": "foo",
          "Parameters": { ... }   
        },
        { }
    ]

}
```

I think it might be better to add a new ClientFilter to the list of ClientFilters only when Name or Parameters is non-null. I believe this is what JsonSerializer does for us currently.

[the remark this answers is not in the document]
amerjusupovic: Oh I see, I guess it makes sense this way if the serializer was already doing it, since we're still just doing parse logic by not adding empty filters. I was trying to compare it to the check for if `id` is null, but that's different.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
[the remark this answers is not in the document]
amerjusupovic: I couldn't see a way to add a test for this, not sure if there's a way to pass a string value that will generate invalid utf8 bytes after encoding.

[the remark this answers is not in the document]
amerjusupovic: Yea we do. Should I remove the catch for InvalidOperationException then? Seems like it shouldn't happen since we have checks for string type before each GetString call and we encode the bytes ourselves.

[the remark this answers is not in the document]
avanigupta: That's a great point! So we dont need to handle InvalidOperationException.

avanigupta (approved):

PR #540, opened 2024-04-02
Add variant allocation and flag telemetry information to request tracing

amerjusupovic:
### Summary

This PR aims to add information about the newest feature flag properties to request tracing. The feature flag properties include variants, allocation, and telemetry.

### Design

Currently, the following information will be added to request tracing:
1. What's the highest number of variants present in any single loaded flag?
2. Does any flag that was loaded use the seed property for allocation?
3. Is telemetry enabled for any loaded feature flags?
4. Is variant configuration reference used for any loaded feature flags?


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -16,11 +16,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManage
...
+        private FeatureFlagTracing _featureFlagTracing;
 
-        public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracing)
+        public FeatureManagementKeyValueAdapter(FeatureFlagTracing featureFilterTracing)
         {
-            _featureFilterTracing = featureFilterTracing ?? throw new ArgumentNullException(nameof(featureFilterTracing));
+            _featureFlagTracing = featureFilterTracing ?? throw new ArgumentNullException(nameof(featureFilterTracing));
         }
 
         public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, Uri endpoint, Logger logger, CancellationToken cancellationToken)
```
jimmyca15, line 26: Is this PR stale? Parsing is still being done inline with writing configuration.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagTracing.cs
```diff
@@ -63,11 +72,19 @@ public void UpdateFeatureFilterTracing(string filterName)
             }
         }
 
+        public void UpdateMaxVariants(int currentFlagTotalVariants)
```
jimmyca15, line 75:
```suggestion
        public void NotifyMaxVariants(int currentFlagTotalVariants)
```

It may not update anything.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagTracing.cs
```diff
@@ -29,18 +28,28 @@ internal class FeatureFilterTracing
...
             return UsesCustomFilter || UsesPercentageFilter || UsesTimeWindowFilter || UsesTargetingFilter;
         }
 
-        public void ResetFeatureFilterTracing()
+        public bool AnyTracingFeaturesUsed()
+        {
+            return UsesSeed || UsesTelemetry || UsesVariantConfigurationReference;
+        }
+
+        public void ResetFeatureFlagTracing()
```
jimmyca15, line 46: You need to reset maxvariants right? And what about uses telemetry and uses variant configuration reference?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: I thought I would be merging the dotnet schema PR first then resolving conflicts here, so I hadn't updated this yet. I forgot to mention that

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -30,12 +30,19 @@ internal class RequestTracingConstants
         public const string KeyVaultConfiguredTag = "UsesKeyVault";
         public const string KeyVaultRefreshConfiguredTag = "RefreshesKeyVault";
         public const string ReplicaCountKey = "ReplicaCount";
+        public const string FeatureFlagUsesTelemetryTag = "FFTelemetry";
+        public const string FeatureFlagUsesSeedTag = "FFSeed";
+        public const string FeatureFlagMaxVariantsKey = "MaxVariants";
+        public const string FeatureFlagUsesVariantConfigurationReferenceTag = "FFConfigRef";
+        public const string FeaturesKey = "Features";
```
zhenlan, line 37: Do we want to scope this to feature flags only? If so, it probably makes more sense to have "FFFeatures". On the other hand, we could drop the FF from others, so we will have, for example, FFFeatures=Seed+Telemetry+ConfigRef.

jimmyca15: It sounds reasonable to me. But that also means we have non-feature flag related items we would have `Features=` and `FFFeatures=`

amerjusupovic: Since we don't currently have anything to under `Features=`, should I just add the constant but not use it yet or is there something we want to move into this section?

jimmyca15: We're not using `Features=` at the moment. There's no reason to have it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -30,12 +30,19 @@ internal class RequestTracingConstants
         public const string KeyVaultConfiguredTag = "UsesKeyVault";
         public const string KeyVaultRefreshConfiguredTag = "RefreshesKeyVault";
         public const string ReplicaCountKey = "ReplicaCount";
+        public const string FeatureFlagUsesTelemetryTag = "Telemetry";
+        public const string FeatureFlagUsesSeedTag = "Seed";
+        public const string FeatureFlagMaxVariantsKey = "MaxVariants";
+        public const string FeatureFlagUsesVariantConfigurationReferenceTag = "ConfigRef";
```
avanigupta, line 36: Do we want to track inline config as well?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagTracing.cs
```diff
@@ -29,18 +28,31 @@ internal class FeatureFilterTracing
...
+        public bool UsesVariantConfigurationReference { get; set; } = false;
+        public int MaxVariants { get; set; }
+
         public bool UsesAnyFeatureFilter()
         {
             return UsesCustomFilter || UsesPercentageFilter || UsesTimeWindowFilter || UsesTargetingFilter;
         }
 
-        public void ResetFeatureFilterTracing()
+        public bool AnyTracingFeaturesUsed()
```
avanigupta, line 41: Suggest rename to match previous method name: `UsesAnyTracingFeature`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -16,12 +16,12 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManage
 {
     internal class FeatureManagementKeyValueAdapter : IKeyValueAdapter
     {
-        private FeatureFilterTracing _featureFilterTracing;
+        private FeatureFlagTracing _featureFlagTracing;
         private int _featureFlagIndex = 0;
 
-        public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracing)
+        public FeatureManagementKeyValueAdapter(FeatureFlagTracing featureFilterTracing)
```
avanigupta, line 22:
```suggestion
        public FeatureManagementKeyValueAdapter(FeatureFlagTracing featureFlagTracing )
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
amerjusupovic: Overall, it seemed like configuration reference was the important one for us to track since it's more of a feature while inline config is kind of the standard usage, but I could add it if it's helpful for us to track people who have no configuration on their variant at all versus inline config.

zhenlan: It doesn't feel necessary.

jimmyca15: I agree, doesn't seem necessary.

avanigupta: Ok sounds good

avanigupta (approved):

PR #556, opened 2024-05-16
Use Microsoft feature flag schema and remove use of reflection for json parsing

amerjusupovic:
Emits the feature flag schema defined [here](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json).

Added to avoid resolving conflicts with main branch, main will be merged in after to ensure necessary tests are added and branches are up to date.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -1222,7 +1222,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     if (featureManagementVersion != null)
                     {
                         // If the version is less than 3.2.0, log the schema version warning
-                        if (featureManagementVersion.Major < 3 || (featureManagementVersion.Major == 3 && featureManagementVersion.Minor <= 1))
+                        if (featureManagementVersion < new Version(3, 2, 0))
                         {
                             _logger.LogWarning(LogHelper.BuildFeatureManagementMicrosoftSchemaVersionWarningMessage());
                         }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -1192,9 +1192,23 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
                 if (_requestTracingEnabled && _requestTracingOptions != null)
                 {
-                    _requestTracingOptions.FeatureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
+                    Version featureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
 
-                    _requestTracingOptions.FeatureManagementAspNetCoreVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAspNetCoreAssemblyName);
+                    Version featureManagementAspNetCoreVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAspNetCoreAssemblyName);
+
+                    if (featureManagementVersion != null)
+                    {
+                        // If the version is less than 3.2.0, log the schema version warning
+                        if (featureManagementVersion.Major <= 3 && featureManagementVersion.Minor <= 1)
+                        {
+                            _logger.LogWarning(LogHelper.BuildFeatureManagementMicrosoftSchemaVersionWarningMessage());
+                        }
+                    }
+
+                    // Return the version using only the first 3 fields and remove additional characters
+                    _requestTracingOptions.FeatureManagementVersion = featureManagementVersion?.ToString(3).Trim('{', '}');
+
+                    _requestTracingOptions.FeatureManagementAspNetCoreVersion = featureManagementAspNetCoreVersion?.ToString(3).Trim('{', '}');
                 }
             }
         }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -556,6 +556,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             // Reset old filter tracing in order to track the filter types present in the current response from server.
             _options.FeatureFilterTracing.ResetFeatureFilterTracing();
 
+            foreach (IKeyValueAdapter adapter in _options.Adapters)
+            {
+                adapter.ResetState();
+            }
+
             foreach (KeyValuePair<string, ConfigurationSetting> kvp in data)
             {
                 IEnumerable<KeyValuePair<string, string>> keyValuePairs = null;
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -388,7 +388,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                             // Invalidate the cached Key Vault secret (if any) for this ConfigurationSetting
                             foreach (IKeyValueAdapter adapter in _options.Adapters)
                             {
-                                adapter.ProcessProviderRefresh(change.Current);
+                                adapter.OnConfigurationRefresh(change.Current);
                             }
                         }
                     }
@@ -399,7 +399,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                         // Invalidate all the cached KeyVault secrets
                         foreach (IKeyValueAdapter adapter in _options.Adapters)
                         {
-                            adapter.ProcessProviderRefresh();
+                            adapter.OnConfigurationRefresh();
                         }
 
                         // Update the next refresh time for all refresh registered settings and feature flags
@@ -734,7 +734,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 // Invalidate all the cached KeyVault secrets
                 foreach (IKeyValueAdapter adapter in _options.Adapters)
                 {
-                    adapter.ProcessProviderRefresh();
+                    adapter.OnConfigurationRefresh();
                 }
 
                 Dictionary<string, ConfigurationSetting> mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
@@ -912,6 +912,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         {
             // Set the application data for the configuration provider
             Data = data;
+
+            foreach (IKeyValueAdapter adapter in _options.Adapters)
+            {
+                adapter.OnConfigurationUpdated();
+            }
 
             // Notify that the configuration has been updated
             OnReload();
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -1222,7 +1222,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     if (featureManagementVersion != null)
                     {
                         // If the version is less than 3.2.0, log the schema version warning
-                        if (featureManagementVersion.Major <= 3 && featureManagementVersion.Minor <= 1)
+                        if (featureManagementVersion.Major < 3 || (featureManagementVersion.Major == 3 && featureManagementVersion.Minor <= 1))
                         {
                             _logger.LogWarning(LogHelper.BuildFeatureManagementMicrosoftSchemaVersionWarningMessage());
                         }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -1211,27 +1211,26 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         {
             if (!_isFeatureManagementVersionInspected)
             {
+                const string FeatureManagementMinimumVersion = "3.2.0";
+
                 _isFeatureManagementVersionInspected = true;
 
                 if (_requestTracingEnabled && _requestTracingOptions != null)
                 {
-                    Version featureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
-
-                    Version featureManagementAspNetCoreVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAspNetCoreAssemblyName);
+                    string featureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
 
                     if (featureManagementVersion != null)
                     {
                         // If the version is less than 3.2.0, log the schema version warning
-                        if (featureManagementVersion < new Version(3, 2, 0))
+                        if (Version.Parse(featureManagementVersion) < Version.Parse(FeatureManagementMinimumVersion))
                         {
                             _logger.LogWarning(LogHelper.BuildFeatureManagementMicrosoftSchemaVersionWarningMessage());
                         }
                     }
 
-                    // Return the version using only the first 3 fields and remove additional characters
-                    _requestTracingOptions.FeatureManagementVersion = featureManagementVersion?.ToString(3).Trim('{', '}');
+                    _requestTracingOptions.FeatureManagementVersion = featureManagementVersion;
 
-                    _requestTracingOptions.FeatureManagementAspNetCoreVersion = featureManagementAspNetCoreVersion?.ToString(3).Trim('{', '}');
+                    _requestTracingOptions.FeatureManagementAspNetCoreVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAspNetCoreAssemblyName);
                 }
             }
         }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -388,7 +388,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                             // Invalidate the cached Key Vault secret (if any) for this ConfigurationSetting
                             foreach (IKeyValueAdapter adapter in _options.Adapters)
                             {
-                                adapter.OnConfigurationRefresh(change.Current);
+                                adapter.OnChangeDetected(change.Current);
                             }
                         }
                     }
@@ -399,7 +399,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                         // Invalidate all the cached KeyVault secrets
                         foreach (IKeyValueAdapter adapter in _options.Adapters)
                         {
-                            adapter.OnConfigurationRefresh();
+                            adapter.OnChangeDetected();
                         }
 
                         // Update the next refresh time for all refresh registered settings and feature flags
@@ -734,7 +734,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 // Invalidate all the cached KeyVault secrets
                 foreach (IKeyValueAdapter adapter in _options.Adapters)
                 {
-                    adapter.OnConfigurationRefresh();
+                    adapter.OnChangeDetected();
                 }
 
                 Dictionary<string, ConfigurationSetting> mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
@@ -915,7 +915,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             foreach (IKeyValueAdapter adapter in _options.Adapters)
             {
-                adapter.OnConfigurationUpdated();
+                adapter.OnConfigUpdated();
             }
 
             // Notify that the configuration has been updated
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -388,7 +388,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                             // Invalidate the cached Key Vault secret (if any) for this ConfigurationSetting
                             foreach (IKeyValueAdapter adapter in _options.Adapters)
                             {
-                                adapter.InvalidateCache(change.Current);
+                                adapter.ProcessProviderRefresh(change.Current);
                             }
                         }
                     }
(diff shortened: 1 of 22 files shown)


amerjusupovic:
I'll be adding more unit tests

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
```diff
@@ -188,95 +420,85 @@ public class FeatureManagementTests
...
+	                        ""allocation"": {
+		                        ""seed"": ""13992821"",
+		                        ""default_when_disabled"": ""Small"",
+		                        ""default_when_enabled"": ""Small"",
+		                        ""user"": [
+			                        {
+				                        ""variant"": ""Big"",
+				                        ""users"": [
+					                        ""Marsha"",
+                                            ""John""
```
rossgrambo, line 451: Indentation inconsistent throughout this file. Might be worth running dotnet format on it.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -247,6 +241,981 @@ public bool NeedsRefresh()
...
+        {
+            return new FormatException(string.Format(
+                ErrorMessages.FeatureFlagInvalidJsonProperty,
+                jsonPropertyName,
+                settingKey,
+                foundJsonValueKind,
+                expectedJsonValueKind));
+        }
+
+        private FeatureFlag ParseFeatureFlag(string settingKey, string settingValue)
```
rossgrambo, line 261: I don't like us having to do this manually. Did we look into source generation as an alternative? https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/reflection-vs-source-generation?pivots=dotnet-8-0#reflection

rossgrambo (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: That's fair, it does require a lot more testing and is more code to maintain.  One reason to continue with this approach is it follows the pattern we established in our other .NET repos of always using Utf8JsonReader for reading Json, which is more efficient compared to the JsonSerializer.Deserialize method. Also, we already added it to the main branch without variants/allocation/telemetry and this PR is re-adding it for merge conflict purposes. 

I'm not sure, but from what I can tell [here](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation?pivots=dotnet-8-0) it looks like source generation isn't available in all of the versions we support, since it's only for C# 9.0 and later, and .NET 6.0 and later.

tests/Tests.AzureAppConfiguration/FeatureManagementTests.cs
amerjusupovic: Tried to do this now, looks like it doesn't work the same with strings within code. I could try to fix it manually too, just not sure if there's a better way

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1210,9 +1215,23 @@ private void EnsureFeatureManagementVersionInspected()
...
                 {
-                    _requestTracingOptions.FeatureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
+                    Version featureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
+
+                    Version featureManagementAspNetCoreVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAspNetCoreAssemblyName);
+
+                    if (featureManagementVersion != null)
+                    {
+                        // If the version is less than 3.2.0, log the schema version warning
+                        if (featureManagementVersion.Major < 3 || (featureManagementVersion.Major == 3 && featureManagementVersion.Minor <= 1))
```
zhiyuanliang-ms, line 1225: `Version` class is `IComparable` which supports `<` operator. So it is possible to do this: 
``` C#
if (featureManagementVersion < new Version(3,2,0))
```
It is fine to keep the current implementation since we have comments to explain the `if` condition. Just another option for us.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ErrorMessages.cs
```diff
@@ -7,5 +7,8 @@ internal class ErrorMessages
     {
         public const string RefreshIntervalTooShort = "The refresh interval cannot be less than {0} milliseconds.";
         public const string SecretRefreshIntervalTooShort = "The secret refresh interval cannot be less than {0} milliseconds.";
+        public const string FeatureFlagInvalidJsonProperty = "Invalid property '{0}' for feature flag. Key: '{1}'. Found type: '{2}'. Expected type: '{3}'.";
+        public const string FeatureFlagInvalidFormat = "Invalid json format for feature flag. Key: '{0}'";
```
zhiyuanliang-ms, line 11:
```suggestion
        public const string FeatureFlagInvalidFormat = "Invalid json format for feature flag. Key: '{0}'.";
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -25,30 +26,27 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
+            keyValues.Add(new KeyValuePair<string, string>($"{featureFlagPath}:{FeatureManagementConstants.Id}", featureFlag.Id));
 
-            string featureFlagPath = $"{FeatureManagementConstants.SectionName}:{featureFlag.Id}";
+            keyValues.Add(new KeyValuePair<string, string>($"{featureFlagPath}:{FeatureManagementConstants.Enabled}", featureFlag.Enabled.ToString()));
 
             if (featureFlag.Enabled)
             {
-                keyValues.Add(new KeyValuePair<string, string>($"{featureFlagPath}:{FeatureManagementConstants.Status}", FeatureManagementConstants.Conditional));
-
                 //if (featureFlag.Conditions?.ClientFilters == null)
```
zhiyuanliang-ms, line 48: Should we remove this comment? 
```
//if (featureFlag.Conditions?.ClientFilters == null)
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -25,30 +26,27 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
-                if (featureFlag.Conditions?.ClientFilters == null || !featureFlag.Conditions.ClientFilters.Any()) // workaround since we are not yet setting client filters to null
-                {
-                     keyValues.Add(new KeyValuePair<string, string>($"{featureFlagPath}:{FeatureManagementConstants.EnabledFor}:{0}:{FeatureManagementConstants.Name}", FeatureManagementConstants.AlwaysOnFilter));
-                }
-                else
+                if (featureFlag.Conditions?.ClientFilters != null && featureFlag.Conditions.ClientFilters.Any()) // workaround since we are not yet setting client filters to null
                 {
                     //
                     // Conditionally on based on feature filters
```
zhiyuanliang-ms, line 52: Should it be "Conditionally based on feature filters"?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -247,6 +241,981 @@ public bool NeedsRefresh()
...
+                ErrorMessages.FeatureFlagInvalidJsonProperty,
+                jsonPropertyName,
+                settingKey,
+                foundJsonValueKind,
+                expectedJsonValueKind));
+        }
+
+        private FeatureFlag ParseFeatureFlag(string settingKey, string settingValue)
+        {
+            FeatureFlag featureFlag = new FeatureFlag();
```
zhiyuanliang-ms, line 263:
```suggestion
            var featureFlag = new FeatureFlag();
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -247,6 +241,981 @@ public bool NeedsRefresh()
...
+                                    {
+                                        if (reader.TokenType == JsonTokenType.StartObject)
+                                        {
+                                            FeatureVariant featureVariant = ParseFeatureVariant(ref reader, settingKey);
+
+                                            if (featureVariant.Name != null)
+                                            {
+                                                variants.Add(featureVariant);
+                                            }
+                                        }
```
zhiyuanliang-ms, line 377: Do we need to throw exception here if `reader.TokenType != JsonTokenType.StartObject` while parsing variant objects?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -247,6 +241,981 @@ public bool NeedsRefresh()
...
+                                    JsonTokenType.String.ToString());
+                            }
+
+                            break;
+                        }
+
+                    case FeatureManagementConstants.From:
+                        {
+                            if (reader.Read() && 
+                                ((reader.TokenType == JsonTokenType.Number && reader.TryGetInt32(out int from)) ||
```
zhiyuanliang-ms, line 994: `from` is C# keyword.
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/

How about change it to `fromPercentile`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -247,6 +241,981 @@ public bool NeedsRefresh()
...
+                                featureTelemetry.Enabled = reader.GetBoolean();
+                            }
+                            else if (reader.TokenType == JsonTokenType.String && bool.TryParse(reader.GetString(), out bool enabled))
+                            {
+                                featureTelemetry.Enabled = enabled;
+                            }
+                            else if (reader.TokenType != JsonTokenType.Null)
+                            {
+                                throw CreateFeatureFlagFormatException(
+                                    FeatureManagementConstants.Enabled,
```
zhiyuanliang-ms, line 1159: The feature flag has `enabled` property. This should be `telemetry.enabled`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -247,6 +241,981 @@ public bool NeedsRefresh()
...
+            while (reader.Read() && reader.TokenType != JsonTokenType.EndObject)
+            {
+                if (reader.TokenType != JsonTokenType.PropertyName)
+                {
+                    continue;
+                }
+
+                string userAllocationPropertyName = reader.GetString();
+
+                switch (userAllocationPropertyName)
```
zhiyuanliang-ms, line 805: Should we handle the case that one of `variant` and `users` is missing? For example,
``` json
"allocation": {
		"user": [
			{
				"users": [
					"a",
					"b"
				]
			}
		]
	}
```
This user allocation will not take effect. Do we still need to put it in the configuration?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: Thanks! I'll change to that

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: Yes, good catch thanks

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: Good question, I think it should check to see if they aren't empty, so if there's anything inside of "user", either a list of users or a variant name, then it would be added. I thought since we shouldn't necessarily be doing any feature management logic, it would make sense to just exclude it if it's completely empty, but otherwise add the object. So in this case that you put up, I think it should be added. I don't think we need to throw an exception though, since we never did something like that if `conditions` was empty, for example.

avanigupta: Agree with Amer here. Feature Management library is responsible for checking the validity of a feature flag. Provider library just reads and flattens whatever is stored in AppConfig.

zhiyuanliang-ms (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1210,9 +1215,23 @@ private void EnsureFeatureManagementVersionInspected()
...
                 {
-                    _requestTracingOptions.FeatureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
+                    Version featureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
+
+                    Version featureManagementAspNetCoreVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAspNetCoreAssemblyName);
+
+                    if (featureManagementVersion != null)
+                    {
+                        // If the version is less than 3.2.0, log the schema version warning
+                        if (featureManagementVersion < new Version(3, 2, 0))
```
avanigupta, line 1225: Min version could be a constant variable.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -247,6 +240,995 @@ public bool NeedsRefresh()
...
+                                    reader.TokenType.ToString(),
+                                    JsonTokenType.String.ToString());
+                            }
+
+                            break;
+                        }
+
+                    case FeatureManagementConstants.ConfigurationValue:
+                        {
+                            if (reader.Read())
```
avanigupta, line 1107: Do we need to check if reader.TokenType == JsonTokenType.StartObject ?

avanigupta: And throw exception if its not an object or null

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -247,6 +240,995 @@ public bool NeedsRefresh()
...
+                            }
+
+                            break;
+                        }
+
+                    case FeatureManagementConstants.Metadata:
+                        {
+                            if (reader.Read() && reader.TokenType == JsonTokenType.StartObject)
+                            {
+                                featureTelemetry.Metadata = new Dictionary<string, string>();
```
avanigupta, line 1185: Metadata looks like a good candidate for parsing using `JsonDocument.ParseValue`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IKeyValueAdapter.cs
```diff
@@ -18,5 +18,7 @@ internal interface IKeyValueAdapter
         void InvalidateCache(ConfigurationSetting setting = null);
 
         bool NeedsRefresh();
+
+        void ResetState();
```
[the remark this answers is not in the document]
amerjusupovic: I think you're right that at least something has to change here - having the new method seems weird since it's just for this one case. I think a rename works based on where we're calling `InvalidateCache` now and where the state needs to reset for the `FeatureManagementKeyValueAdapter` anyway.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: Maybe, but from what I saw, telemetry metadata should be a dictionary with string keys and string values, and not any json object. Is there a way to validate after it gets parsed that it fits that structure?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: I don't think we should need that check, `configuration_value` can be any valid json so a number, boolean, string, etc. would all be valid.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementConstants.cs
```diff
@@ -7,36 +7,42 @@ internal class FeatureManagementConstants
...
+        public const string Users = "users";
+        public const string Groups = "groups";
+        public const string From = "from";
+        public const string To = "to";
+        public const string Seed = "seed";
+
+        // Telemetry metadata keys
         public const string ETag = "ETag";
         public const string FeatureFlagId = "FeatureFlagId";
         public const string FeatureFlagReference = "FeatureFlagReference";
```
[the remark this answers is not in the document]
amerjusupovic: Looks like feature management still emits values [in PascalCase](https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs#L41), and I confirmed with @rossgrambo.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
avanigupta: Good point

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -247,6 +240,995 @@ public bool NeedsRefresh()
...
+                                        continue;
+                                    }
+
+                                    string metadataPropertyName = reader.GetString();
+
+                                    if (reader.Read() && reader.TokenType == JsonTokenType.String)
+                                    {
+                                        featureTelemetry.Metadata[metadataPropertyName] = reader.GetString();
+                                    }
+                                    else if (reader.TokenType != JsonTokenType.Null)
```
avanigupta, line 1200: So null is not a valid value here? We might not need the null check and throw for anything thats not a string.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeaturePercentileAllocation.cs
```diff
@@ -2,19 +2,15 @@
...
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManagement
 {
     internal class FeaturePercentileAllocation
     {
-        [JsonPropertyName("variant")]
         public string Variant { get; set; }
 
-        [JsonPropertyName("from")]
-        public double From { get; set; }
+        public double? From { get; set; }
```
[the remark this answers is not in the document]
amerjusupovic: Oh good point, I was thinking to do that so we could tell if they were set later on in the code when the `percentile` object is added to the key values output, but I don't think there's a point in doing that anyway. They should just be 0 by default like it was before.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
amerjusupovic: Oh `null` is a valid string, since `JsonSerializer.Deserialize` would treat null as a valid string. In other cases for strings it wouldn't matter adding them since referencing configuration would treat it the same way, getting a missing key would just return null. For that reason, I thought it would be the same case here because adding the key or not results in the same behavior when reading the configuration. Let me know if I missed something there though

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1206,11 +1211,24 @@ private void EnsureFeatureManagementVersionInspected()
...
 
                 if (_requestTracingEnabled && _requestTracingOptions != null)
                 {
-                    _requestTracingOptions.FeatureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
+                    string featureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
+
+                    if (featureManagementVersion != null)
+                    {
+                        // If the version is less than 3.2.0, log the schema version warning
+                        if (Version.Parse(featureManagementVersion) < Version.Parse(FeatureManagementMinimumVersion))
```
avanigupta, line 1225: Both if conditions can be merged

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IKeyValueAdapter.cs
```diff
@@ -15,7 +15,9 @@ internal interface IKeyValueAdapter
 
         bool CanProcess(ConfigurationSetting setting);
 
-        void InvalidateCache(ConfigurationSetting setting = null);
+        void OnConfigurationRefresh(ConfigurationSetting setting = null);
+
+        void OnConfigurationUpdated();
```
avanigupta, line 20: How about `OnChangeDetected` and `OnConfigUpdated`?

amerjusupovic: Sure, I think that's clearer for the first one

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
```diff
@@ -99,5 +104,10 @@ public bool NeedsRefresh()
         {
             return false;
         }
+
+        public void ResetState()
+        {
+            return;
+        }
```
avanigupta, line 111: Dont need ResetState anymore

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -98,9 +98,19 @@ public void InvalidateCache(ConfigurationSetting setting = null)
...
+        {
+            return;
+        }
+
         public bool NeedsRefresh()
         {
             return _secretProvider.ShouldRefreshKeyVaultSecrets();
         }
+
+        public void ResetState()
```
avanigupta, line 111: same here

avanigupta (approved):

PR #558, opened 2024-05-30
Show full informational version for feature management library

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -79,8 +80,27 @@ public static string GetAssemblyVersion(string assemblyName)
         {
             if (!string.IsNullOrEmpty(assemblyName))
             {
-                // Return the version using only the first 3 fields and remove additional characters
-                return AppDomain.CurrentDomain.GetAssemblies().SingleOrDefault(assembly => assembly.GetName().Name == assemblyName)?.GetName().Version?.ToString(3).Trim('{', '}');
+                AssemblyInformationalVersionAttribute[] infoVersionAttributes = AppDomain.CurrentDomain.GetAssemblies().SingleOrDefault(assembly => assembly.GetName().Name == assemblyName)?
+                    .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute), false) as AssemblyInformationalVersionAttribute[];
```
avanigupta, line 84: Why dont we use `GetCustomAttribute<AssemblyInformationalVersionAttribute>()` instead?

amerjusupovic: Not sure what happened but I tried using that method originally and I just didn't see it show up, I must've been calling in the wrong place or something. Silly mistake, let me fix it

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -79,8 +80,24 @@ public static string GetAssemblyVersion(string assemblyName)
         {
             if (!string.IsNullOrEmpty(assemblyName))
             {
-                // Return the version using only the first 3 fields and remove additional characters
-                return AppDomain.CurrentDomain.GetAssemblies().SingleOrDefault(assembly => assembly.GetName().Name == assemblyName)?.GetName().Version?.ToString(3).Trim('{', '}');
+                AssemblyInformationalVersionAttribute infoVersionAttribute = AppDomain.CurrentDomain.GetAssemblies().SingleOrDefault(assembly => assembly.GetName().Name == assemblyName)?
```
avanigupta, line 83: This is getting too long. We could split it into two parts - get the assembly first. If it's not null, get the attribute and version.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -79,8 +80,24 @@ public static string GetAssemblyVersion(string assemblyName)
...
+                    .GetCustomAttribute<AssemblyInformationalVersionAttribute>();
+                
+                string informationalVersion = infoVersionAttribute?.InformationalVersion;
+
+                if (string.IsNullOrEmpty(informationalVersion))
+                {
+                    return null;
+                }
+
+                int plusIndex = informationalVersion.IndexOf('+');
```
avanigupta, line 93: could you add a short comment about why this is needed

avanigupta (approved):

PR #565, opened 2024-06-24
Test linting rule set

zhiyuanliang-ms:
The purpose of this PR is to show the changed files after applying the linting rules.


tests/Tests.AzureAppConfiguration.Functions.Worker/RefreshMiddlewareTests.cs
```diff
@@ -1,14 +1,14 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 //
+using System;
```
Eskibear, line 4: About sorting usings, current order looks like alphabetical, and you place `System` directives first. Is there any specific benefit to change it? Or a convention we didn't follow? I'm asking because the change affects almost all files. Can we just keep it as is and make the corresponding linting rule to match the current sorting strategy?

zhiyuanliang-ms: This is the default settings of .NET formatting options. 
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/dotnet-formatting-options#dotnet_sort_system_directives_first

> Can we just keep it as is and make the corresponding linting rule to match the current sorting strategy?

`dotnet_sort_system_directives_first = false` will just sort the all using directives in alphabetical order.

zhiyuanliang-ms: Discussed with Jimmy, we will follow the visual studio behavior: 
image
The `Sort Usings` function provided by visual studio has the same behavior as `dotnet_sort_system_directives_first = false` does.

Eskibear: AFAIU, the behavior of "Sort Usings" is matching with your VS settings. And if you check "place System directives first" in settings, I assume it does put System on top. 

BTW, I'm ok with either style if there's no side-effects. And I also vote for the alphabetical order because:
a) the order of usings in our code is alphabetical now, 
b) VS default behavior is alphabetical out-of-box, as the goal is to unify the code styles instead of bringing obstacles for contributors.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs
```diff
@@ -144,7 +144,7 @@ private static TimeSpan Jitter(this TimeSpan timeSpan, double ratio)
 
             var rand = new Random();
 
-            double jitter = ratio * (rand.NextDouble() * 2 - 1);
+            double jitter = ratio * ((rand.NextDouble() * 2) - 1);
```
Eskibear, line 147: `rand.NextDouble() * 2 - 1` looks good enough, the brackets looks unnecessary in `(a*b) -c` and won't cause ambiguity I believe.

So what's the related rule, is it too strict?

zhiyuanliang-ms: There are these options:

Now the rule is
dotnet_style_parentheses_in_arithmetic_binary_operators = always_for_clarity
dotnet_style_parentheses_in_relational_binary_operators = always_for_clarity
dotnet_style_parentheses_in_other_binary_operators = always_for_clarity
dotnet_style_parentheses_in_other_operators = always_for_clarity

ref: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0047-ide0048

How about this?
dotnet_style_parentheses_in_arithmetic_binary_operators = never_if_unnecessary
dotnet_style_parentheses_in_relational_binary_operators = always_for_clarity
dotnet_style_parentheses_in_other_binary_operators = always_for_clarity
dotnet_style_parentheses_in_other_operators = never_if_unnecessary

Eskibear: By changing it to "never_if_unnecessary", does it means `(a*b) -c` gets a warning? 
Can we simply remove this rule without checking it? My expectation is, no matter I write `(a * b) - c` or `a * b - c`. Because I think this case doesn't really matter much. I would expect to have "critical" rules only.

zhiyuanliang-ms: > By changing it to "never_if_unnecessary", does it means (a*b) -c gets a warning?

Yes.

> Can we simply remove this rule without checking it? 

Yes. I can suppress this rule.

PR #566, opened 2024-06-26
Support both .NET and microsoft schema for feature flags

amerjusupovic:


avanigupta:
We might want to update/remove this warning: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/c3b5ac1551cd9944db114201b35285bb5268bd33/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/LoggingConstants.cs#L36

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementConstants.cs
```diff
@@ -44,5 +44,10 @@ internal class FeatureManagementConstants
         public const string ETag = "ETag";
         public const string FeatureFlagId = "FeatureFlagId";
         public const string FeatureFlagReference = "FeatureFlagReference";
+
+        // Dotnet schema keys
```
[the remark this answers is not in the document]
amerjusupovic: I believe those were all as a result of variants, so since the feature management library is only processing flags as .NET schema if they don't include variants, these fields shouldn't be necessary. @zhiyuanliang-ms can you fact check me here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -26,13 +27,107 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
 
         public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, Uri endpoint, Logger logger, CancellationToken cancellationToken)
         {
+            _isMicrosoftSchema = false;
```
jimmyca15, line 30: I'm expecting this to be on a per flag basis.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -26,13 +27,107 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
...
+        {
+            _featureFlagIndex = 0;
+
+            return;
+        }
+
+        private List<KeyValuePair<string, string>> ProcessDotnetSchemaFeatureFlag(FeatureFlag featureFlag, ConfigurationSetting setting, Uri endpoint)
+        {
+            var keyValues = new List<KeyValuePair<string, string>>();
+
```
jimmyca15, line 76: `ProcessMicrosoftSchemaFeatureFlag` does 

```

            if (string.IsNullOrEmpty(featureFlag.Id))
            {
                return keyValues;
            }
```

I'd suggest we unify the approach.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
jimmyca15: Otherwise adding an unrelated flag that has variants will cause old flags that are working to break.

jimmyca15 (changes requested):
per flag schema state

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
avanigupta: ProcessKeyValue is invoked once per flag. So each flag will either use .NET schema or MS schema. Am I missing something?

jimmyca15: You're right. The code would function correctly. I was thrown off because instance state is being used to flow a condition that's only relevant to the method invocation. To avoid that, the decision of which schema to use should be done by the one who calls parse feature flag and does not need to be saved as instance state.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -30,9 +30,104 @@ public FeatureManagementKeyValueAdapter(FeatureFilterTracing featureFilterTracin
 
             var keyValues = new List<KeyValuePair<string, string>>();
 
+            // Check if we need to process the feature flag using the microsoft schema
+            if (featureFlag.Variants != null || featureFlag.Allocation != null || featureFlag.Telemetry != null)
```
jimmyca15, line 34: I'd like to treat null and empty the same. Similarly to how client_filters are treated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementConstants.cs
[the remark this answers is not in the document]
zhiyuanliang-ms: The field `AlwaysOn` is still being used by .NET schema. [ref](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L220)  [doc](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference?pivots=stable-version#feature-flag-declaration)

In our [preview release](https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/FeatureManager.cs#L483), we even support `On` filter which was introduced by @Amer's [PR](https://github.com/microsoft/FeatureManagement-Dotnet/pull/250/files#diff-7726d09bdcf1556c447fe12c6bd0f83385916697784034cec5d959edf8e7021aR214-R215) to support variant.

A feature flag like this:
``` json
{
  "id": "AlwaysOnFeature",
  "enabled": true
}
```
will look like this in .NET schema
``` json
"AlwaysOnFeature": {
  "EnabledFor": [
     {"Name": "AlwaysOn"}
  ]
}
```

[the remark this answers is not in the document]
zhiyuanliang-ms: But .NET schema also support declaring feature flag in this way:
``` json
"AlwaysOnFeature": true
```
That's how currently .NET provider does. [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs#L35)

jimmyca15 (approved):

avanigupta (approved):

Issue #569, opened 2024-07-02
Detect Blazor and ASP.NET applications for request tracing


amerjusupovic:
Knowing the approximate proportion of ASP.NET applications out of all customers could be helpful for understanding which features are most important to our users, and which kinds of applications are using certain features.

zhenlan:
@mrm9084 and I were discussing the flask and django usage for python applications. It's good if we can design something to capture the framework information.

zhiyuanliang-ms:
Apps utilizing Blazor are not fundamentally different from other types of ASP.NET apps. There's no fool proof method to detect whether users are using Blazor or not. According to https://github.com/dotnet/aspnetcore/issues/56633

ASP.NET Core supports mixing various UI frameworks together, so you could even have an app that utilizes both Razor Pages and Blazor. In fact, prior to .NET 8, most Blazor Server apps also used Razor Pages to host the root Blazor component. There isn't much that a class library can do to observe these types of detailed characteristics about the app. 

The best we can do is to check which assemblies have been loaded (e.g. SignalR related stuff), but that's only a heuristic and subject to change at any point in the future.

zhiyuanliang-ms:
Discussed with @jimmyca15 and @rossgrambo offline. We are going to collect whether there is `Microsoft.AspNetCore.SignalR` assembly. This assembly will be loaded as long as Blazor is unused. If we collect data showing that many people are using SignalR, it doesn't necessarily mean our users are using Blazor apps. However, if our data shows that few people are using SignalR, it definitely means the number of our users using Blazor apps won't be higher than this. Therefore, collecting information about SignalR can still provide us with some useful insights.

PR #570, opened 2024-07-10
Detect SignalR usage for request tracing

zhiyuanliang-ms:
## Why this PR?
#569 

## Visible changes
If `Microsoft.AspNetCore.SignalR` assembly can be found, "UsesSignalr" tag will be added to the correlation context header.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         public const string KeyVaultConfiguredTag = "UsesKeyVault";
         public const string KeyVaultRefreshConfiguredTag = "RefreshesKeyVault";
         public const string ReplicaCountKey = "ReplicaCount";
-        public const string SignalrUsedTag = "UsesSignalr";
+        public const string SignalRUsedTag = "UsesSignalR";
 
         public const string DiagnosticHeaderActivityName = "Azure.CustomDiagnosticHeaders";
         public const string CorrelationContextHeader = "Correlation-Context";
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -30,12 +30,15 @@ internal class RequestTracingConstants
         public const string KeyVaultConfiguredTag = "UsesKeyVault";
         public const string KeyVaultRefreshConfiguredTag = "RefreshesKeyVault";
         public const string ReplicaCountKey = "ReplicaCount";
+        public const string SignalrUsedTag = "UsesSignalR";
```
jimmyca15, line 33: Should the r be capitalized or no? Because the casing is different between the two variables.

zhiyuanliang-ms: It should be "UserSignalr" here to keep the consistency. I will update it.

amerjusupovic: It's alright now, but I noticed everything except the `SignalRAssemblyName` variable has a lowercase 'r', is there a reason why?

amerjusupovic (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
juniwang: `SignalR` with upper case R is right which stands for Realtime, not part of the word `Signal`

zhiyuanliang-ms: For pascal casing, should it be `SignalRUserdTag` or `SignalrUserdTag`?
We used `AspNetCore` instead of `ASPNETCore`.

juniwang: It's SignalRXXX. example here: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/signalr/Microsoft.Azure.Management.SignalR/src/Generated/SignalRManagementClient.cs

zhiyuanliang-ms: Updated to SignalR

Issue #274, opened 2021-08-19
Add a SetSecretClientOptions to the AzureAppConfigurationKeyVaultOptions


acgritt:
I know this was partially covered in #232 but it would be better to provide an extension to supply default SecretClientOptions for the setup. This way we could write:
```cs
options.Connect(appConfigUrl, credential)
  .ConfigureKeyVault(vault =>
  {
    vault
      .SetCredential(credential)
      .SetSecretClientOptions(clientOptions);
  });
```

Using the [AzureAppConfigurationKeyVaultOptions.Register(SecretClient)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.azureappconfiguration.azureappconfigurationkeyvaultoptions.register?view=azure-dotnet-preview) has the drawback of the client code having to know what all the key vault connections are at startup. Rather than allowing that to be masked behind the App Configuration and it just being a matter of the credential having permissions to talk to the various key vaults.

Issue #106, opened 2019-11-07
Add a new API to allow custom client options


zhenlan:
Proposed API
```cs
public AzureAppConfigurationOptions ConfigureClientOptions(Action<ConfigurationClientOptions> options)
```
This API should allow a user to overwirte the properties of `ConfigurationClientOptions` we use by default.

For example, in the case of Azure/AppConfiguration#179, the customer can do
```cs
using HttpClient client = new HttpClient(
    new HttpClientHandler()
    {
        Proxy = new WebProxy(new Uri("http://example.com"))
    });

builder.AddAzureAppConfiguration(o =>
    o.Connect(...)
     .ConfigureClientOptions(co => co.Transport = new HttpClientTransport(client));
)
```
Reference [here](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/samples/Configuration.md#configuring-a-proxy).

cc: @drago-draganov @jimmyca15

abhilasharora:
This feature has been released as a part of version `3.0.0` of the package `Microsoft.Extensions.Configuration.AzureAppConfiguration`.

PR #577, opened 2024-07-17
Allow user to set keyvault secret client option

zhiyuanliang-ms:
## Why this PR?
#274  

## Visible change

New API
`AzureAppConfigurationKeyVaultOptions.ConfigureClientOptions`

usage:
``` C#
IConfiguration config = new ConfigurationBuilder()
    {
        options.ConfigureKeyVault(kv =>
        {
            kv.SetCredential(new DefaultAzureCredential());
            kv.ConfigureClientOptions(sco => sco.Retry.MaxAttempts = 3;
        });
    })
    .Build();
```


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -31,6 +32,16 @@ public AzureAppConfigurationKeyVaultOptions SetCredential(TokenCredential creden
             return this;
         }
 
+        /// <summary>
+        /// Sets the client options used when connecting to key vaults that have no registered <see cref="SecretClient"/>
+        /// </summary>
+        /// <param name="clientOptions">Default secret client options</param>
```
amerjusupovic, line 38:
```suggestion
        /// <param name="clientOptions">Default secret client options.</param>
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -115,8 +115,13 @@ private SecretClient GetSecretClient(Uri secretUri)
                 return null;
             }
 
-            client = new SecretClient(new Uri(secretUri.GetLeftPart(UriPartial.Authority)), _keyVaultOptions.Credential);
+            client = new SecretClient(
+                new Uri(secretUri.GetLeftPart(UriPartial.Authority)),
+                _keyVaultOptions.Credential,
+                _keyVaultOptions.ClientOptions ?? new SecretClientOptions());
```
amerjusupovic, line 121: The `SecretClient` class seems to expect null to be passed for `options`, so we don't need the null check here necessarily. I don't mind it this way, but you could simplify it if you want.

zhiyuanliang-ms:
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs#L362

We have `ConfigureClientOptions` and other `ConfigureXXX` methods for `AzureAppConfigurationOptions`. However, this naming convention wasn't followed by `AzureAppConfigurationKeyVaultOptions`. We have `SetXXX` methods for `AzureAppConfigurationKeyVaultOptions`. I think `SetClientOptions` is more appropriate here. Any thought?

@jimmyca15  @amerjusupovic @avanigupta @zhenlan

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
zhiyuanliang-ms: > The SecretClient class seems to expect null to be passed for options

Yea, I know that. 
The reason I add null check here is that I want to make it more robust and don't rely on KeyVault lib to handle the null. 
Besides, if user reads our code, they will know the default option will be used without going to the code base of keyvault lib.

amerjusupovic:
> I think `SetClientOptions` is more appropriate here. What do you think?

I think part of the reason all the methods in `AzureAppConfigurationKeyVaultOptions` start with `Set` is that the methods aren't really configuring anything, it's just passing in a secret resolver, credential, or refresh interval. For example, `AzureAppConfigurationRefreshOptions.SetRefreshInterval` follows that pattern. However, I think the other `Configure` methods in the main `AzureAppConfigurationOptions` involve actually configuring an options class, which is what we're doing in the proposed `SetClientOptions`. 

I think that means `AzureAppConfigurationKeyVaultOptions.ConfigureClientOptions` would make more sense to me. That's just how I see it right now though, any other thoughts?

zhenlan:
It's not just the naming. The benefit of `ConfigureClientOptions` is that users can make necessary client option changes directly inline (#106). They don't have to create a new `SecretClientOptions` object in advance and replace the current one.

```cs
public AzureAppConfigurationOptions ConfigureClientOptions(Action<ConfigurationClientOptions> configure)
```

jimmyca15:
> It's not just the naming. The benefit of `ConfigureClientOptions` is that users can make necessary client option changes directly inline (#106). They don't have to create a new `SecretClientOptions` object in advance and replace the current one.
> 
> ```cs
> public AzureAppConfigurationOptions ConfigureClientOptions(Action<ConfigurationClientOptions> configure)
> ```

I agree here, I prefer the case where the user gets passed a ClientOptions in the callback.

So in your example the usage becomes

```
IConfiguration config = new ConfigurationBuilder()
    {
        options.ConfigureKeyVault(kv =>
        {
            kv.SetCredential(new DefaultAzureCredential());
            kv.ConfigureClientOptions(sco =>
            {
                        sco.Retry.MaxAttempts = 3;
            });
        });
    })
    .Build();
```

zhiyuanliang-ms:
@jimmyca15 @zhenlan I agree with the proposal. But my only concern is that user will not be able to set SecretClientOptions.Version through the callback pattern.

This is the implementation of `SecretClientOptions`.

``` C#
public class SecretClientOptions : ClientOptions
{
    internal const ServiceVersion LatestVersion = ServiceVersion.V7_5;

    public ServiceVersion Version { get; }

    public bool DisableChallengeResourceVerification { get; set; }

    public SecretClientOptions(ServiceVersion version = ServiceVersion.V7_5)
    {
        Version = version;
        this.ConfigureLogging();
    }
}
```

If user want to configure the keyvault version, it can only be achieved through constructor because `Version` property doesn't have a public setter.

jimmyca15:
@zhiyuanliang-ms I think it's a good call out, but I still think it's the pattern we want.

jimmyca15 (approved):

zhenlan (approved):

Issue #555, opened 2024-05-13
Add strict mode option for parsing of JSON values


amerjusupovic:
Currently, if the provider attempts to parse an invalid JSON key-value with content type `application/json`, it will default to interpreting the key-value as a string. However, this also means there are no exceptions thrown or logging to indicate that the key-value had an invalid JSON. 

A potential solution is to add an option to the provider that can be enabled to either log a warning or throw an exception when loading a key-value if it can't be parsed as a JSON and has the "application/json" content type. This would be added to the `AzureAppConfigurationOptions` class.

zhenlan:
This should fall into server side validation, especially with AI. It's dangerous to validate this during application runtime. No one asked for this. Suggest closing.

PR #579, opened 2024-07-18
Allow strict mode while parsing JSON content type

zhiyuanliang-ms:
## Why this PR?

#555


zhenlan:
Unless you change the default, do you expect anyone to use it?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
```diff
@@ -42,7 +44,13 @@ internal class JsonKeyValueAdapter : IKeyValueAdapter
             }
             catch (JsonException)
             {
-                // If the value is not a valid JSON, treat it like regular string value
+                if (ThrowOnInvalidJson)
+                {
+                    throw new FormatException($"Invalid JSON value for key '{setting.Key}' with content type '{setting.ContentType}'. Please ensure the value is properly formatted JSON.");
+                }
+
+                logger.LogWarning($"Invalid JSON value for key '{setting.Key}' with content type '{setting.ContentType}'. Treated as a string value.");
```
amerjusupovic, line 52: Maybe we can mention the new API in this warning log?

samsadsam: Why do we need to mention the API?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -444,6 +444,15 @@ public AzureAppConfigurationOptions ConfigureStartupOptions(Action<StartupOption
             return this;
         }
 
+        /// <summary>
+        /// By default, the provider will throw <see cref="FormatException"/> if a key-value with JSON content type has invalid JSON value. This method silences the behaviour and interpret the value as string instead.
```
amerjusupovic, line 448:
```suggestion
        /// By default, the provider will throw a <see cref="FormatException"/> if a key-value with JSON content type has an invalid JSON value. This method silences the behavior and interprets the value as a string instead.
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
amerjusupovic: I don't think we have to add it, at the time I was thinking about the scenario where someone gets this exception for multiple key-values and how it could be hard to discover the API to disable it, but I think the reaction to seeing this message shouldn't be just to ignore it and disable strict parsing anyway.

zhenlan:
We won't be able to release this until the next major version. No customers ask/complain about the current behavior. If someone really wants this, they can do JSON validation using the Map API.

zhenlan:
So if we all agree with the reasoning above, I would drop this and close the PR for now.

PR #582, opened 2024-07-19
Use GitHub workflow for PR builds

amerjusupovic:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+      - preview
+      - release/*
+  pull_request:
+    branches:
+      - main
+      - preview
+      - release/*
+
+permissions:
+  security-events: write
```
Eskibear, line 16: what's this for?

amerjusupovic: From what I understand, it's to allow codeql to push any issues to [this alerts page](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/582/checks?check_run_id=27765974502). I thought we might want to run codeql since it was on our other pipelines and it's supported by github.

Eskibear: > Write permissions might also be necessary if you want to automatically create or update security advisories based on the analysis results.

It makes sense to have write permission.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+        run: |
+          pwsh build.ps1
+
+      - name: Dotnet Pack
+        run: |
+          pwsh pack.ps1
+
+      - name: Dotnet Test
+        run: |
+          pwsh test.ps1 || exit 0
```
Eskibear, line 49: Does ` || exit 0` mean the task will be successful even if test fails? Is that on purpose, as I would expect test failure to fail the CI pipeline.

amerjusupovic: Yea good point, missed that

Eskibear (approved):

Issue #583, opened 2024-07-23
Add failover property to Correlation context between requests that failover


zhiyuanliang-ms:
Our failover is purely client-side. We want to provide more insight about whether failover happens by adding some information with the correlation context on the client.

PR #584, opened 2024-07-25
Add failover request

zhiyuanliang-ms:
## Why this PR?
#583 

## Visible change

Correlation context will have two more tags: `UsesReplicaDiscovery` and `FailoverRequest`

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -962,7 +962,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 IsKeyVaultConfigured = _options.IsKeyVaultConfigured,
                 IsKeyVaultRefreshConfigured = _options.IsKeyVaultRefreshConfigured,
                 ReplicaCount = _options.Endpoints?.Count() - 1 ?? _options.ConnectionStrings?.Count() - 1 ?? 0,
-                FilterTracing = _options.FeatureFilterTracing
+                FilterTracing = _options.FeatureFilterTracing,
+                IsReplicaDiscoveryEnabled = _options.ReplicaDiscoveryEnabled
             };
         }
 
@@ -1004,6 +1005,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             Uri previousEndpoint = _configClientManager.GetEndpointForClient(clientEnumerator.Current);
             ConfigurationClient currentClient;
+
+            _requestTracingOptions.IsFailoverRequest = false;
 
             while (true)
             {
@@ -1077,6 +1080,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 }
 
                 previousEndpoint = currentEndpoint;
+
+                _requestTracingOptions.IsFailoverRequest = true;
             }
         }
 
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -186,6 +186,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
             // Ensure that concurrent threads do not simultaneously execute refresh operation.
             if (Interlocked.Exchange(ref _networkOperationsInProgress, 1) == 0)
             {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1077,6 +1080,8 @@ private void UpdateCacheExpirationTime(KeyValueWatcher changeWatcher)
                 }
 
                 previousEndpoint = currentEndpoint;
+
+                _requestTracingOptions.IsFailoverRequest = true;
```
RichardChen820, line 1084: Let's consider this concrete scenario, customer passes an endpoint (static) and we discover a replica endpoint (dynamic)
1. At a time point, the static one starts to fail, failover to the dynamic one, current code can handle the dynamic client as a failover
2. The dynamic one keeps failing and it refresh before its backoff time due, I think the dynamic client still need to have the failover tag in correlation context, but the current code will treat it as not a failover

RichardChen820: And we need to consider the loadbalance feature, the load balance requests should not be considered as failover

zhiyuanliang-ms: If we don't consider loadbalancing, then we can have a very simple way to decide whether the request is from a failover: we can just check whether the current client is the first client from the list returned by `ConfigurationClientManager.GetClients`. But this way will break when load balancing is enabled.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/RequestTracingOptions.cs
```diff
@@ -51,5 +51,15 @@ internal class RequestTracingOptions
         /// Flag to indicate whether Microsoft.AspNetCore.SignalR assembly is present in the application.
         /// </summary>
         public bool IsSignalRUsed { get; set; } = false;
+
+        /// <summary>
+        /// Flag to indicate whether replica discovery is enabled.
+        /// </summary>
+        public bool IsReplicaDiscoveryEnabled { get; set; } = false;
```
RichardChen820, line 58: It's quite easy to know the count of replica discovered by the srv dns lookup, how about set `DiscoveredReplicaCount=x` rather than using a boolean to indicate enabled or not.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -223,6 +230,11 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                     }
                     );
 
+                    if (clients.Count() < totalClientCount && _requestTracingEnabled && _requestTracingOptions != null)
```
RichardChen820, line 233: Comparing the client count can't tell failover happen or not, it just tell some client's backoff time is not due.

zhiyuanliang-ms: https://github.com/Azure/AppConfiguration-DotnetProvider/pull/584#discussion_r1690967802

> Let's consider this concrete scenario, customer passes an endpoint (static) and we discover a replica endpoint (dynamic)
> 
> At a time point, the static one starts to fail, failover to the dynamic one, current code can handle the dynamic client as a failover, add failover tag in correlation context.
> The static client keeps failing and refresh trigger before its backoff time due, I think the dynamic client still need to have the failover tag in correlation context, but the current code will treat it as not a failover

For this scenario, if there is any client's backoff time is not due, it means failover happened before.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhiyuanliang-ms: @RichardChen820 I sent an email to discuss this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhiyuanliang-ms: I planned to move this PR on preview branch where we have load balancing implementated.

zhiyuanliang-ms:
Move this PR on preview branch so that we can think more clearly about how we should consider failover when load balancing is introduced.

zhiyuanliang-ms:
Move to #586

PR #586, opened 2024-07-26
Add failover & load balancing telemetry

zhiyuanliang-ms:
## Why this PR?
#583 

We want to know these things.
1) How many configuration stores have load balancing enabled
2) How many application startups were saved by failing over
3) How many refresh operations were saved by failing over

Replica auto discovery is enabled by default. So we don't track it.

The request type (startup/refresh) is already added to the request tracing. So as long as we add tag for the request if it is a failover request, we can tell whether it saved startup or refresh operation.

## Visible change

Correlation context will have two more tags: `UsesLoadBalancing` and `FailoverRequest`


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -36,6 +36,8 @@ internal class RequestTracingConstants
         public const string FeatureFlagUsesVariantConfigurationReferenceTag = "ConfigRef";
         public const string FeatureFlagFeaturesKey = "FFFeatures";
         public const string SignalRUsedTag = "UsesSignalR";
+        public const string LoadBalancingEnabledTag = "UsesLoadBalancing";
```
jimmyca15, line 39: I think we need to adopt a pattern like we did here:

https://github.com/Azure/AppConfiguration-DotnetProvider/pull/540#discussion_r1673021342

And do something like `Features=LB`.

jimmyca15: Also signal r could be reduced. Features=LB+SignalR

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1107,6 +1113,11 @@ private void UpdateNextRefreshTime(KeyValueWatcher changeWatcher)
                 }
 
                 previousEndpoint = currentEndpoint;
```
RichardChen820, line 1115: This comment really refers to the line 1110:

In what situation it reach out here with `previousEndpoint == currentEndpoint`?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
zhiyuanliang-ms: This pattern `Features=LB` looks good to me.

The signalR telemetry has been released in 7.3.0. Also signalR is not a feature from App Config. I prefer to not include SignalR in `Features`.

zhiyuanliang-ms: Do you think it is necessary to move "UsesKeyVault" and "RefreshKeyVault" here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhiyuanliang-ms: Some corner cases like if you set some identical connection strings, then there will be duplicated endpoints in the client list.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
zhiyuanliang-ms: Discussed offline. Some of the data that we put into correlation context we already use in our Power BI charts. This pattern will not include existing fields.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
RichardChen820: Okay, then do we need to put the  `_requestTracingOptions.IsFailoverRequest = true;` into this `if (previousEndpoint != currentEndpoint)` block?

zhiyuanliang-ms: I prefer to not put it into `if (previousEndpoint != currentEndpoint)` block. I think the current way is easier to read for me.

RichardChen820: It affects the result, when `previousendpoint == currentEndpoint` it still be treated as failover, a little bit odd, but I agree that it's not a big deal if it is just a super corner case.

RichardChen820:
LGTM

RichardChen820 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/RequestTracingOptions.cs
```diff
@@ -51,5 +52,50 @@ internal class RequestTracingOptions
...
+        }
+
+        /// <summary>
+        /// Returns a formatted string containing code names, indicating which tracing features are used by the application.
+        /// </summary>
+        /// <returns>Formatted string like: "LB+SignalR". If no tracing feature are used, empty string will be returned.</returns>
+        public string CreateFeaturesString()
+        {
+            var sb = new StringBuilder();
+
```
avanigupta, line 82: This class defines options, so its unusual to see helper methods here. Can we move them to TracingUtils?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -35,7 +35,10 @@ internal class RequestTracingConstants
         public const string FeatureFlagMaxVariantsKey = "MaxVariants";
         public const string FeatureFlagUsesVariantConfigurationReferenceTag = "ConfigRef";
         public const string FeatureFlagFeaturesKey = "FFFeatures";
-        public const string SignalRUsedTag = "UsesSignalR";
+        public const string FeaturesKey = "Features";
+        public const string LoadBalancingEnabledTag = "LB";
+        public const string SignalRUsedTag = "SignalR";
+        public const string FailoverRequestTag = "FailoverRequest";
```
avanigupta, line 41: To keep a check on the size of Correlation-Context header, can we shorten the tag to "Failover"?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/RequestTracingOptions.cs
zhiyuanliang-ms: Sure

zhiyuanliang-ms: We have these helper [methods](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/preview/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagTracing.cs#L90) defined in FeatureFlagTracing class. I will also move them to `TracingUtils.cs`

avanigupta: FeatureFilterTracing is a helper class for FF. It's ok to define helper methods there. But we shouldn't have methods defined in this "options" class.

amerjusupovic: Looks like we didn't move `UsesAnyTracingFeature`, I think the idea was to move that one too

zhiyuanliang-ms: I think `requestTracingoptions.UsesAnyTracingFeature()` makes more sense to me than `UsesAnyTracingFeature(requestTracingoptions)`. I am ok to move it to TracingUtils. But the difference between
calling `requestTracingOptions.FeatureFlagTracing.UsesAnyTracingFeature()` and `UsesAnyTracingFeature(requestTracingoptions) makes me so uncomfortable.

avanigupta: `UsesAnyTracingFeature` is also a helper method like the other helpers for building strings. `UsesAnyTracingFeature(requestTracingOptions)` is no different than `CreateFeaturesString(requestTracingOptions)`. But I didn't realize that we define the same UsesAnyTracingFeature method in `FeatureFlagTracing` too. 

In that case, we can consolidate the pattern in one of these ways:
1. All helper methods move to `TracingUtils`.
2. What you had earlier - leave regular KV helper methods in `RequestTracingOptions` and FF helper methods in `FeatureFlagTracing`. This makes the code in `TracingUtils` more uniform.

I'm now leaning towards your original approach because I don't like mixing FF tracing logic with KV logic :)

avanigupta (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagTracing.cs
```diff
@@ -128,14 +123,23 @@ public string CreateFiltersString()
...
+                sb.Append(RequestTracingConstants.TargetingFilter);
             }
 
             return sb.ToString();
         }
 
+        /// <summary>
+        /// Returns a formatted string containing code names, indicating which tracing features are used by feature flags.
+        /// </summary>
+        /// <returns>Formatted string like: "Seed+ConfigRef+Telemnetry". If no tracing features are used, empty string will be returned.</returns>
```
linglingye001, line 135: typo: `Telemnetry` -> `Telemetry`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -970,7 +970,8 @@ private void SetRequestTracingOptions()
                 IsKeyVaultConfigured = _options.IsKeyVaultConfigured,
                 IsKeyVaultRefreshConfigured = _options.IsKeyVaultRefreshConfigured,
                 ReplicaCount = _options.Endpoints?.Count() - 1 ?? _options.ConnectionStrings?.Count() - 1 ?? 0,
-                FeatureFlagTracing = _options.FeatureFlagTracing
+                FeatureFlagTracing = _options.FeatureFlagTracing,
+                IsLoadBalancingEnabled = _options.LoadBalancingEnabled
```
avanigupta, line 974: We missed adding this to the telemetry. @zhiyuanliang-ms could you create another PR to add telemetry when load balancing is enabled?

PR #590, opened 2024-08-21
Enable nullable reference types

maxkoshevoi:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -230,23 +227,23 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
 
-                Select(featureFlagFilter, labelFilter);
+                Select(featureFlagFilter!, labelFilter!);
 
                 _multiKeyWatchers.AppendUnique(new KeyValueWatcher
                 {
-                    Key = featureFlagFilter,
-                    Label = labelFilter,
+                    Key = featureFlagFilter!,
+                    Label = labelFilter!,
```
maxkoshevoi, line 246: Here there's no check that `featureFlagSelector.KeyFilter` and `featureFlagSelector.LabelFilter` are not null. They might be since `featureFlagSelector` could be initialized with `SnapshotName`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1071,11 +1079,11 @@ private async Task<T> ExecuteWithFailOverPolicyAsync<T>(
                     }
                 }
 
-                Uri currentEndpoint = _configClientManager.GetEndpointForClient(clientEnumerator.Current);
+                Uri currentEndpoint = _configClientManager.GetEndpointForClient(clientEnumerator.Current)!;
```
maxkoshevoi, line 1082: `IConfigurationClientManager.GetEndpointForClient` can return `null` if it doesn't find the client, but everywhere it's used return value is never checked for null (here's an example)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/JsonFlattener.cs
```diff
@@ -54,20 +56,22 @@ private void VisitJsonElement(JsonElement element)
                 case JsonValueKind.True:
                 case JsonValueKind.False:
                 case JsonValueKind.Null:
-                    _data.Add(new KeyValuePair<string, string>(_currentPath, element.ToString()));
+                    _data.Add(new KeyValuePair<string, string?>(_currentPath!, element.ToString()));
```
maxkoshevoi, line 59: `_currentPath` can be `null` here if we never execute `EnterContext` or `ExitContext`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs
```diff
@@ -275,7 +276,7 @@ private async Task RefreshFallbackClients(CancellationToken cancellationToken)
                     var targetEndpoint = new Uri($"https://{host}");
 
                     var configClient = _credential == null
-                        ? new ConfigurationClient(ConnectionStringUtils.Build(targetEndpoint, _id, _secret), _clientOptions)
+                        ? new ConfigurationClient(ConnectionStringUtils.Build(targetEndpoint, _id!, _secret!), _clientOptions)
```
maxkoshevoi, line 279: If `ConfigurationClientManager` is initialized using the constructor with `endpoints`, `_id` and `_secret` will be null here

PR #592, opened 2024-09-04
Increase default MaxRetries for Key Vault client

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -86,9 +86,18 @@ public bool ShouldRefreshKeyVaultSecrets()
...
+            foreach (KeyValuePair<string, CachedKeyVaultSecret> secret in _cachedKeyVaultSecrets)
+            {
+                if (secret.Value.LastRefreshTime + RefreshConstants.MinimumSecretRefreshInterval < DateTimeOffset.UtcNow)
+                {
+                    _cachedKeyVaultSecrets.Remove(secret.Key);
+                }
+
+                if (secret.Key == _nextRefreshKey)
+                {
+                    UpdateNextRefreshableSecretFromCache();
```
avanigupta, line 98: Do we need to update next refresh key in the for loop?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/CachedKeyVaultSecret.cs
```diff
@@ -22,11 +22,23 @@ internal class CachedKeyVaultSecret
         /// </summary>
         public int RefreshAttempts { get; set; }
 
-        public CachedKeyVaultSecret(string secretValue = null, DateTimeOffset? refreshAt = null, int refreshAttempts = 0)
+        /// <summary>
+        /// The last time this secret was reloaded from Key Vault.
+        /// </summary>
+        public DateTimeOffset? LastRefreshTime { get; set; }
```
avanigupta, line 28: Doesnt need to be nullable

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -15,7 +15,12 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     public class AzureAppConfigurationKeyVaultOptions
     {
         internal TokenCredential Credential;
-        internal SecretClientOptions ClientOptions = new SecretClientOptions();
+        internal SecretClientOptions ClientOptions = new SecretClientOptions
+        {
+            Retry = {
+                MaxRetries = int.MaxValue
```
avanigupta, line 21: How do we enforce an upper limit for the time spent on KeyVault retries?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
amerjusupovic: Oh I think I can just do it after the loop if I change the condition

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
amerjusupovic: Yea fair question, I'm not sure how to choose a time if our goal with this change is to prevent a refresh from completing without updating a secret that should've been updated if it weren't for the requests failing temporarily.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -86,9 +87,18 @@ public bool ShouldRefreshKeyVaultSecrets()
...
         public void ClearCache()
         {
-            _cachedKeyVaultSecrets.Clear();
-            _nextRefreshKey = null;
-            _nextRefreshTime = null;
+            foreach (KeyValuePair<string, CachedKeyVaultSecret> secret in _cachedKeyVaultSecrets)
+            {
+                if (secret.Value.LastRefreshTime + RefreshConstants.MinimumSecretRefreshInterval < DateTimeOffset.UtcNow)
+                {
+                    _cachedKeyVaultSecrets.Remove(secret.Key);
```
jimmyca15, line 94: Modifing an enumerable while iterating over it should be avoided.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -86,9 +87,18 @@ public bool ShouldRefreshKeyVaultSecrets()
...
+                }
+            }
+
+            if (_nextRefreshKey != null && !_cachedKeyVaultSecrets.ContainsKey(_nextRefreshKey))
+            {
+                UpdateNextRefreshableSecretFromCache();
```
jimmyca15, line 100: This seems worth calling at this point as long as `_cachedKeyVaultSecrets` has elements.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
jimmyca15: However, I do think the default retry count of 3 is a bit too low. The SDK uses an exponential backoff by strategy, so even a minor bump makes a notable difference.

I did some testing in a console app by inducing a socket exception.
A retry count of 3 was exhausted in 6 seconds.
A retry count of 6 was exhausted in 46 seconds
A retry count of 7 was exhausted in 99 seconds.

I think we could benefit by bumping this up, and I would suggest 6. The reason being that it still falls under our startuptimeout, which could be important for diagnosing an issue like if you reference a key vault domain that doesn't actually exist, given this is classified as retriable.

avanigupta: It could be 46 seconds per KVR, is that a concern?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -5,7 +5,7 @@
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions;
 using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.Net.Sockets;
```
avanigupta, line 8: What's using System.Net.Sockets?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
jimmyca15: > It could be 46 seconds per KVR, is that a concern?

I don't see it as a concern.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
amerjusupovic: Thanks, nothing apparently

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -40,6 +41,7 @@ public async Task<string> GetSecretValue(KeyVaultSecretIdentifier secretIdentifi
             string secretValue = null;
 
             if (_cachedKeyVaultSecrets.TryGetValue(key, out CachedKeyVaultSecret cachedSecret) &&
+                    (cachedSecret.SourceId == secretIdentifier.SourceId) &&
                     (!cachedSecret.RefreshAt.HasValue || DateTimeOffset.UtcNow < cachedSecret.RefreshAt.Value))
```
jimmyca15, line 45: It looks like we need to cache by source id. By the way can you give me an example of what that might look like?

amerjusupovic: Source id looks like: https://{vault name}.vault.azure.net/secrets/{secret name}

avanigupta: Do you mean we use source id + AppConfig key as cache key?

avanigupta: > Source id looks like: https://{vault name}.vault.azure.net/secrets/{secret name}

It should include the version too, right?

jimmyca15: Discussed offline w/ Avani, based on our usage of the secret provider caching by key works as expected. Resolving.

amerjusupovic: One thing I ran into, I'm not sure how that would work for calling `InvalidateCache(ConfigurationSetting)` from `RefreshAsync`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
avanigupta: Discussed with Jimmy. If the first KVR fails to resolve, we throw KeyVaultReferenceException and halt the startup. So 46 seconds is not a concern since its under the allowed startup timeout.

avanigupta (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -14,11 +14,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     /// </summary>
     public class AzureAppConfigurationKeyVaultOptions
     {
+        private const int KeyVaultMaxRetries = 6;
```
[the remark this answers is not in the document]
amerjusupovic: sounds good

Issue #380, opened 2023-02-09
Unable to inject reusable ConfigurationClient per recommendation of SDK guide.


goldsam:
[The SDK Guide, recommends reusing your ConfigurationClient](https://devblogs.microsoft.com/azure-sdk/lifetime-management-and-thread-safety-guarantees-of-azure-sdk-net-clients/#client-lifetime), however the APIs of this library prevent this by disallowing injection of a custom `ConfigurationClient` instance. The library creates a new instance of `ConfigurationClient` for each call to `ConfigurationBuilder.AddAzureAppConfiguration`, which is unduly restrictive.

zhenlan:
When `ConfigurationBuilder.AddAzureAppConfiguration` is called more than once, do you connect to the same App Configuration store or different stores?

goldsam:
I'm working on a multi-tenanted project where a distinct `IConfiguration` (and thus `ConfigurationBuilder`) is created and cached for each tenant. Depending on the environment and tenant, `ConfigurationBuilder.AddAzureAppConfiguration` is sometimes called passing the same `ConfigurationClient` instance but using different label filters.

goldsam:
@zhenlan Had time to look at this? My PR went stale. Is it going to be ignored again if I update it?

amerjusupovic:
Hi @goldsam, I'm sorry about the delayed response here. I think this looks like a valid point from the blog you linked and related issues I found. Right now, I'm just getting more context on this from the team to see if there was a reason for omitting that capability. If not, then I'll get back to you about the PR, but I'm working on it!

goldsam:
@amerjusupovic Thank you. I had actually submitted a PR which now has I conflicts. I would be happed to update it if you are willing to give it some attention afterwards. If you would review my PR and provide feedback, I would be happy to do the work to get this ready. Just let me know. Thank you very much.

goldsam:
@amerjusupovic Did you see that I had created a PR previously?

amerjusupovic:
@goldsam I saw your PR but from what I gathered so far from others, it's unfortunately not the way we want to implement this. We're keeping this in mind as a future improvement, but for now it's not supported.

carlin-q-scott:
@amerjusupovic As an alternative to @goldsam 's PR, could the options class be modified to make the client setter public so that we can inject the client ourselves? It's a smaller change and allows us users to decide what to do.

goldsam:
[...]

PR #598, opened 2024-10-08
Give the users the ability to have control over ConfigurationClient instance(s) used by the provider

samsadsam:
Addresses #380

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
@@ -1,0 +1,67 @@
+﻿using System;
+using Azure;
+using Azure.Data.AppConfiguration;
+using Azure.Core;
+using Microsoft.Extensions.Azure;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal class AzureAppConfigurationClientFactory : IAzureClientFactory<ConfigurationClient>
+    {
+        private readonly ConfigurationClientOptions _clientOptions;
+
+        private readonly TokenCredential _credential;
+
+        private readonly string _secret;
+        private readonly string _id;
+        private IDictionary<string, ConfigurationClient> _clients = new Dictionary<string, ConfigurationClient>();
+
+        public AzureAppConfigurationClientFactory(
+            string connectionString,
+            ConfigurationClientOptions clientOptions)
+        {
+            if (string.IsNullOrEmpty(connectionString))
+            {
+                throw new ArgumentNullException(nameof(connectionString));
+            }
+
+            _clientOptions = clientOptions ?? throw new ArgumentNullException(nameof(clientOptions));
+
+            _secret = ConnectionStringUtils.Parse(connectionString, ConnectionStringUtils.SecretSection);
+            _id = ConnectionStringUtils.Parse(connectionString, ConnectionStringUtils.IdSection);
+        }
+
+        public AzureAppConfigurationClientFactory(
+            TokenCredential credential,
+            ConfigurationClientOptions clientOptions)
+        {
+            _credential = credential ?? throw new ArgumentNullException(nameof(credential));
+            _clientOptions = clientOptions ?? throw new ArgumentNullException(nameof(clientOptions));
+        }
+
+        public ConfigurationClient CreateClient(string endpoint)
+        {
+            if (string.IsNullOrEmpty(endpoint))
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+
+            if (!Uri.TryCreate(endpoint, UriKind.Absolute, out Uri uriResult) || !(uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps))
+            {
+                throw new ArgumentException("Invalid host URI.");
+            }
+
+            if (!_clients.TryGetValue(endpoint, out ConfigurationClient configClient))
+            {
+                configClient = _credential == null
+                        ? new ConfigurationClient(ConnectionStringUtils.Build(new Uri(endpoint), _id, _secret), _clientOptions)
+                        : new ConfigurationClient(new Uri(endpoint), _credential, _clientOptions);
+
+                _clients[endpoint] = configClient;
+            }
+
+            return configClient;
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 throw new ArgumentNullException(nameof(endpoint));
             }
 
-            if (!Uri.TryCreate(endpoint, UriKind.Absolute, out Uri uriResult) || !(uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps))
+            if (!Uri.TryCreate(endpoint, UriKind.Absolute, out Uri uriResult))
             {
                 throw new ArgumentException("Invalid host URI.");
             }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
@@ -5,6 +5,8 @@ using Azure.Core;
 using Azure.Data.AppConfiguration;
 using Microsoft.Extensions.Azure;
 using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
```
zhiyuanliang-ms, line 1:
```suggestion
﻿// Copyright (c) Microsoft Corporation.
// Licensed under the MIT license.
//
﻿using System;
```

Missed the copyright :)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+using Azure;
+using Azure.Data.AppConfiguration;
+using Azure.Core;
+using Microsoft.Extensions.Azure;
+using System.Collections.Generic;
```
zhiyuanliang-ms, line 6: image


It seems like there are unused packages. Could you verify it? BTW, could you also run `dotnet format` in the root folder to make sure import ordering follows alphabetical order。

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
samsadsam: Thanks done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
samsadsam: Thanks done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public AzureAppConfigurationClientFactory(
+            TokenCredential credential,
+            ConfigurationClientOptions clientOptions)
+        {
+            _credential = credential ?? throw new ArgumentNullException(nameof(credential));
+            _clientOptions = clientOptions ?? throw new ArgumentNullException(nameof(clientOptions));
+        }
+
+        public ConfigurationClient CreateClient(string endpoint)
```
zhiyuanliang-ms, line 43: May I know why we choose to use endpoint here?

image

According to the api and [example](https://learn.microsoft.com/en-us/dotnet/api/overview/azure/microsoft.extensions.azure-readme?view=azure-dotnet#create-named-instances)

I think the parameter of CreateClient should be a logical name. But our current implementation will rely on this endpoint to create a real app config client.

Besides, endpoint is included in the connection string. If user creates AzureAppConfigurationClientFactory with connectionstring, then this information will be redundant.

How about change this constructor to

```public AzureAppConfigurationClientFactory(
            Url endpoint
            TokenCredential credential,
            ConfigurationClientOptions clientOptions)
```

This will be consistent with our existing [connect](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs#L309) api.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -29,35 +33,33 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
...
+                if (options.ConnectionStrings != null && options.ConnectionStrings.Any())
                 {
-                    clientManager = new ConfigurationClientManager(
-                        options.ConnectionStrings,
-                        options.ClientOptions,
-                        options.ReplicaDiscoveryEnabled,
-                        options.LoadBalancingEnabled);
+                    endpoints = options.ConnectionStrings.Select(cs => new Uri(ConnectionStringUtils.Parse(cs, ConnectionStringUtils.EndpointSection)));
+
+                    clientFactory ??= new AzureAppConfigurationClientFactory(options.ConnectionStrings.First(), options.ClientOptions);
```
amerjusupovic, line 49: I'm still thinking about how we want to handle this implementation of `AzureAppConfigurationClientFactory`. We can't pass the first connection string like this since we still support both APIs that allow passing `IEnumerable` of connection strings or endpoints, and we're not making it clear anywhere that the other connection strings are being ignored. I think this is especially an issue if someone disables replica discovery.

amerjusupovic: Of course, one option is to just add both `Connect` patterns that use `IEnumerable` to this factory.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
amerjusupovic: I think you brought up a good point about the current usage of `IAzureClientFactory`, and now I don't think it makes sense to go against the pattern for our scenario. It would probably make the most sense to create a new interface for a factory with 2 `CreateClient` APIs, one that accepts a `Uri` endpoint and one for a `string` connection string. 

@zhenlan @avanigupta @jimmyca15

samsadsam: @zhiyuanliang-ms: I agree with you that it's weird to use the endpoint as the logical name. But it's the only way to solve the problem with `IAzureClientFactory`.

> endpoint is included in the connection string

@zhiyuanliang-ms: the idea was that from one replica's connection string, we can create multiple replica clients since they all share the same secrets. In that regard, `endpoint` is not redundant since there might be multiple endpoints that map to multiple clients.

That being said, I think having a custom interface for the factory instead of `IAzureClientFactory` gives us more flexibility, and we can follow the `Connect` pattern easily.

--> We will go with a custom factory interface if there are no objections.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
samsadsam: Agree that's a flaw, we will pass all the connection strings

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
samsadsam: I just pushed a new change, the factory stores all the connection strings provided by the customer.

When trying to create a client from an endpoint given provided connection string, we try to find the one that matches the endpoint. Otherwise, we default to the first connection string (replica auto discovery scenario).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
samsadsam: I just pushed a new change, the factory stores all the connection strings provided by the customer.

When trying to create a client from an endpoint given provided connection string, we try to find the one that matches the endpoint. Otherwise, we default to the first connection string (replica auto discovery scenario).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+            }
+
+            if (!Uri.TryCreate(endpoint, UriKind.Absolute, out Uri uriResult) || !(uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps))
+            {
+                throw new ArgumentException("Invalid host URI.");
+            }
+
+            if (_credential != null)
+            {
+                return new ConfigurationClient(new Uri(endpoint), _credential, _clientOptions);
```
avanigupta, line 56:
```suggestion
                return new ConfigurationClient(uriResult, _credential, _clientOptions);
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+        private readonly TokenCredential _credential;
+        private readonly IEnumerable<string> _connectionStrings;
+
+        public AzureAppConfigurationClientFactory(
+            IEnumerable<string> connectionStrings,
+            ConfigurationClientOptions clientOptions)
+        {
+            if (connectionStrings == null || !connectionStrings.Any())
+            {
+                throw new ArgumentException(nameof(connectionStrings));
```
avanigupta, line 26: ArgumentNullException

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        public ConfigurationClient CreateClient(string endpoint)
+        {
+            if (string.IsNullOrEmpty(endpoint))
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+
+            if (!Uri.TryCreate(endpoint, UriKind.Absolute, out Uri uriResult) || !(uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps))
```
avanigupta, line 49: Why do we validate scheme?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
samsadsam: done thanks

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationClientFactory.cs
samsadsam: Our store endpoints have the `https` scheme, so I thought it should be validated

samsadsam: done, removed it

avanigupta (approved):

PR #600, opened 2024-10-21
Add allocation id when telemetry is enabled for feature flag

samsadsam:


rossgrambo:
Ty Sami! This looks good to me. I compared it directly to python and confirmed they'll spit out the same thing- once python adds key sorting.

Could you add a test case for invalid json? We should throw saying the configuration_value of the variant was not valid json

rossgrambo (approved):

PR #601, opened 2024-10-22
add connect Cdn API

samsadsam:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/EmptyTokenCredential.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    /// <summary>
+    /// A token credential that provides an empty token.
+    /// </summary>
+    public class EmptyTokenCredential : TokenCredential
```
avanigupta, line 11: internal, no need to expose this implementation.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -301,6 +301,20 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
             return this;
         }
 
+        /// <summary>
+        /// Connect the provider to CDN endpoint.
+        /// </summary>
+        /// <param name="cdnEndpoint">The endpoint of the Azure App Configuration CDN to connect to.</param>
```
avanigupta, line 307:
```suggestion
        /// <param name="cdnEndpoint">The endpoint of the CDN instance to connect to.</param>
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/EmptyTokenCredential.cs
samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -301,6 +301,20 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
             return this;
         }
 
+        /// <summary>
+        /// Connect the provider to CDN endpoint.
+        /// </summary>
+        /// <param name="cdnEndpoint">The endpoint of the CDN instance to connect to.</param>
+        public AzureAppConfigurationOptions ConnectCdn(Uri cdnEndpoint)
```
avanigupta, line 308: We allow users to specify multiple `Connect` calls and the last one always wins. 
We are preserving that behavior with `ConnectCdn`, but with CDN endpoint involved, it's not that intuitive. I think we should throw if Connect and ConnectCdn are invoked together. This should not be allowed:

```
options.Connect(appConfigEndpoint, tokenCredential)
       .ConnectCdn(cdnEndpoint)
```

@zhenlan @drago-draganov what do you think?

samsadsam: What is "not that intuitive" ?

avanigupta: `Connect` is loading from AppConfig service, but `ConnectCdn` is loading from CDN service. With a setup like above, I would expect my provider to load data from both endpoints.

zhenlan: Agree with Avani's suggestion - the Connect and ConnectCdn shouldn't be allowed to be used at the same time.

Even if there is such a scenario, users can create two separate configuration builders.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -318,6 +323,27 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
...
+            {
+                throw new InvalidOperationException("Cannot connect to both Azure App Configuration and CDN at the same time.");
+            }
+
+            if (endpoint == null)
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+
+            ClientOptions.AddPolicy(new CdnApiVersionPolicy(), HttpPipelinePosition.PerCall);
```
zhiyuanliang-ms, line 342: https://learn.microsoft.com/en-us/dotnet/api/azure.core.httppipelineposition?view=azure-dotnet
Did you verify that setting the position to be `PerCall` can work?

`PerCall` doesn't work in JS sdk. I need to use `PerRetry` instead.

avanigupta: Should be done per retry

avanigupta:
Note: Correlation-Context header should contain a special tag when `ConnectCdn` is used.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
samsadsam: Yes you are right, changed to per retry

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
samsadsam: Done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/CdnApiVersionPolicy.cs
```diff
@@ -0,0 +1,55 @@
...
+    public class CdnApiVersionPolicy : HttpPipelinePolicy
+    {
+        /// <summary>
+        /// Processes the HTTP message by adding the API version query parameter.
+        /// </summary>
+        /// <param name="message">The HTTP message to process.</param>
+        /// <param name="pipeline">The pipeline of HTTP policies to apply.</param>
+        public override void Process(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
+        {
+            message.Request.Uri.Reset(AlterApiVersion(message.Request.Uri.ToUri()));
```
amerjusupovic, line 25: Seems like we can follow the pattern used in [`UserAgentHeaderPolicy`](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/90835cf5efbfb26d6de6068e4c31c3050947b78a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/UserAgentHeaderPolicy.cs#L17) so we don't have to repeat the same logic in `Process`/`ProcessAsync`

samsadsam: In order to not repeat one line (which I consider completely fine), I would need to add another wrapper method that is basically an if-else statement with a boolean.

I would say that pattern is overkill and not the simplest solution

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/CdnConfigurationClientManager.cs
```diff
@@ -0,0 +1,72 @@
+﻿using Azure.Data.AppConfiguration;
```
amerjusupovic, line 1: nit: add mit license comment

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -355,6 +372,35 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
...
 
+        /// <summary>
+        /// Connect the provider to CDN endpoint.
+        /// </summary>
+        /// <param name="endpoint">The endpoint of the CDN instance to connect to.</param>
+        public AzureAppConfigurationOptions ConnectCdn(Uri endpoint)
+        {
+            if (IsCdnEnabled)
+            {
+                throw new InvalidOperationException("Please call ConnectCdn() only once.");
```
amerjusupovic, line 383: nit: could we use `nameof` for the method name?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Cdn/ConfigurationClientManager.cs
```diff
@@ -0,0 +1,75 @@
...
+using Azure.Data.AppConfiguration;
+using Microsoft.Extensions.Azure;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Cdn
+{
+    internal class ConfigurationClientManager : IConfigurationClientManager
+    {
+        private readonly IList<ConfigurationClientWrapper> _clients;
```
amerjusupovic, line 13: Do we think it makes sense to have multiple endpoints/clients for CDN? If not, maybe we can change the list to just a single client wrapper for clarity.

samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -314,7 +326,13 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
                         }
                         else
                         {
-                            refreshAll = await RefreshIndividualKvWatchers(
+                            if (_options.IsCdnEnabled)
+                            {
+                                _configVersion ??= CalculateCacheConsistencyToken(_watchedIndividualKvs.Select(kvp => kvp.Value.ETag.ToString()));
```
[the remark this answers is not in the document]
samsadsam: when there is no change, then the configuration version will still be the same, so no need to recalculate it.

We only recalculate the config version when we detect a change.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1000,7 +1050,14 @@ await CallWithRequestTracing(
...
                     {
-                        return true;
+                        return change.Current.ETag.ToString();
+                    }
+
+                    if (_options.IsCdnEnabled)
+                    {
+                        //
+                        // even if the change is not refresh all, we still need to reset stale config version.
+                        _configVersion = null;
```
jimmyca15, line 1060: I'm a bit confused as to what this is used for. It appears to be redundant with `_options.CacheConsistencyTokenAccessor.Current`

samsadsam: There is _ffCollectionVersion as well

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1328,33 +1395,56 @@ private void UpdateClientBackoffStatus(Uri endpoint, bool successful)
...
+                    // If we have a changed ETag, we can stop checking further selectors
+                    return changedEtag;
                 }
             }
 
-            return haveCollectionsChanged;
+            return changedEtag;
+        }
+
+        private static string CalculateHash(IEnumerable<string> etags)
```
drago-draganov: The name is misleading. This is not a hash function. It uses hash to do something else. Perhaps CalculateEtag fits better?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1328,33 +1395,56 @@ private void UpdateClientBackoffStatus(Uri endpoint, bool successful)
...
+
+            //
+            // purpose
+            inputBuilder.Append("etags\n");
+
+            foreach (string etag in etags)
+            {
+                inputBuilder.Append(etag);
+                inputBuilder.Append('\n');
+            }
```
drago-draganov, line 1434: What about stability? The same elements in different order will produce different result. I don't think it's intended for the purpose of this.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1328,33 +1395,56 @@ private void UpdateClientBackoffStatus(Uri endpoint, bool successful)
...
+            StringBuilder inputBuilder = new StringBuilder();
+
+            //
+            // purpose
+            inputBuilder.Append("etags\n");
+
+            foreach (string etag in etags)
+            {
+                inputBuilder.Append(etag);
+                inputBuilder.Append('\n');
```
drago-draganov, line 1433: We need to port this between libraries. I assume it needs to be documented. The last '\n' seems harder to justify.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
samsadsam: This is outdated, the name already changed

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
samsadsam: Jimmy has comment on this, will push the fix

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
samsadsam: The only way I see for the hash to be different is if a client calls the Select() multiple times, and doesn’t do the calling in the same order.

like app1:
Select(key1*)
Select(key2*)

and app2:
Select(key2*)
Select(key1*)

samsadsam: Don’t mind sorting btw

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1328,33 +1386,56 @@ private void UpdateClientBackoffStatus(Uri endpoint, bool successful)
...
+            StringBuilder inputBuilder = new StringBuilder();
+
+            //
+            // purpose
+            inputBuilder.Append("CacheConsistency\n");
+
+            foreach (string etag in etags)
+            {
+                inputBuilder.Append(etag);
+                inputBuilder.Append('\n');
```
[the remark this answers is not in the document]
samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
samsadsam: I sort etags before now, done.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -324,22 +342,40 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
...
+                            if (_options.IsCdnEnabled)
+                            {
+                                //
+                                // Break cdn cache
+                                _options.CacheConsistencyTokenAccessor.Current = refreshAllChangedEtag;
+
+                                // 
+                                // Reset versions so that next watch request will not use stale versions.
+                                _configVersion = null;
+                                _ffCollectionVersion = null;
```
zhiyuanliang-ms, line 362: ffCollectionVersion is set to null again in Line 402

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -314,7 +326,13 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
                         }
                         else
                         {
-                            refreshAll = await RefreshIndividualKvWatchers(
+                            if (_options.IsCdnEnabled)
+                            {
```
zhiyuanliang-ms, line 330: Is it possible for a deleted watched kv to have null etag. Then `kvp.Value.ETag.ToString()` will be a null reference exception.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -56,10 +56,17 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
...
                 else
                 {
-                    throw new ArgumentException($"Please call {nameof(AzureAppConfigurationOptions)}.{nameof(AzureAppConfigurationOptions.Connect)} to specify how to connect to Azure App Configuration.");
+                    throw new ArgumentException($"Please call {nameof(AzureAppConfigurationOptions)}.{nameof(AzureAppConfigurationOptions.Connect)} or {nameof(AzureAppConfigurationOptions)}.{nameof(AzureAppConfigurationOptions.ConnectCdn)} to specify how to connect to Azure App Configuration.");
                 }
 
-                provider = new AzureAppConfigurationProvider(new ConfigurationClientManager(clientFactory, endpoints, options.ReplicaDiscoveryEnabled, options.LoadBalancingEnabled), options, _optional);
+                if (options.IsCdnEnabled)
+                {
+                    provider = new AzureAppConfigurationProvider(new Cdn.ConfigurationClientManager(clientFactory, endpoints.First()), options, _optional);
```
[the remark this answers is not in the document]
samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Cdn/ConfigurationClientManager.cs
```diff
@@ -0,0 +1,73 @@
...
+// Licensed under the MIT license.
+//
+using Azure.Data.AppConfiguration;
+using Microsoft.Extensions.Azure;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Cdn
+{
+    internal class ConfigurationClientManager : IConfigurationClientManager
```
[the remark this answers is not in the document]
samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -324,22 +342,40 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
...
                             // or if any key-value collection change was detected.
                             kvEtags = new Dictionary<KeyValueSelector, IEnumerable<MatchConditions>>();
                             ffEtags = new Dictionary<KeyValueSelector, IEnumerable<MatchConditions>>();
                             ffKeys = new HashSet<string>();
 
+                            if (_options.IsCdnEnabled)
+                            {
+                                //
+                                // Break cdn cache
+                                _options.CacheConsistencyTokenAccessor.Current = refreshAllChangedEtag;
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Should we calculate hash for this single etag?

I think it is more easy to follow if we change the implementation to:

1. Record all page etags or etags of sentinel keys after the initial load
2. Before each refresh(watch) request, re-calculate the cache break string based on the current etags hash
3. During the refresh, if there is any etag change, update the recorded etags. Trigger a re-load.
4. Re-calculate the cache break string and before send the load request.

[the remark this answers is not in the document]
zhiyuanliang-ms: @avanigupta When there is any change on the etag, we will trigger a re-load operation. 

Case 1 (page monitoring refresh): 

After initial load, we have
pageEtag1, pageEtag2, pageEtag3

For the first refresh operation, we will add `hash(pageEtag1+pageEtag2+ pageEtag3)` to the request url

During the refresh, we find pageEtag2 changed to pageEtag2_new

Should we send a re-load request with `pageEtag2_new` / `hash(pageEtag2_new)` or `hash(pageEtag1+pageEtag2_new+ pageEtag3)`?

Case 2 (sentinel key base refresh):

Let's say we have sentinel keys: key1, key2, key3

After initial load, we have
key1_etag, key2_etag, key3_etag

For the first refresh operation, we will add `hash(key1_etag+key2_etag+key3_etag)` to the request url

i) During the refresh, we find key2's etag changed to key2_etag_new

Should we send a re-load request with `key2_etag_new` / `hash(key2_etag_new)` or `hash(key1_etag+key2_etag_new+key3_etag)`

ii) During the refresh, we find key2 is deleted

Should we send a re-load request with `key2_etag` (previous etag) / `hash(key2_etag)` or `hash(key1_etag+key3_etag)`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -355,6 +372,35 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
             return this;
         }
 
+        /// <summary>
+        /// Connect the provider to CDN endpoint.
+        /// </summary>
+        /// <param name="endpoint">The endpoint of the CDN instance to connect to.</param>
+        public AzureAppConfigurationOptions ConnectCdn(Uri endpoint)
+        {
+            if (IsCdnEnabled)
```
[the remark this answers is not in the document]
samsadsam: We do not care about this stuff when cdn is enabled. We don’t need to throw since we don’t use this variable, we completely ignore it

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
samsadsam: Here we are triggering a reload that also reloads feature flags and then calling “return”, skipping ff refresh logic.

Does it make sense now?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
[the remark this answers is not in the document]
samsadsam: Outdated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
samsadsam: No need for this hashing, resolving

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
samsadsam: I disabled load balancing and replica discovery when we are connecting to a cdn.

but again, there is no side effect to just let them enabled.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
samsadsam: Outdated, resolving

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
jimmyca15: > I mean we should throw exception, when people set ReplicaDiscoveryEnabled = true.

Isn't replica discovery enabled by default though? I'm all for throwing if the user opts-in to things that don't make sense when using CDN, like perhaps load balancing, but since replica discovery is enabled by default I wouldn't want the user to have to disable it explicitly when connecting to cdn. It's extra work.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -39,6 +40,11 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
         private Dictionary<Uri, ConfigurationClientBackoffStatus> _configClientBackoffs = new Dictionary<Uri, ConfigurationClientBackoffStatus>();
         private DateTimeOffset _nextCollectionRefreshTime;
 
+        #region Cdn
+        private string _configVersion = null;
```
jimmyca15, line 44: All we should need is a single etag. Doesn't matter if it's configuration or a feature flag.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
samsadsam: Done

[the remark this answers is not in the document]
avanigupta: @jimmyca15 What if load balancing is enabled after calling ConnectCdn?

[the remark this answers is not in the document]
samsadsam: Agree, need to add another check in src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs when creating provider class.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1000,7 +1047,17 @@ await CallWithRequestTracing(
 
                     if (kvWatcher.RefreshAll)
                     {
-                        return true;
+                        string changedEtag = change.Current?.ETag.ToString();
```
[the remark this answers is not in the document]
samsadsam: If there is a change (look at the if change != none above), and the current kv is null then it must be a delete scenario.

also, it’s really a fallback.

I’d say we should stay with fallback pattern.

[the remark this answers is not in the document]
samsadsam: Actually, your suggestion yields cleaner code, will do!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
avanigupta: Agreed, then we dont need the check in this method

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -355,6 +371,35 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
...
+
+            if (endpoint == null)
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+
+            var result = Connect(new List<Uri>() { endpoint }, new EmptyTokenCredential());
+
+            CdnTokenAccessor = new CdnTokenAccessor();
+            ClientOptions.AddPolicy(new CdnPolicy(CdnTokenAccessor), HttpPipelinePosition.PerCall);
```
avanigupta, line 398: Shouldn't this policy be invoked `PerRetry`?

zhiyuanliang-ms: @avanigupta I remembered that we discussed this before. Could you search "perCall" in our chat history?

Put some contexts here:
1. Our runner is using perCall policy to modify the api version in url [ref](https://msazure.visualstudio.com/Azure%20AppConfig/_search?action=contents&text=perCall&type=code&lp=code-Project&filters=ProjectFilters%7BAzure%20AppConfig%7DRepositoryFilters%7BRunner%7D&pageSize=25&result=DefaultCollection%2FAzure%20AppConfig%2FRunner%2FGBdev%2F%2Fsrc%2FMicrosoft.AppConfig.TestInfra%2FCommon%2FUtils.cs)
2. In Azure JS SDK,  the perRetry policy will be put after Sign phase: [ref](https://github.com/Azure/azure-sdk-for-js/blob/808942c902aef4df80a1361d944c4f6481a42699/sdk/core/ts-http-runtime/src/client/getClient.ts#L60)
Which means it will modify the url after HMAC, which will cause HMAC not match the actual url. But for our CDN case, we are using entra id authentication. So it is unrelated. But I think the request url is persistent across retries. I forget whether I verified this. But we should check it.
3. I remembered that the suggestion from SDK team is to use `perCall`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -314,7 +325,12 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
                         }
                         else
                         {
-                            refreshAll = await RefreshIndividualKvWatchers(
+                            if (_options.IsCdnEnabled)
+                            {
+                                _options.CdnTokenAccessor.Current = _configVersion;
+                            }
+
+                            refreshAllChangedEtag = await RefreshIndividualKvWatchers(
```
jimmyca15, line 333: This code is quite difficult to wrap my head around. The RefreshIndividualKvWatchers method is doing so much. refreshing key values, checking if anything is a refresh all and short-circuiting, and returning a value which is an etag of what changed but only if it's a refresh all kv.

It seems to much to put into a helper method.

I don't think it will be too much to bring some of it up and refactor a bit. What do you think

```
            List<KeyValueChange> keyValueChanges = new();

            foreach (KeyValueWatcher kvWatcher in refreshableIndividualKvWatchers)
            {
                KeyValueChange change = await CheckForChange(kvWatcher, cancellationToken).ConfigureAwait(false);

                //
                // Skip if no change detected
                if (change.ChangeType == KeyValueChangeType.None)
                {
                    logDebugBuilder.AppendLine(LogHelper.BuildKeyValueReadMessage(change.ChangeType, change.Key, change.Label, endpoint.ToString()));

                    continue;
                }

                logDebugBuilder.AppendLine(LogHelper.BuildKeyValueReadMessage(change.ChangeType, change.Key, change.Label, endpoint.ToString()));

                logInfoBuilder.AppendLine(LogHelper.BuildKeyValueSettingUpdatedMessage(change.Key));

                keyValueChanges.Add(change);

                if (kvWatcher.RefreshAll)
                {
                    bool refreshAll = true;

                    cdnToken = change.GetCdnToken();

                    break;
                }
            }
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: Thanks for the reminder. If the runner is using PerCall policy to overwrite api-version in url, the same should work for adding etag to the url.

jimmyca15:
done reviewing.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -355,6 +371,35 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
...
 
+        /// <summary>
+        /// Connect the provider to CDN endpoint.
+        /// </summary>
+        /// <param name="endpoint">The endpoint of the CDN instance to connect to.</param>
+        public AzureAppConfigurationOptions ConnectCdn(Uri endpoint)
+        {
+            if (IsCdnEnabled)
+            {
+                throw new InvalidOperationException($"Please call {nameof(AzureAppConfigurationOptions.ConnectCdn)} only once.");
```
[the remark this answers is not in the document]
samsadsam: thanks for suggestion, done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Cdn/CdnConfigurationClientManager.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Azure.Data.AppConfiguration;
+using Microsoft.Extensions.Azure;
+using System;
+using System.Collections.Generic;
+using System.Linq;
```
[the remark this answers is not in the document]
samsadsam: thanks, done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
samsadsam: The first pipeline that was executed per service call will be triggered. The url will already have the string when retries happen. No need to call the pipeline unnecessarily on retries.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -355,6 +371,29 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
...
+            {
+                throw new InvalidOperationException("Cannot connect to both Azure App Configuration and CDN at the same time.");
+            }
+
+            if (endpoint == null)
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+
+            var result = Connect(new List<Uri>() { endpoint }, new EmptyTokenCredential());
```
avanigupta, line 390: The second call to ConnectCdn will throw, right?
We can set Credential, Endpoints and ConnectionStrings here instead of calling Connect overload.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -206,6 +206,11 @@ private static string CreateCorrelationContextHeader(RequestType requestType, Re
                 correlationContextTags.Add(RequestTracingConstants.PushRefreshTag);
             }
 
+            if (requestTracingOptions.IsCdnEnabled)
```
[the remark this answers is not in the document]
samsadsam: thanks, like the new pattern, done.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
samsadsam: bug fixed

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
samsadsam: done!!!!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -56,10 +58,25 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
...
+                {
+                    if (options.LoadBalancingEnabled)
+                    {
+                        throw new InvalidOperationException("Load balancing is not supported for CDN endpoint.");
+                    }
+
+                    options.CdnTokenAccessor = new CdnTokenAccessor();
+                    options.ClientOptions.AddPolicy(new CdnPolicy(options.CdnTokenAccessor), HttpPipelinePosition.PerCall);
+
+                    provider = new AzureAppConfigurationProvider(new CdnConfigurationClientManager(clientFactory, endpoints.First()), options, _optional);
```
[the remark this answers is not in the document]
samsadsam: it is not about this, it is about re-using the same ConfigurationClient instance when calling the same cdn endpoint with parallel apps. I think we should still offer this functionality.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
samsadsam: done

samsadsam:
> Can we add some tests?

done @avanigupta

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -16,6 +16,7 @@
 using System.Net.Http;
 using System.Net.Sockets;
 using System.Security;
+using System.Security.Cryptography;
```
avanigupta, line 19: not needed

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -309,28 +315,94 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
...
+                                    if (_options.IsCdnEnabled)
+                                    {
+                                        //
+                                        // Break cdn cache
+                                        string token = change.GetCdnToken();
+                                        _options.CdnTokenAccessor.Current = token;
+
+                                        // 
+                                        // Reset versions so that next watch request will not use stale versions.
+                                        _configVersion = token;
```
jimmyca15, line 399: I suppose that should be _configCdnToken now

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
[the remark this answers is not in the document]
jimmyca15: Can a user shoot themselves in the foot easily if they have their own custom client factory and are using CDN? I suppose the question is, what care does a user need to take if they were to use client factory and also want to use cdn?

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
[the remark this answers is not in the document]
avanigupta: It the cache breaking policy that would be missing from custom clients. If someone wants to use client factory with cdn, their requests will not contain etag in the query params.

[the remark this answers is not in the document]
samsadsam: Your concern is valid Avani. If a customer uses a custom client factory, they have to add somehow `CdnPolicy` in policy pipeline when creating  `ConfigurationClient` for cdn scenario to work.

We should expose the type `CdnPolicy`, and recommend to always consider `AzureAppConfigurationOptions.ClientOptions` when creating custom `ConfigurationClient`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -348,17 +420,52 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
...
+                            if (changedPage != null)
+                            {
+                                ffCollectionUpdated = true;
+
+                                if (_options.IsCdnEnabled)
+                                {
+                                    //
+                                    // Break cdn cache
+                                    string token = changedPage.GetCdnToken();
+                                    _options.CdnTokenAccessor.Current = token;
```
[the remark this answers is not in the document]
samsadsam: which line?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Cdn/CdnPolicy.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        private static Uri AddTokenToUri(Uri uri, string token)
+        {
+            Debug.Assert(!string.IsNullOrEmpty(token));
+
+            var uriBuilder = new UriBuilder(uri);
+
+            var query = HttpUtility.ParseQueryString(uriBuilder.Query);
+            query["_"] = token;
```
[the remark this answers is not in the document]
samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -154,6 +155,16 @@ internal IEnumerable<IKeyValueAdapter> Adapters
...
 
+        /// <summary>
+        /// An accessor for current token to be used for CDN cache breakage/consistency.
+        /// </summary>
+        internal ICdnTokenAccessor CdnTokenAccessor { get; set; }
+
+        /// <summary>
+        /// Gets a value indicating whether CDN is enabled.
+        /// </summary>
+        internal bool IsCdnEnabled { get; private set; } = false;
```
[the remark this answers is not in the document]
samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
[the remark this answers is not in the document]
samsadsam: we use it to reload all feature flags, because a change was detected. Resolving

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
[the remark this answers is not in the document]
avanigupta: Then we'd also have to expose CdnTokenAccessor. And preferably tell users to setup their client with EmptyTokenCredentials (this is not mandatory but an optimization since auth flow can be avoided).

Btw, this made me realize that CdnTokenAccessor is used like a string and we may not need the extra class and interface to support this. Just define a string called `CurrentCdnToken` in AzureAppConfigurationOptions to replace CdnTokenAccessor.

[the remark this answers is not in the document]
samsadsam: I don't find it quiet intuitive to have `CdnPolicy` depending on `AzureAppConfigurationOptions` if I implement your suggestion.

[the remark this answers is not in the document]
samsadsam: I pushed my solution with tests as well, please check this [commit](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/601/commits/48802690bacffc2758612c6b436bd4d73faea20a)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -34,13 +36,29 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
...
+                    {
+                        throw new InvalidOperationException("Load balancing is not supported for CDN endpoint.");
+                    }
+
+                    if (clientFactory != null && !options.IsClientOptionsProvided)
+                    {
+                        throw new InvalidOperationException($"Please provide the optional param {nameof(ConfigurationClientOptions)} when calling {nameof(AzureAppConfigurationOptions.SetClientFactory)} with CDN.");
+                    }
+
+                    options.ClientOptions.AddPolicy(new CdnPolicy(options.CdnTokenAccessor), HttpPipelinePosition.PerCall);
```
avanigupta, line 53: The policy is only added to `options.ClientOptions`. These options are always ignored when `clientFactory` is not null. See line 73. So any client created using custom client factory will not have access to the cdn cache break policy.

samsadsam: I set the client options to the external one, when custom client factory is used.

please check SetClientFactory

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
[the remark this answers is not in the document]
jimmyca15: I think before we do any implementing, we need to agree on the interface that we want to expose. It's the most important part of shipping anything. I don't particularly like the interface change here to have optional options when setting configuration client factory. It looks like an afterthought to me, and doesn't show any clear connection to cdn. It doesn't seem obvious that I need to pass the options when I'm using CDN, and I get no benefit when passing them otherwise.

This is what I'm thinking. 


```
////
// Option 1: Avoid exposing Pipeline Policy, mutable object. Enable future customization of client options for cdn
////

//
// My client factory uses a single options
builder.AddAzureAppConfiguration(o =>
{
    o.ConnectCdn();

    ConfigurationClientOptions configurationClientOptions = new();

    o.SetupCdn(configurationClientOptions);

    o.SetClientFactory(
        new MyClientFactory(configurationClientOptions));
});

//
// My client factory uses different client options
builder.AddAzureAppConfiguration(o =>
{
    o.ConnectCdn();

    o.SetClientFactory(
        new MyClientFactory(
            o.SetupCdn)); // Take the cdn configuration method to be able to setup options for cdn
});
```


```
////
// Option 2: Expose cdn policy and allow users to inject it into their pipeline as needed
////

//
// My client factory uses a single options
builder.AddAzureAppConfiguration(o =>
{
    o.ConnectCdn();

    ConfigurationClientOptions configurationClientOptions = new();

    o.SetupCdn(configurationClientOptions);

    o.SetClientFactory(
        new MyClientFactory(o.CdnPolicy));
});
```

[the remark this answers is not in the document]
jimmyca15: @avanigupta ?

jimmyca15:
@samsadsam, @avanigupta led a portal design review recently in which it came up that we should use Azure Front Door as API terminology instead of CDN. There were a couple of reasons.
* Azure CDN is being retired
* We don't support generic CDN, just Azure Front Door

So instead of `ConnectCdn` as the public method, I believe this will need to become `ConnectAzureFrontDoor`. Any summaries/supporting APIs would need an update as well.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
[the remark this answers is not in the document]
zhenlan: +1 to what @avanigupta said.

We're using a client designed for one service (App Configuration) to talk to another (Azure Front Door CDN), which is already a bit of a workaround/hack. I'm unsure how well we can support a custom App Configuration client factory in this context, given the unconventional setup.

We also have the ConfigureClientOptions API, but it's unclear to what extent Azure Front Door will respect those settings. Our stance may end up being: "_If it works, great; if not, there's not much we can do._" Therefore, I'm hesitant to support using custom clients with CDN.

On the other hand, I don't think the custom App Configuration client factory has too many users (we probably should collect telemetry), so I don't think we miss too much anyway even if we don't support them for CDN.

jimmyca15:
> > So instead of `ConnectCdn` as the public method, I believe this will need to become `ConnectAzureFrontDoor`. Any summaries/supporting APIs would need an update as well.
> 
> Any preference between `ConnectAzureFrontDoor` and `ConnectAzureFrontDoorCdn`? AFD has many other things than CDN.

According to their docs, it looks like the CDN part is implied because "Azure Front Door **is** Microsoft's advanced cloud Content Delivery Network (CDN)"

image

In that case I'd say we go with `ConnectAzureFrontDoor`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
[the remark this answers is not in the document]
samsadsam: done

jimmyca15:
> > Looks good, the only remaining work is to rename cdn->afd everywhere.
> 
> I don’t think we need to do this for internal classes, I think cdn is better than afd as we can support other CDN providers in the future with same internal classes.
> 
> My understanding is that we only need to do this for the public apis
> 
> cc @jimmyca15

If it's easier to understand by linking the two with afd, then I'm fine with it.

avanigupta:
> > > Looks good, the only remaining work is to rename cdn->afd everywhere.
> > 
> > 
> > I don’t think we need to do this for internal classes, I think cdn is better than afd as we can support other CDN providers in the future with same internal classes.
> > My understanding is that we only need to do this for the public apis
> > cc @jimmyca15
> 
> If it's easier to understand by linking the two with afd, then I'm fine with it.

My thinking is that if we were starting from scratch today, knowing we’re adding a new API called `ConnectAzureFrontDoor`, and commit to supporting AFD (not generic CDN) - would we still name classes/interfaces with CDN?

tests/Tests.AzureAppConfiguration/RefreshTests.cs
avanigupta, line 1: Empty file? This and TestHelper.cs in the same folder

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -373,6 +390,33 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
             return this;
         }
 
+        /// <summary>
+        /// Connect the provider to Azure Front Door endpoint.
+        /// </summary>
+        /// <param name="endpoint">The endpoint of the Azure Front Door AFD instance to connect to.</param>
```
avanigupta, line 396:
```suggestion
        /// <param name="endpoint">The endpoint of the Azure Front Door instance to connect to.</param>
```

avanigupta (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
samsadsam: Done intentionally to tell the user that going forward the code assumes that reader knows afd=azure front door

tests/Tests.AzureAppConfiguration/RefreshTests.cs
samsadsam: Copilot hallucinated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
jimmyca15: But also it doesn't look like AFD is used anywhere in this summary after the acronym definition. Was it intended to denote further usage somewhere else?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -34,13 +36,29 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
             {
                 AzureAppConfigurationOptions options = _optionsProvider();
 
+                IAzureClientFactory<ConfigurationClient> clientFactory = options.ClientFactory;
+
+                if (options.IsAfdEnabled)
+                {
+                    if (options.LoadBalancingEnabled)
+                    {
+                        throw new InvalidOperationException("Load balancing is not supported for AFD endpoint.");
```
jimmyca15, line 45:
```suggestion
                        throw new InvalidOperationException("Load balancing is not supported when connecting to AFD.");
```

For consistency.

jimmyca15: Also, I would just type out the whole Azure Front Door to be more clear.

jimmyca15 (approved):

tests/Tests.AzureAppConfiguration/RefreshTests.cs
samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -373,6 +390,33 @@ public AzureAppConfigurationOptions Connect(IEnumerable<string> connectionString
             return this;
         }
 
+        /// <summary>
+        /// Connect the provider to Azure Front Door endpoint.
+        /// </summary>
+        /// <param name="endpoint">The endpoint of the Azure Front Door (Afd) instance to connect to.</param>
```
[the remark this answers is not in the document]
samsadsam: done

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Afd/AfdConfigurationClientManager.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+
+            _clientWrapper = new ConfigurationClientWrapper(endpoint, clientFactory.CreateClient(endpoint.AbsoluteUri));
+        }
+
+        public IEnumerable<ConfigurationClient> GetClients()
+        {
+            return new[] { _clientWrapper.Client };
```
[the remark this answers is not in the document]
samsadsam: sounds like we do not adopt the correct pattern for `IEnumerator`. For now, will do this minimal change in this PR, and fix the `IEnumerator` pattern in `AzureAppConfigurationProvider.ExecuteWithFailOverPolicyAsync ` in another PR.

Resolving this

avanigupta (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/KeyValueChange.cs
```diff
@@ -23,5 +26,22 @@ internal struct KeyValueChange
...
         public ConfigurationSetting Previous { get; set; }
+
+        public string GetAfdToken()
+        {
+            string token;
+
+            if (ChangeType == KeyValueChangeType.Deleted)
+            {
+                using SHA256 sha256 = SHA256.Create();
+                token = sha256.ComputeHash(Encoding.UTF8.GetBytes($"ResourceDeleted\n{Previous.ETag}")).ToBase64Url();
```
rossgrambo, line 37: Interesting edge case:

App1 & App2 watching for changes (previous etag: foo)
Setting1 Deleted
App1 refreshes (previous etag: foo) - gets new etag: bar
Setting2 Deleted
App2 refreshes (previous etag: foo)- this hits the cache created by App1 refresh
App1 refreshes (previous etag: bar)- sees the deleted Setting2 - gets new etag: baz
 
This would continue until a setting was changed to bust the etag or the cache expires. Etc:
 
Setting3 Deleted
App2 refreshes (previous etag: foo)- this still hits the cache created by App1 first refresh
App1 refreshes (previous etag: baz)- sees the deleted Setting3 - gets new etag: buzz

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Afd/AfdPolicy.cs
```diff
@@ -43,7 +47,38 @@ public override async System.Threading.Tasks.ValueTask ProcessAsync(HttpMessage
...
+            NameValueCollection query = HttpUtility.ParseQueryString(uriBuilder.Query);
+
+            if (query.Count > 0)
+            {
+                IEnumerable<string> orderedParams = query.AllKeys
+                    .Where(key => key != null)
+                    .OrderBy(key => key.ToLowerInvariant())
+                    .Select(key =>
+                    {
+                        string value = query[key];
```
jimmyca15, line 68: Key is no longer guaranteed to be lowercase here.

zhiyuanliang-ms: @avanigupta @jimmyca15 

Didn't we decide to add the policy of query params normalization to SDK instead of provider?

@mrm9084 has the PR for Java SDK: https://github.com/Azure/azure-sdk-for-java/pull/46665
And I have the PR for JS SDK: https://github.com/Azure/azure-sdk-for-js/pull/35962

zhiyuanliang-ms:
Move to #706

PR #623, opened 2025-02-18
Bugfix for collection monitoring etag logic

amerjusupovic:
Discovered a bug in collection monitoring where an `InvalidOperationException` could happen when casting a null `ETag` from a response. This PR fixes tests that check if kvs/flags don't change to correctly fail when this happens and removes the casting from the `HaveCollectionsChanged` method.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
@@ -95,12 +95,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
             {
                 using Response response = page.GetRawResponse();
 
-                ETag serverEtag = (ETag)response.Headers.ETag;
-
                 // Return true if the lists of etags are different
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
```diff
@@ -95,11 +95,9 @@ public static async Task<bool> HaveCollectionsChanged(this ConfigurationClient c
             {
                 using Response response = page.GetRawResponse();
 
-                ETag serverEtag = (ETag)response.Headers.ETag;
-
                 // Return true if the lists of etags are different
                 if ((!existingMatchConditionsEnumerator.MoveNext() ||
-                    !existingMatchConditionsEnumerator.Current.IfNoneMatch.Equals(serverEtag)) &&
+                    !existingMatchConditionsEnumerator.Current.IfNoneMatch.Equals(response.Headers.ETag)) &&
```
avanigupta, line 100: This could throw null ref exception since `existingMatchConditionsEnumerator.Current.IfNoneMatch` can be null now.

avanigupta (approved):

PR #626, opened 2025-02-21
Add request tracing for push refresh usage

amerjusupovic:


zhiyuanliang-ms (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -33,6 +33,7 @@ internal class RequestTracingConstants
         public const string LoadBalancingEnabledTag = "LB";
         public const string SignalRUsedTag = "SignalR";
         public const string FailoverRequestTag = "Failover";
+        public const string PushRefreshTag = "PushRefresh";
```
zhiyuanliang-ms, line 36: How about just calling it `Refresh`? There is no poll refresh in providers other than java and .net. We are promoting pull refresh. I don't think most people will use poll model.

avanigupta: This tag is only applicable for .net and spring providers for tracking how many customers use push refresh. We already have "Refresh" in correlation context for any refresh operation.

avanigupta (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/RequestTracingOptions.cs
```diff
@@ -63,6 +63,12 @@ internal class RequestTracingOptions
         /// </summary>
         public bool IsFailoverRequest { get; set; } = false;
 
+        /// <summary>
+        /// Flag to indicate whether push refresh is used.
+        /// </summary>
+
```
RichardChen820, line 69: Nit: extra empty line

Issue #632, opened 2025-03-03
BUG: Refresh interval failed to work when there is only feature flag refresh is configured


zhiyuanliang-ms:
A bug reported by our customer [#527](https://github.com/microsoft/FeatureManagement-Dotnet/issues/527)

[Code](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L215)

`_nextCollectionRefreshTime` will not be set if key value refresh is not configured ([kvCollectionRefreshInterval](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs#L436) will be null). This [line](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L417) will always set the current time as the next refresh due which will cause refresh is always due.

The result is that if user only enabled the feature flag refresh and set a refresh interval for feature flag, each time calling `TryRefreshAsync`, the refresh operation will always send requests. But the refresh will happen every time.

amerjusupovic:
@jimmyca15 helped investigate offline, the logic still needs some changes but does not manifest into an actual bug because of how the list of key-values and flags that need refresh are set on [these lines](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/961cdebd2c379a03f715db28b3c5fa07f9772252/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L207-L208). The provider normally requests app config for each watcher as it iterates over the list, but because the list is empty no additional calls are made. This logic still needs an update though, and a PR with the changes will be up soon.

zhiyuanliang-ms:
The bug is not as bad as thought.
The `refreshableFfWatchers`, `refreshableIndividualKvWatchers` and `_options.RegisterAllEnabled` will be empty/false. [code](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L208) So even if it passed the short circuit, there will no actual request.

PR #633, opened 2025-03-05
Ensure kv collection refresh settings are not considered unless the feature is enabled.

jimmyca15:
This PR does the following:

* Ensure that kv collection refresh interval is not used unless collection based refresh of key-values is enabled.
* Fix short circuit code path when no refresh is due.
* Add tests to ensure that minimum refresh interval is respected for key-values and feature flags.

Impact wise, these changes should have little to no behavioral/performance impact. Mostly code clarity.

In response to https://github.com/Azure/AppConfiguration-DotnetProvider/issues/632

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -118,6 +118,13 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             if (options.RegisterAllEnabled)
             {
+                if (options.KvCollectionRefreshInterval <= TimeSpan.Zero)
+                {
+                    throw new ArgumentException(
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -118,6 +118,13 @@ public AzureAppConfigurationProvider(IConfigurationClientManager configClientMan
 
             if (options.RegisterAllEnabled)
             {
+                if (options.KvCollectionRefreshInterval <= TimeSpan.Zero)
```
amerjusupovic, line 121: Just wanted to mention that this shouldn't be possible based on how the refresh interval is set by default if `ConfigureRefresh` and `RegisterAll` were called, but this is fine if we want to just check here for the future

jimmyca15:
@amerjusupovic it looks like the tests i added already existed. I removed them. But I noticed that tests were using Thread.Sleep, I replaced them with Task.Delay.

amerjusupovic (approved):

PR #634, opened 2025-03-07
Add integration testing

amerjusupovic:
First draft of integration testing using copilot


tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+                    foreach (var setting in config.AsEnumerable())
+                    {
+                        if (!string.IsNullOrEmpty(setting.Value) &&
+                            Environment.GetEnvironmentVariable(setting.Key) == null)
+                        {
+                            Environment.SetEnvironmentVariable(setting.Key, setting.Value);
+                        }
+                    }
+                }
+                catch (Exception ex)
```
jimmyca15, line 99: Never catch base exception.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+        private string _skipReason = null;
+
+        /// <summary>
+        /// Loads environment variables from a local settings file if it exists.
+        /// </summary>
+        private void LoadEnvironmentVariablesFromFile()
+        {
+            string localSettingsPath = Path.Combine(AppContext.BaseDirectory, LocalSettingsFile);
+
+            if (File.Exists(localSettingsPath))
```
jimmyca15, line 82: Can the tests run without the file?

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Tests.AzureAppConfiguration
+{
+    /// <summary>
+    /// Integration tests for Azure App Configuration that connect to a real service.
+    /// Creates a temporary App Configuration store for testing and deletes it after the tests are complete.
```
jimmyca15, line 24: How are stale stores cleaned up?

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+
+        // Keys to create for testing
+        private readonly List<ConfigurationSetting> _testSettings = new List<ConfigurationSetting>
+        {
+            new ConfigurationSetting($"{TestKeyPrefix}:Setting1", "InitialValue1"),
+            new ConfigurationSetting($"{TestKeyPrefix}:Setting2", "InitialValue2"),
+            new ConfigurationSetting(SentinelKey, "Initial"),
+            ConfigurationModelFactory.ConfigurationSetting(
+                FeatureFlagKey,
+                @"{""id"":""" + TestKeyPrefix + @"Feature"",""description"":""Test feature"",""enabled"":false}",
```
jimmyca15, line 52: The `@` looks unnecessary here.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+        private const string SentinelKey = TestKeyPrefix + ":Sentinel";
+        private const string FeatureFlagKey = ".appconfig.featureflag/" + TestKeyPrefix + "Feature";
+
+        // Azure Resource Management constants
+        private const string ResourceGroupEnvVar = "AZURE_APPCONFIG_RESOURCE_GROUP";
+        private const string SubscriptionIdEnvVar = "AZURE_SUBSCRIPTION_ID";
+        private const string LocationEnvVar = "AZURE_LOCATION";
+        private const string CreateResourceGroupEnvVar = "AZURE_CREATE_RESOURCE_GROUP";
+        private const string DefaultLocation = "eastus";
+        private const string LocalSettingsFile = "local.settings.json";
```
jimmyca15, line 42: We typically use appsettings.json or appsettings.test.json

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+
+                // Clean up any partially created resources
+                await DisposeAsync();
+            }
+        }
+
+        /// <summary>
+        /// Cleans up the temporary App Configuration store after tests are complete.
+        /// </summary>
+        public async Task DisposeAsync()
```
jimmyca15, line 234: I think we should have Cleanup method that we call on intialization and tear down. That way we don't have stale stores.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+            }
+
+            try
+            {
+                if (_appConfigStore != null)
+                {
+                    await _appConfigStore.DeleteAsync(WaitUntil.Completed);
+                }
+
+                if (_shouldDeleteResourceGroup && _resourceGroup != null)
```
jimmyca15, line 249: Do you think we'll ever want to delete the resource group?

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+                        refresh.Register(sentinelKey, refreshAll: true)
+                              .SetRefreshInterval(TimeSpan.FromSeconds(1));
+                    });
+
+                    refresher = options.GetRefresher();
+                })
+                .Build();
+
+            // Verify initial values
+            Assert.Equal("InitialValue1", config[$"{keyPrefix}:Setting1"]);
```
jimmyca15, line 343: I think you should have one or multiple pre-cache expiration calls to ensure that the keys aren't refreshed until the cache expiration is up. That way cache expiration both ways can be tested here.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+                await _configClient.SetConfigurationSettingAsync(setting);
+            }
+
+            return (keyPrefix, sentinelKey, featureFlagKey);
+        }
+
+        [Fact]
+        public void LoadConfiguration_RetrievesValuesFromAppConfiguration()
+        {
+            Skip.If(_skipTests, _skipReason);
```
jimmyca15, line 302: I don't really understand why we'd want skip. If an exception occurs during initialization, just let it bubble and blow up.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+
+            // Update values in the store
+            await _configClient.SetConfigurationSettingAsync(new ConfigurationSetting($"{keyPrefix}:Setting1", "UpdatedValue1"));
+            await _configClient.SetConfigurationSettingAsync(new ConfigurationSetting(sentinelKey, "Updated"));
+
+            // Wait for cache to expire
+            await Task.Delay(TimeSpan.FromSeconds(2));
+
+            // Act
+            var result = await refresher.TryRefreshAsync();
```
jimmyca15, line 353: I would use RefreshAsync instead of Try.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+            var config = new ConfigurationBuilder()
+                .AddAzureAppConfiguration(options =>
+                {
+                    options.Connect(GetConnectionString());
+                    options.Select($"{keyPrefix}:*");
+
+                    // Only refresh Setting1 when sentinel changes
+                    options.ConfigureRefresh(refresh =>
+                    {
+                        refresh.Register(sentinelKey, $"{keyPrefix}:Setting1", refreshAll: false)
```
jimmyca15, line 378: Is this registering a key with a label `$"{keyPrefix}:Setting1"` ? That looks wrong.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+        private string GetUniqueKeyPrefix(string testName)
+        {
+            // Use a combination of the test prefix and test method name to ensure uniqueness
+            return $"{TestKeyPrefix}_{testName}_{Guid.NewGuid().ToString("N").Substring(0, 8)}";
+        }
+
+        /// <summary>
+        /// Setup test-specific keys and settings
+        /// </summary>
+        private async Task<(string keyPrefix, string sentinelKey, string featureFlagKey)> SetupTestKeys(string testName)
```
jimmyca15, line 272: Avoid returning tuples.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,544 @@
...
+            // Act
+            var result = await refresher.TryRefreshAsync();
+
+            // Assert
+            Assert.True(result);
+            Assert.Equal("True", config[$"FeatureManagement:{keyPrefix}Feature:Enabled"]);
+        }
+
+        [Fact]
+        public async Task RegisterAll_RefreshesAllKeys_WhenSentinelChanged()
```
jimmyca15, line 456: sentinel and refresh all are two completely different paths. I don't see a reason to mix them.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: Yea that's fair, I was thinking about whether the authentication would still work when running in a github workflow and how to handle that. This still isn't what I originally intended though, so I'll look into it.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: Yea I'll take a closer look at the unit tests now, posted draft a bit early

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: If we're creating it for the purpose of the test, I assumed we delete it along with the store. If there's a different to get a resource group or have whoever calls the test pass one in, then no.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: Probably altering the authentication so environment variables from file not needed. This is also requiring the variables to be defined in the file, which is not required.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: I think the name of the resource group/store might need to be deterministic so that we can check for it on startup. This is for the case where `DisposeAsync` doesn't run and stale rgs/stores are left over.

amerjusupovic:
Still needs some cleanup, adding changes based on comments and will continue adding tests. Will use this PR to test integration test authentication in github workflow

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: When removed it's a compile error, I think it might just be easier to leave as is?

jimmyca15: Why is it a compilation error

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: Discussed offline, will no longer be creating/cleaning up stores

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: The choice is basically to escape each double quote with backslash or with 2 quotes like when we use the `@` symbol, the compile error was just because I didn't escape it correctly the first time. I feel like keeping it this way (using `@`) is more legible since it's consistent with other tests.

jimmyca15: The purpose of `@` is to avoid having to escape characters that would otherwise require escaping in a regular string. However, if you are saying we are using this elsewhere then I'm fine to keep it.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+
+                AppConfigurationStoreResource appConfigStore = null;
+                KeyVaultResource keyVault = null;
+
+                try
+                {
+                    // Get App Configuration store directly using the resource group and store name
+                    appConfigStore = await resourceGroup.GetAppConfigurationStores().GetAsync(AppConfigStoreName);
+                }
+                catch (RequestFailedException ex)
```
jimmyca15, line 198: The error message expects a not found failure, but you're catching all request failed exceptions.

tests/Tests.AzureAppConfiguration/Integration/GetAzureSubscription.ps1
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env pwsh
+
+# GetAzureSubscription.ps1
+# This script gets the current Azure subscription ID and saves it to a JSON file
+# Must be logged into AppConfig - Dev subscription for the integration tests to work.
+
```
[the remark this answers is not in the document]
amerjusupovic: Got it, I'll take a look at the existing pattern and update the PR

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+            // Assert
+            Assert.Equal("InitialValue1", config[$"{testContext.KeyPrefix}:Setting1"]);
+            Assert.Equal("InitialValue2", config[$"{testContext.KeyPrefix}:Setting2"]);
+        }
+
+        [Fact]
+        public async Task RefreshAsync_UpdatesConfiguration_WhenSentinelKeyChanged()
+        {
+            // Arrange - Setup test-specific keys
+            var testContext = await SetupTestKeys("UpdatesConfig");
```
jimmyca15, line 514: avoid var if the type does not appear in the assignment.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+            {
+                throw new InvalidOperationException($"Subscription JSON file not found at {jsonPath}. Run the GetAzureSubscription.ps1 script first.");
+            }
+
+            string jsonContent = File.ReadAllText(jsonPath);
+
+            using JsonDocument doc = JsonDocument.Parse(jsonContent);
+            JsonElement root = doc.RootElement;
+
+            bool success = root.GetProperty("Success").GetBoolean();
```
jimmyca15, line 115: I don't understand why this is valuable?

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+                throw new InvalidOperationException(root.GetProperty("ErrorMessage").GetString());
+            }
+
+            return root.GetProperty("SubscriptionId").GetString();
+        }
+
+        /// <summary>
+        /// Returns the connection string for connecting to the app configuration store.
+        /// </summary>
+        private string GetConnectionString()
```
jimmyca15, line 128: Unnecessary. It's a private method, anyone who can access this can just access `_connectionString`

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+        {
+            return _connectionString;
+        }
+
+        /// <summary>
+        /// Creates a unique prefix for test keys to ensure test isolation
```
jimmyca15, line 134: These summaries on private methods are a bit verbose. Seem self explanatory.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+        private readonly HashSet<string> _createdConfigKeys = new HashSet<string>();
+        private readonly HashSet<string> _createdSecretNames = new HashSet<string>();
+        private readonly HashSet<string> _createdSnapshotNames = new HashSet<string>();
+
+        /// <summary>
+        /// Gets a DefaultAzureCredential for authentication.
+        /// </summary>
+        private DefaultAzureCredential GetCredential()
+        {
+            return new DefaultAzureCredential(new DefaultAzureCredentialOptions
```
jimmyca15, line 83: Can you have this as a private field?

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// Initialize clients to connect to the existing App Configuration store and Key Vault.
+        /// </summary>
+        public async Task InitializeAsync()
+        {
+            try
```
jimmyca15, line 179: try should only surround code that can throw. And it should only catch the exceptions that the code its surrounding can throw.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+                // Get connection string from the store
+                var accessKeys = appConfigStore.GetKeysAsync();
+                var primaryKey = await accessKeys.FirstOrDefaultAsync();
+
+                if (primaryKey == null)
+                {
+                    throw new InvalidOperationException("Failed to retrieve access keys from App Configuration store.");
+                }
+
+                _connectionString = primaryKey.ConnectionString;
```
jimmyca15, line 214:
```suggestion
                var primaryKey = await accessKeys.FirstOrDefaultAsync();

                if (primaryKey == null)
                {
                    throw new InvalidOperationException("Failed to retrieve access keys from App Configuration store.");
                }

                _connectionString = (await accessKeys.FirstAsync()).ConnectionString;
```

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+                CreateSnapshotOperation operation = await _configClient.CreateSnapshotAsync(WaitUntil.Completed, snapshotName, snapshot);
+
+                // Track created snapshot for cleanup
+                _createdSnapshotNames.Add(snapshotName);
+
+                return operation.Value.Name;
+            }
+            catch (RequestFailedException ex)
+            {
+                Console.WriteLine($"Error creating snapshot: {ex.Message}");
```
jimmyca15, line 169: Avoid catch/log/throw pattern. The exception bubbling up and crashing the tests will create enough logging.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,1923 @@
...
+        private string GetUniqueKeyPrefix(string testName)
+        {
+            // Use a combination of the test prefix and test method name to ensure uniqueness
+            return $"{TestKeyPrefix}-{testName}-{Guid.NewGuid().ToString("N").Substring(0, 8)}";
+        }
+
+        /// <summary>
+        /// Creates a snapshot with the given name containing the test context's settings
+        /// </summary>
+        private async Task<string> CreateSnapshot(string snapshotName, TestContext testContext)
```
jimmyca15, line 145: Most of `TestContext` isn't used here. I'd like to see the caller pass what's needed only. In this case I'd suggest passing `IEnumerable<ConfigurationSettingsFilter>`

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: accidentally left this in after change, will update

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: yea it can be removed

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: Not sure I understand, `FirstAsync` instead of `FirstOrDefaultAsync`?

jimmyca15: Yes, FirstAsync has the invalid operation exception built-in.

amerjusupovic: Oh I see thanks

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
amerjusupovic: Removed most summaries

.github/code-gen-instructions.md
```diff
@@ -0,0 +1,93 @@
+# AppConfiguration-DotnetProvider Coding Guidelines
```
jimmyca15, line 1: Can we put this in a separate PR?

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,2041 @@
...
+                    $"Key Vault '{KeyVaultName}' not found in subscription {subscriptionId}. " +
+                    "This resource is required for integration tests. " +
+                    "Please create the Key Vault with the appropriate permissions before running tests.");
+            }
+
+            _keyVaultEndpoint = keyVault.Data.Properties.VaultUri;
+
+            _secretClient = new SecretClient(_keyVaultEndpoint, credential);
+
+            Console.WriteLine($"Successfully connected to App Configuration store '{AppConfigStoreName}' and Key Vault '{KeyVaultName}'");
```
jimmyca15, line 166: XUnit exposes some type like ITestOutputHelper that's supposed to replace Console when writing messages during tests.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -0,0 +1,2041 @@
...
+        {
+            None = 0,
+            KeyValues = 1,
+            FeatureFlags = 2,
+            KeyVaultReferences = 4,
+            TaggedSettings = 8,
+            All = KeyValues | FeatureFlags | KeyVaultReferences
+        }
+
+        private async Task<TestContext> SetupTestKeys(string testName, TestDataTypes dataTypesToCreate = TestDataTypes.All)
```
jimmyca15, line 255: Creating a type adds a bit more complexity then we usually aim for. I'd suggest updating SetupTestKeys a bit.

Suggestion

```
TestContext context = await SetupTestKeys("footest", TestDataTypes.KeyValues | TestDataTypes.FeatureFlags)
```

becomes

```
TestContext context = await CreateTestContext("footest");

await SetupFeatureFlags(context);

await SetupTestKeys(context);
```

amerjusupovic: Makes sense, removed the flag and created a method for each flag value that existed instead.

jimmyca15 (approved):

PR #637, opened 2025-03-18
Add support for filtering by tags

amerjusupovic:
Extends the existing APIs `AzureAppConfigurationOptions.Select` and `FeatureFlagOptions.Select` to have a new parameter that supports passing tags that will be used to filter key-values and feature flags. Key-values and feature flags will only be returned if they contain all of the tags and corresponding values passed to the `tagFilters` parameter.

### Updated APIs
```csharp
public AzureAppConfigurationOptions Select(
    string keyFilter, 
    string labelFilter = LabelFilter.Null, 
    IEnumerable<string> tagFilters = null)
```

```csharp
public FeatureFlagOptions Select(
    string featureFlagFilter, 
    string labelFilter = LabelFilter.Null, 
    IEnumerable<string> tagFilters = null)
```

### Example Usage

```csharp
config.AddAzureAppConfiguration(options =>
{
    options.Select(KeyFilter.Any, LabelFilter.Null, new List<string> { "group=g1", "tag=value" });
})
```

```csharp
config.AddAzureAppConfiguration(options =>
{
    options.UseFeatureFlags(ffOptions =>
    {
        ffOptions.Select(KeyFilter.Any, LabelFilter.Null, new List<string> { "group=g1", "tag=value" });
    });
})
```

#### Filtering for tags with empty and null value

You can use the `TagValue` class to filter for tags that have `null` value. You can also filter for empty string by constructing a tag filter string like `"EmptyTag="`.

Key-value as a json:

```json
{
	"key": "TestKey",
	"label": null,
	"value": "TestValue",
	"content_type": "",
	"tags": {
		"EmptyTag": "",
		"NullTag": null
	}
}
```

To select key-value:

```csharp
options.Select(KeyFilter.Any, LabelFilter.Null, new List<string> { "EmptyTag=", $"NullTag={TagValue.Null}" });
```


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
```diff
@@ -40,7 +47,8 @@ public override bool Equals(object obj)
             {
                 return KeyFilter == selector.KeyFilter
                     && LabelFilter == selector.LabelFilter
-                    && SnapshotName == selector.SnapshotName;
+                    && SnapshotName == selector.SnapshotName
+                    && TagsFilter == selector.TagsFilter;
```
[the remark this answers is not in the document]
amerjusupovic: Will add copilot suggestions

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -202,7 +202,13 @@ public AzureAppConfigurationOptions SetClientFactory(IAzureClientFactory<Configu
         /// The label filter to apply when querying Azure App Configuration for key-values. By default the null label will be used. Built-in label filter options: <see cref="LabelFilter"/>
         /// The characters asterisk (*) and comma (,) are not supported. Backslash (\) character is reserved and must be escaped using another backslash (\).
         /// </param>
-        public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter = LabelFilter.Null)
+        /// <param name="tagsFilter">
+        /// The tag filter to apply when querying Azure App Configuration for key-values. By default no tags will be used.
+        /// Each tag provided must follow the format "tag=value". A key-value will only be returned if its tags contain all tags provided
+        /// in the filter, or if the filter is empty.
+        /// The characters asterisk (*), comma (,) and backslash (\) are reserved and must be escaped using a backslash (\).
+        /// </param>
```
avanigupta, line 210:
```suggestion
        /// <param name="tagsFilter">
        /// In addition to key and label filters, key-values from Azure App Configuration can be filtered based on their tag names and values.
        /// Each tag filter must follow the format "tagName=tagValue". Only those key-values will be loaded whose tags match all the tags provided here.
        /// The characters asterisk (*), comma (,) and backslash (\) are reserved and must be escaped using a backslash (\).
        /// Up to 5 tag filters can be provided. If no tag filters are provided, key-values will not be filtered based on tags.
        /// </param>
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -220,6 +226,17 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
                 labelFilter = LabelFilter.Null;
             }
 
+            if (tagsFilter != null)
+            {
+                foreach (var tag in tagsFilter)
+                {
+                    if (string.IsNullOrEmpty(tag) || !tag.Contains('=') || tag.IndexOf('=') == 0)
```
avanigupta, line 233: Does SDK throw any error if the tag filter is malformed?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -220,6 +226,17 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
...
+                    {
+                        throw new ArgumentException($"Tag '{tag}' does not follow the format \"tag=value\".", nameof(tagsFilter));
```
avanigupta, line 235:
```suggestion
                        throw new ArgumentException($"Tag '{tag}' does not follow the format \"tagName=tagValue\".", nameof(tagsFilter));
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -814,6 +814,14 @@ private async Task<Dictionary<string, ConfigurationSetting>> LoadSelected(
                         LabelFilter = loadOption.LabelFilter
                     };
```
avanigupta, line 816: Do we need to setup tag filters here: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/87f0f85ca2e4011f82d93a864e35c6c804cd6c39/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L343-L348

tests/Tests.AzureAppConfiguration/TagsFilterTests.cs
```diff
@@ -0,0 +1,496 @@
...
+            var mockClient = new Mock<ConfigurationClient>(MockBehavior.Strict);
+
+            // Verify that an ArgumentException is thrown when using an invalid tag format
+            var exception = Assert.Throws<ArgumentException>(() =>
+            {
+                new ConfigurationBuilder()
+                    .AddAzureAppConfiguration(options =>
+                    {
+                        options.ClientManager = TestHelpers.CreateMockedConfigurationClientManager(mockClient.Object);
+                        options.Select(KeyFilter.Any, "label", new List<string> { "InvalidTagFormat" });
```
avanigupta, line 212: Can you add some more invalid tag filter cases

tests/Tests.AzureAppConfiguration/TagsFilterTests.cs
```diff
@@ -0,0 +1,496 @@
...
+            if (tags != null)
+            {
+                foreach (var tag in tags)
+                {
+                    setting.Tags.Add(tag.Key, tag.Value);
+                }
+            }
+
+            return setting;
+        }
```
avanigupta, line 85: Can you add a test for tag filter with null/empty value and a test for more than 5 tag filters.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -202,7 +202,13 @@ public AzureAppConfigurationOptions SetClientFactory(IAzureClientFactory<Configu
         /// The label filter to apply when querying Azure App Configuration for key-values. By default the null label will be used. Built-in label filter options: <see cref="LabelFilter"/>
         /// The characters asterisk (*) and comma (,) are not supported. Backslash (\) character is reserved and must be escaped using another backslash (\).
         /// </param>
-        public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter = LabelFilter.Null)
+        /// <param name="tagsFilter">
+        /// The tag filter to apply when querying Azure App Configuration for key-values. By default no tags will be used.
+        /// Each tag provided must follow the format "tag=value". A key-value will only be returned if its tags contain all tags provided
+        /// in the filter, or if the filter is empty.
```
avanigupta, line 208: How can users specify tags with null value? Maybe we need to add a default tag filter value like we have LabelFilter.Null today.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
```diff
@@ -52,9 +62,33 @@ public override bool Equals(object obj)
         /// <returns>A hash code for the current object.</returns>
         public override int GetHashCode()
         {
+            int tagsFilterHash = 3;
+
+            if (TagsFilter != null && TagsFilter.Any())
+            {
+                var sortedTags = new SortedSet<string>(TagsFilter);
+
+                if (sortedTags.Any())
```
avanigupta, line 71: Why check for sortedTags.Any() here if its already been checked with TagsFilter.Any()

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
```diff
@@ -52,9 +62,33 @@ public override bool Equals(object obj)
...
+                {
+                    // Concatenate tags into a single string with a delimiter
+                    string tagsString = string.Join("|", sortedTags);
+
+                    // Use SHA256 to generate a hash for the tags
+                    using (var sha256 = System.Security.Cryptography.SHA256.Create())
```
avanigupta, line 77: Crypto hash is an overkill here. Can we get the hash of each tag filter (string) instead of using SHA256?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
```diff
@@ -52,9 +62,33 @@ public override bool Equals(object obj)
...
+                    }
+                }
+            }
+
             return (KeyFilter?.GetHashCode() ?? 0) ^
                    (LabelFilter?.GetHashCode() ?? 1) ^
-                   (SnapshotName?.GetHashCode() ?? 2);
+                   (SnapshotName?.GetHashCode() ?? 2) ^
+                   tagsFilterHash ^
+                   (IsFeatureFlagSelector.GetHashCode());
```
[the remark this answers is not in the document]
avanigupta: [HashCode.Combine](https://learn.microsoft.com/en-us/dotnet/api/system.hashcode?view=netcore-2.1) seems like a better approach.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: "tag=" is valid, that's the only way right now. They still have to specify a tag name though so would they have to use the default tag filter value as part of a concatenated/formatted string?

`"tag=" + TagFilter.Null` ?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
amerjusupovic: Oh yea good point my fault

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
amerjusupovic: Oh yea that makes more sense here, thanks

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
avanigupta: "tag=" means the value is empty string. "tag=\0" means value is null.
I was thinking of something like:
`$"tag={TagValue.Null}`

amerjusupovic: I see, yea I can add that and mention it in the comment here

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -202,7 +202,13 @@ public AzureAppConfigurationOptions SetClientFactory(IAzureClientFactory<Configu
...
+        /// The characters asterisk (*), comma (,) and backslash (\) are reserved and must be escaped using a backslash (\).
+        /// </param>
+        public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter = LabelFilter.Null, IEnumerable<string> tagsFilter = null)
```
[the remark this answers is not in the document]
amerjusupovic: I don't think that came up, I just mentioned that the SDK used `tagsFilter`. That's a reasonable change though, would you want to discuss it again? I could just ask everyone involved

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
amerjusupovic: yes, will update

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: Yes, a RequestFailedException that will say invalid parameter for tags. This also changes the optional behavior, is it better to just not check the format here?

avanigupta: So it sounds like this error is coming from the backend. SDK is just relaying the error as a RequestFailedException. It's a bad input error so we should be okay to throw it even if optional is true. I think its ok to have basic validation here to fail fast, just like we validate key and label filters above. All possible scenarios will not be covered here, but checking for '=' sign and non-empty tag name is a good basic check.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
avanigupta: I feel SDK name is a miss - it should've been `tagsFilters` or `tagFilters`. But we can still use a better name here since we dont necessarily need to use the same name as SDK. Thoughts @jimmyca15 ?

tests/Tests.AzureAppConfiguration/TagsFilterTests.cs
amerjusupovic: I was thinking about how to test more than 5 tag filters. Does it make sense to validate this in the provider like the other basic validation for tags? Otherwise, I can still add a test that will just mock the `RequestFailedException` returned

avanigupta: We can add basic validation to throw if user provides more than 5 tag filters

amerjusupovic: Talked to jimmy offline, seems like it would be better to not add this validation since we don't validate this kind of thing in other APIs and the error message that comes from the exception is already sufficient. We don't need to restrict the tag filters count from the provider side

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
amerjusupovic: Talked to jimmy offline, seems like we agree that `tagsFilter` isn't great. The `ConfigurationSnapshot` class from the SDK has a property called `Filters` which is a collection of `ConfigurationSettingsFilter` objects. Based on this, I think this can at least be `tagFilters` since each string is filtering for one tag, unlike how the `ConfigurationSettingsFilter` class can sort for multiple with that one object. I think `tagsFilters` is also reasonable though.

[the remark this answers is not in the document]
amerjusupovic: We can go with `tagsFilters` to better match existing docs

tests/Tests.AzureAppConfiguration/TagsFilterTests.cs
avanigupta: I'd forgotten we allow comma separated key filters too and dont do this validation for key filters. So it makes sense to ski the validation for tags also

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
```diff
@@ -24,6 +28,11 @@ public class KeyValueSelector
         /// </summary>
         public string SnapshotName { get; set; }
 
+        /// <summary>
+        /// A filter that determines what tags to require when selecting key-values for the the configuration provider.
+        /// </summary>
+        public IEnumerable<string> TagsFilters { get; set; }
```
zhiyuanliang-ms, line 34:
```suggestion
        public IEnumerable<string> TagsFilter { get; set; }
```

JS sdk uses the name `tagsFilter`. [ref](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/appconfiguration/app-configuration/src/models.ts#L349)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
```diff
@@ -15,10 +15,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Data.AppConfiguration" Version="1.4.1" />
+    <PackageReference Include="Azure.Data.AppConfiguration" Version="1.5.0" />
```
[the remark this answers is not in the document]
amerjusupovic: Oh didn't see the new release, will update

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueSelector.cs
zhiyuanliang-ms: Interesting. I planned to use this name for JS provider.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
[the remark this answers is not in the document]
zhiyuanliang-ms: I think `tagFilters` and `tagsFilter` both make sense.

I prefer to `tagsFilter`. Multiple tags together form a `tagsFilter`. It is a filter which is an array of tags. I want to use the name `tagsFilter` for JS provider as it is consistent with [SDK](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/appconfiguration/app-configuration/src/models.ts#L349)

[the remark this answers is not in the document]
zhenlan: Tag is the adjective in this case, so it should be TagFilter(s), not TagsFilter(s).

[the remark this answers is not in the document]
jimmyca15: Had the chance to think about this a bit more. Should be either TagFilters or TagsFilter. Both are correct in different ways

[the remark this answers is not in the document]
avanigupta: I think the contention comes from the name of our REST query parameter, which is "tags" instead of "tag". This made sense on the API level because we wanted that name to represent multiple tag filters (its an array). But the same does not hold true for our SDKs where we use the suffix of "Filter" in the property name to denote the same thing as the REST API. 

I agree with Jimmy and prefer TagFilters over TagsFilter because it follows natural language naming conventions better.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -220,6 +227,17 @@ public AzureAppConfigurationOptions Select(string keyFilter, string labelFilter
                 labelFilter = LabelFilter.Null;
             }
 
+            if (tagFilters != null)
+            {
+                foreach (var tag in tagFilters)
```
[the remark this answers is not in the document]
amerjusupovic: updated in `FeatureFlagOptions` too

tests/Tests.AzureAppConfiguration/TagFiltersTests.cs
```diff
@@ -0,0 +1,630 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Tests.AzureAppConfiguration
+{
+    public class TagFiltersTests
+    {
+        private List<ConfigurationSetting> _kvCollection;
+        private List<ConfigurationSetting> _ffCollection;
```
avanigupta, line 21: Unused collection

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -74,7 +74,14 @@ public FeatureFlagOptions SetRefreshInterval(TimeSpan refreshInterval)
         /// The label filter to apply when querying Azure App Configuration for feature flags. By default the null label will be used. Built-in label filter options: <see cref="LabelFilter"/>
         /// The characters asterisk (*) and comma (,) are not supported. Backslash (\) character is reserved and must be escaped using another backslash (\).
         /// </param>
-        public FeatureFlagOptions Select(string featureFlagFilter, string labelFilter = LabelFilter.Null)
+        /// <param name="tagFilters">
+        /// In addition to key and label filters, key-values from Azure App Configuration can be filtered based on their tag names and values.
```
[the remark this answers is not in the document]
amerjusupovic: Thanks, will update

avanigupta (approved):

avanigupta:
Can you also update the PR description with the right parameter name?

amerjusupovic:
> Can you also update the PR description with the right parameter name?

Thanks, updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TagValue.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    /// <summary>
+    /// Defines well known tag values that are used within Azure App Configuration.
+    /// </summary>
+    public class TagValue
```
jimmyca15, line 9: Can you give an example of this usage in the description

jimmyca15: Perhaps demonstrating the content of a key-value that has a null tag to demonstrate the concept.

amerjusupovic:
Still need to update to remove preview commits

amerjusupovic:
Updated after changing target branch to main

jimmyca15 (approved):

PR #643, opened 2025-04-07
Use TryAddSingleton instead of AddSingleton for extension

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs
```diff
@@ -13,11 +13,38 @@ internal class AzureAppConfigurationRefresherProvider : IConfigurationRefresherP
     {
         private static readonly PropertyInfo _propertyInfo = typeof(ChainedConfigurationProvider).GetProperty("Configuration", BindingFlags.Public | BindingFlags.Instance);
 
-        public IEnumerable<IConfigurationRefresher> Refreshers { get; }
+        private readonly IConfiguration _configuration;
+        private readonly ILoggerFactory _loggerFactory;
+        private IEnumerable<IConfigurationRefresher> _refreshers;
+        private bool _rediscoveredRefreshers = false;
```
zhiyuanliang-ms, line 19: This is property is always false. Why do we need it?

jimmyca15 (approved):

PR #644, opened 2025-04-09
Support health check

zhiyuanliang-ms:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -198,6 +204,7 @@ public override void Load()
 
             // Mark all settings have loaded at startup.
             _isInitialLoadComplete = true;
+            LastSuccessfulAttempt = DateTime.UtcNow;
```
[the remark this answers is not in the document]
zhiyuanliang-ms: The keyvaule secret resolving doesn't happen in the `ExecuteWithFailOverPolicyAsync` path.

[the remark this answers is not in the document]
zhiyuanliang-ms: Also, if I am right, when there is no available backoff clients, the `RefreshAsync` method even doesn't have chance to enter `ExecuteWithFailOverPolicyAsync`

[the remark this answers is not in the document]
zhiyuanliang-ms: Updated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationHealthCheck.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    /// <summary>
+    /// An interface for Azure App Configuration health check.
+    /// </summary>
+    public interface IConfigurationHealthCheck : IHealthCheck
```
zhenlan, line 11: This is different than `IConfigurationRefresher`. Suggest renaming to `IAzureAppConfigurationHealthCheck`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationHealthCheck.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
```
zhenlan, line 6: Zhiyuan, can you help check what namespaces people use for their HealthCheck implementation? Do they put it in their own namespace or do they put it in the `Microsoft.Extensions.Diagnostics.HealthChecks` namespace?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthCheck.cs
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using System;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal class AzureAppConfigurationHealthCheck : IConfigurationHealthCheck
```
zhenlan, line 11: As we discussed, just like the refresher, this should be added to DI as part of `AddAzureAppConfiguration(this IServiceCollection services)`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthCheck.cs
```diff
@@ -0,0 +1,41 @@
...
+                return HealthCheckResult.Unhealthy("Configuration provider is not set.");
+            }
+
+            if (!_provider.LastSuccessfulAttempt.HasValue)
+            {
+                return HealthCheckResult.Unhealthy("The initial load is not completed.");
+            }
+
+            if (_provider.LastFailedAttempt.HasValue &&
+                _provider.LastSuccessfulAttempt.Value < _provider.LastFailedAttempt.Value)
```
zhenlan, line 33: If the refresh occurs frequently (eg. every a few seconds), we shouldn't report unhealthy just because last refresh failed for whatever reason. It causes unnecessary panic/alarm.

jimmyca15: I wouldn't say reporting unhealthy should be tied to the frequency of refresh. After all, follow up attempts on failure are not based on the configured frequency, they are based on our backup strategy. 

Typically, the orchestrator that is monitoring the health of the app is the one to take an action on whether the app is healthy or not and is the one to protect against unnecessary action.

If any change is made, I would expect an option, either a failure duration or failure count to represent how long the app configuration provider has failed to reach the service which leads to unhealthy. Otherwise, anything we pick would be arbitrary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthCheck.cs
jimmyca15: I had suggested the same, additionally looking to see the PR description updated to describe the usage with DI.

zhiyuanliang-ms: IHealthCheck should have 1:1 map to AzureAppConfigurationProvider. In `AddAzureAppConfiguration(this IServiceCollection services)`, we add [`IConfigurationRefresherProvider`](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs) to the DI, which will recursivly find all AzureAppConfigurationProviders in the `IConfiguration` and put them in `Refreshers` property.

The normal DI way to add health check is:
```C#
builder.Services
    .AddHealthChecks() // returns an IHealthChecksBuilder
    .Add(new HealthCheckRegistration(
        name,
        sp => {
            IHealthCheck healthCheck;
            // do something
            return healthCheck;
        },
        failureStatus,
        tags,
        timeout
    ))
```

[Add](https://github.com/dotnet/aspnetcore/blob/main/src/HealthChecks/HealthChecks/src/DependencyInjection/HealthChecksBuilder.cs#L19) method and [HealthCheckRegistration](https://github.com/dotnet/aspnetcore/blob/main/src/HealthChecks/Abstractions/src/HealthCheckRegistration.cs#L103) class.

Usually, an extension method on IHealthChecksBuilder will be provided, so the above code will be simplified to:

```C#
builder.Services
    .AddHealthChecks()
    .AddAzureAppConfiguration();
```
[Here](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.Azure.KeyVault.Secrets/DependencyInjection/AzureKeyVaultHealthChecksBuilderExtensions.cs#L34) is an example of key vault.


The [`IHealthCheckBuilder`](https://github.com/dotnet/aspnetcore/blob/main/src/HealthChecks/HealthChecks/src/DependencyInjection/IHealthChecksBuilder.cs#L11) only has access to `IServiceCollection`. It is not guaranteed to get the IConfiguration there.

zhiyuanliang-ms: The usage I expected is:

```C#
public static IHealthChecksBuilder AddAzureKeyVaultSecrets(
    this IHealthChecksBuilder builder,
    string name,
    Func<IServiceProvider, IHealthCheck> factory = default
    HealthStatus failureStatus = default,
    IEnumerable<string> tags = default,
    TimeSpan timeout = default)
{
    return builder.Add(new HealthCheckRegistration(
       name,
       sp => factory?.Invoke(sp) ?? sp.GetRequiredService<IAzureAppConfigurationHealthCheck>()),
       failureStatus,
       tags,
       timeout));
}



var builder = Host.CreateApplicationBuilder(args);
builder.Configuration.AddAzureAppConfiguration(options =>
{
    options.Connect("connection-string");
    builder.Services.AddSingleton(options.GetHealthCheck());
});

builder.Services
    .AddHealthChecks()
    .AddAzureAppConfiguration()
```

----------------------------------------------------------
Or 
``` C#
public static IHealthChecksBuilder AddAzureKeyVaultSecrets(
    this IHealthChecksBuilder builder,
    string name,
    IHealthCheck healthCheck,
    HealthStatus failureStatus = default,
    IEnumerable<string> tags = default,
    TimeSpan timeout = default)
{
    return builder.Add(new HealthCheckRegistration(
       name,
       healthCheck,
       failureStatus,
       tags,
       timeout));
}

var builder = Host.CreateApplicationBuilder(args);
IHealthCheck healthCheck = null;
builder.Configuration.AddAzureAppConfiguration(options =>
{
    options.Connect("connection-string");
    healthCheck = options.GetHealthCheck();
});

builder.Services
    .AddHealthChecks()
    .AddAzureAppConfiguration(healthCheck);
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IConfigurationHealthCheck.cs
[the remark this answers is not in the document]
zhiyuanliang-ms: Checked existing examples. IHealthCheck is enough.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthCheck.cs
zhiyuanliang-ms: Discussed with @jimmyca15, landed on having an "aggregated" `AzureAppConfigurationHealthCheck` which will do health check for all `AzureConfigurationProviders` in the IConfiguration. And added an extension method on IHealthChecksBuilder to enable the usage as below:

``` C#
builder.Services
    .AddHealthChecks()
    .AddAzureAppConfiguration()
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthCheck.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Threading;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal class AzureAppConfigurationHealthCheck : IHealthCheck
+    {
+        private static readonly PropertyInfo _propertyInfo = typeof(ChainedConfigurationProvider).GetProperty("Configuration", BindingFlags.Public | BindingFlags.Instance);
```
jimmyca15, line 16: This property was made public. Do we still support a version where it isn't public?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthCheck.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+        {
+            if (!_providers.Any())
+            {
+                return HealthCheckResult.Unhealthy(HealthCheckConstants.NoProviderFoundMessage);
+            }
+
+            foreach (var provider in _providers)
```
jimmyca15, line 37: avoid var unless the type is in the assignment.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthCheck.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal class AzureAppConfigurationHealthCheck : IHealthCheck
+    {
+        private static readonly PropertyInfo _propertyInfo = typeof(ChainedConfigurationProvider).GetProperty("Configuration", BindingFlags.Public | BindingFlags.Instance);
+        private readonly List<AzureAppConfigurationProvider> _providers = new List<AzureAppConfigurationProvider>();
```
jimmyca15, line 17: IEnumerable

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1152,6 +1159,7 @@ private async Task<T> ExecuteWithFailOverPolicyAsync<T>(
                     if (!IsFailOverable(ae) || !clientEnumerator.MoveNext())
                     {
                         backoffAllClients = true;
```
jimmyca15, line 1161: It looks like this should be set in the finally block inside the `if (!success...)` check.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT license.
```
jimmyca15, line 2: I don't see any changes. I'm wondering why this is in the diff.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -58,6 +58,9 @@ private class ConfigurationClientBackoffStatus
             public DateTimeOffset BackoffEndTime { get; set; }
         }
 
+        public DateTimeOffset? LastSuccessfulAttempt { get; private set; } = null;
```
jimmyca15, line 61: I'd rather see the provider still implement IHealthCheck, and the new aggregate class you add simply aggregate the health checks of providers, rather than externalizing the determination of what's healthy.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthCheck.cs
zhiyuanliang-ms: I copied this from [`AzureAppConfigurationRefresherProvider`](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefresherProvider.cs#L14)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
zhiyuanliang-ms: The end of the line, I guess... The linting tool makes it change from CRLF to LF.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhiyuanliang-ms: Updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthCheck.cs
jimmyca15: I know that. The question still stands. If That property (`ChainedConfigurationProvider.Configuration`) is public in all the languages we support, then we no longer need reflection.

zhiyuanliang-ms: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.chainedconfigurationprovider?view=net-9.0-pp&viewFallbackFrom=net-6.0-pp

It seems like `ChainedConfigurationProvider.Configuration` is public for .NET 8+

jimmyca15: I left another comment, looks like it didn't go through.

Since we're targeting NetStandard we really don't have a chance here to drop this reflection.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthChecksBuilderExtensions.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.DependencyInjection
```
jimmyca15, line 10: Just want to confirm before check-in, this is the same namespace other health check extensions are in right?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthChecksBuilderExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    /// <summary>
+    /// Extension methods to configure <see cref="AzureAppConfigurationHealthCheck"/>.
+    /// </summary>
+    public static class AzureAppConfigurationHealthChecksBuilderExtensions
+    {
+        /// <summary>
+        /// Add a health check for Azure App Configuration by registering <see cref="AzureAppConfigurationHealthCheck"/> for given <paramref name="builder"/>.
```
jimmyca15, line 18: Don't put an internal type reference in the summary of a public type.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationHealthChecksBuilderExtensions.cs
zhiyuanliang-ms: Yes. [ref](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.Azure.KeyVault.Secrets/DependencyInjection/AzureKeyVaultHealthChecksBuilderExtensions.cs#L5)

PR #645, opened 2025-04-11
Add activity source to allow OpenTelemetry to collect tracing

zhiyuanliang-ms:
## Why this PR?

[OpenTelemetry](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/observability-with-otel) tracing in .NET is built on [`Activity`](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing-instrumentation-walkthroughs).

Only `Activity` from `ActivitySource` which is registered to `TracerProvider` by `AddSource` method will be traced by OpenTelemetry. （Implementation details: [TracerProviderSdk](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L263) OTEL will create activity listener to trace activities.)

This PR adds an `ActivitySource` with name called `"Microsoft.Extensions.Configuration.AzureAppConfiguration"` to allow OpenTelemetry to collect tracing for the `Load` and `Refresh` activities.

## Bug fixes

Activity created by the .NET SDK([`ConfigurationClient`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs#L614)) cannot be traced because of the request tracing in .NET provider which is built on "CustomDiagnosticsHeaders" activity.

Below are the implementation details of the request tracing in .NET SDK and .NET provider:
Request tracing hook: [SDK](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/appconfiguration/Azure.Data.AppConfiguration/src/CustomHeadersPolicy.cs#L14), [provider](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs#L119)

The .NET provider will start a new Activity with no source/listener. This activity will have [`ActivityTraceFlags`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitytraceflags?view=net-6.0) set to `ActivityTraceFlags.None`. According to [doc](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing-collection-walkthroughs), this kind of activity will not be traced. And the activity created by the `ConfigurationClient` from SDK will be the child activity of the activity we created and it will inherit the `ActivityTraceFlags` which will cause the child activity not to be traced.

The below code demostrates the issue in a simplified way:
```C#
using System.Diagnostics;
using OpenTelemetry.Trace;

var builder = Host.CreateApplicationBuilder(args);

List<Activity> exportedActivities = new();
var activitySource = new ActivitySource("TestSource"); // mocked activity source of ConfigurationClient

builder.Services.AddOpenTelemetry()
    .WithTracing(builder =>
    {
        builder.AddInMemoryExporter(exportedActivities)
            .AddSource(["TestSource"]); // OTEL sdk will create an ActivityListener to listen to the registered ActivitySource
    });


using (var host = builder.Build())
{
    host.Start();
    var activityWithoutSource = new Activity("mocked activity started by .NET provider for request tracing");
    activityWithoutSource.Start();
    Console.WriteLine(Activity.Current?.DisplayName);
    Console.WriteLine(Activity.Current?.ActivityTraceFlags);
    //Activity.Current.ActivityTraceFlags = ActivityTraceFlags.Recorded; // if we manually set the flag to Recorded, then the activity below will then be catched by OTEL

    var testActivity = activitySource.StartActivity("mocked activity created by ConfigurationClient");
    Console.WriteLine(Activity.Current?.DisplayName);
    Console.WriteLine(Activity.Current?.ActivityTraceFlags); // the child activity will inherit the ActivityTraceFlags from its parent activity
    testActivity?.Stop();

    activityWithoutSource.Stop();

    Console.WriteLine(exportedActivities.Count());
    if (exportedActivities.Count() > 0)
    {
        Console.WriteLine(exportedActivities[0].OperationName);
    }
}
```

### Verify that request tracing can work with the change in this PR

Currently, we are lacking testcases about request tracing in this repo.

The following code can be used to verify that the request tracing won't be effected.  (OpenTelemetry.Extensions.Hosting and OpenTelemetry.Exporter.InMemory packages are required)

```C#
using System.Diagnostics;
using OpenTelemetry.Trace;

var builder = Host.CreateApplicationBuilder(args);

var client = new ConfigurationClient("<connection-string>");
List<Activity> exportedActivities = new();

builder.Services.AddOpenTelemetry()
    .WithTracing(builder =>
    {
        builder.AddInMemoryExporter(exportedActivities)
            //.AddSource([$"{typeof(ConfigurationClient).Namespace}.*", "ConfigurationProvider"]);
            .AddSource(["ConfigurationProvider"]);
        //.AddSource([$"{typeof(ConfigurationClient).Namespace}.*"]);
    });

var activitySource = new ActivitySource("ConfigurationProvider");

using (var host = builder.Build())
{
    host.Start();

    using (var refreshActivity = activitySource.StartActivity("Refresh"))
    {
        Console.WriteLine($"refreshActivity is null: {refreshActivity == null}");

        var requestTracingActivity = new Activity("Azure.CustomDiagnosticHeaders");
        requestTracingActivity.Start();
        client.GetConfigurationSetting("test"); // the SDK has its own activity source: "Azure.Data.AppConfiguration.ConfigurationClient"
                                                // Add breakpoint in the SDK's CustomHeadersPolicy.cs file can verify that the "CustomDiagnosticHeaders" activity will be catched.
        requestTracingActivity.Stop();
    }

    Console.WriteLine(exportedActivities.Count());
    if (exportedActivities.Count() > 0)
    {
        Console.WriteLine(exportedActivities[0].OperationName);
    }
}
```


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -45,6 +45,7 @@ internal class RequestTracingConstants
         public const string FeatureFlagUsesSeedTag = "Seed";
         public const string FeatureFlagMaxVariantsKey = "MaxVariants";
 
+        public const string AzureAppConfigurationActivitySource = "Microsoft.Extensions.Configuration.AzureAppConfiguration";
```
zhiyuanliang-ms, line 48: The activity source SDK is using is: "Azure.Data.AppConfiguration.ConfigurationClient"

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -50,6 +50,11 @@ public class AzureAppConfigurationOptions
         /// </summary>
         public bool LoadBalancingEnabled { get; set; }
 
+        /// <summary>
+        /// An optional timespan value to set the minimum backoff duration to a value other than the default.
+        /// </summary>
+        public TimeSpan MinBackoffDuration { get; set; } = FailOverConstants.MinBackoffDuration;
```
zhiyuanliang-ms, line 56: Make AzureAppConfigurationOptions.MinBackoffDuration public to faciliate writing testcases in aspire repo (I don't want to use reflection)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -163,304 +165,310 @@ public AzureAppConfigurationProvider(IConfigurationClientManager configClientMan
...
-
-            // Mark all settings have loaded at startup.
-            _isInitialLoadComplete = true;
         }
 
         public async Task RefreshAsync(CancellationToken cancellationToken)
         {
-            // Ensure that concurrent threads do not simultaneously execute refresh operation.
-            if (Interlocked.Exchange(ref _networkOperationsInProgress, 1) == 0)
+            using (var activity = _activitySource.StartActivity(ActivityConstants.RefreshActivityName))
```
zhiyuanliang-ms, line 216: Start the refresh activity only when a real refresh happen

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
zhiyuanliang-ms: Discussed with @jimmyca15 , we decided to use package name: `Microsoft.Extensions.Configuration.AzureAppConfiguration` as activity source name. We have used this pattern for .NET FM.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
amerjusupovic: Are we concerned about users setting this to some very small value? Technically this could be set to 1 tick and backoff is essentially gone

zhiyuanliang-ms: I think if people are intended to set this to a very small value. They must be aware of the potential consequences. Unless they are insane, I don't think they will set it to 1 tick. But I can add a minimum backoff time like 1 seconds.

zhiyuanliang-ms: I just found our testcase was using 1 tick as backoff duration. [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/tests/Tests.AzureAppConfiguration/RefreshTests.cs#L809)

> But I can add a minimum backoff time like 1 seconds.

revert this change

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ActivityConstants.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal class ActivityConstants
```
jimmyca15, line 7:
```suggestion
    internal static class ActivityNames
```

Constants is to vague to be useful for anything. I would also recommend not having a folder named constants.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -158,46 +159,48 @@ public AzureAppConfigurationProvider(IConfigurationClientManager configClientMan
         public override void Load()
         {
             var watch = Stopwatch.StartNew();
-
-            try
+            using (var activity = _activitySource.StartActivity(ActivityConstants.LoadActivityName))
```
jimmyca15, line 162:
```suggestion
            using (var activity = _activitySource.StartActivity(ActivityNames.Load))
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -158,46 +159,48 @@ public AzureAppConfigurationProvider(IConfigurationClientManager configClientMan
...
             {
```
jimmyca15, line 163: Avoid the brackets, save the indentation.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -258,196 +261,199 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                         return;
                     }
 
-                    // Check if initial configuration load had failed
-                    if (_mappedData == null)
+                    using (var activity = _activitySource.StartActivity(ActivityConstants.RefreshActivityName))
                     {
```
jimmyca15, line 265: avoid the brackets, save the indentaiton.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ActivityConstants.cs
zhiyuanliang-ms: image

image

I followed the existing pattern. Should I also update them?

jimmyca15: No, my suggestion is just go with a new class ActivityNames that is clear what it's for.

jimmyca15 (approved):

PR #651, opened 2025-04-17
Separate content type exclusion logic from IsJson extension

amerjusupovic:


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ContentTypeExtensions.cs
```diff
@@ -45,37 +43,43 @@ public static bool IsJson(this ContentType contentType)
...
                     }
+
+                    subTypeSpan = plusIndex == -1 ? ReadOnlySpan<char>.Empty : subTypeSpan.Slice(plusIndex + 1);
                 }
             }
 
             return false;
         }
+
+        public static bool IsFeatureFlag(this ContentType contentType)
```
jimmyca15, line 75: Why don't the ff and kv adapters use these extensions?

amerjusupovic: Good point, I didn't realize since we only used the `ContentType` class in the `JsonKeyValueAdapter`, but I can add it to those other classes

jimmyca15: If we're going to add it, I'd say we should.

amerjusupovic: updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -45,10 +46,24 @@ public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(Configura
...
+            }
+
+            if (setting.Key.StartsWith(FeatureManagementConstants.FeatureFlagMarker))
+            {
+                return true;
+            }
 
-            return string.Equals(contentType, FeatureManagementConstants.ContentType) ||
-                                 setting.Key.StartsWith(FeatureManagementConstants.FeatureFlagMarker);
+            if (setting.ContentType.TryParseContentType(out ContentType contentType))
```
jimmyca15, line 61: This can be simplified to

```
return setting.ContentType.TryParseContentType(out ContentType contentType) &&
    contentType.IsFeatureFlag();
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
```diff
@@ -53,7 +53,7 @@ public bool CanProcess(ConfigurationSetting setting)
 
             if (setting.ContentType.TryParseContentType(out ContentType contentType))
             {
-                return contentType.IsJson();
+                return contentType.IsJson() && !contentType.IsFeatureFlag() && !contentType.IsKeyVaultReference();
```
jimmyca15, line 56: can you multiline that?

jimmyca15 (approved):

jimmyca15 (approved):

PR #620, opened 2025-02-13
Shorten the default timeout of individual call to backend

RichardChen820:
Shorten the default timeout of individual call to backend

PR #657, opened 2025-04-28
Shorten default timeout of individual calls to backend

amerjusupovic:
This is a redo of a PR that was previously merged and reverted: #620

This PR adds a way to dispose of the `HttpClientTransport` class that was added, but otherwise the change is intended to be the same.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 //
 using Azure.Core;
+using Azure.Core.Pipeline;
 using Azure.Data.AppConfiguration;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions;
@@ -10,6 +11,7 @@ using Microsoft.Extensions.Configuration.AzureAppConfiguration.Models;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net.Http;
 using System.Threading.Tasks;
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
@@ -18,10 +20,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     /// Options used to configure the behavior of an Azure App Configuration provider.         
     /// If neither <see cref="Select"/> nor <see cref="SelectSnapshot"/> is ever called, all key-values with no label are included in the configuration provider.
     /// </summary>
-    public class AzureAppConfigurationOptions
+    public class AzureAppConfigurationOptions : IDisposable
     {
         private const int MaxRetries = 2;
         private static readonly TimeSpan MaxRetryDelay = TimeSpan.FromMinutes(1);
+        private static readonly TimeSpan NetworkTimeout = TimeSpan.FromSeconds(10);
         private static readonly KeyValueSelector DefaultQuery = new KeyValueSelector { KeyFilter = KeyFilter.Any, LabelFilter = LabelFilter.Null };
 
         private List<KeyValueWatcher> _individualKvWatchers = new List<KeyValueWatcher>();
@@ -31,6 +34,10 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         private List<KeyValueSelector> _selectors;
         private IConfigurationRefresher _refresher = new AzureAppConfigurationRefresher();
         private bool _selectCalled = false;
+        private HttpClientTransport _clientOptionsTransport = new HttpClientTransport(new HttpClient()
+        {
+            Timeout = NetworkTimeout
+        });
 
         // The following set is sorted in descending order.
         // Since multiple prefixes could start with the same characters, we need to trim the longest prefix first.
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -18,10 +20,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
     /// Options used to configure the behavior of an Azure App Configuration provider.         
     /// If neither <see cref="Select"/> nor <see cref="SelectSnapshot"/> is ever called, all key-values with no label are included in the configuration provider.
     /// </summary>
-    public class AzureAppConfigurationOptions
+    public class AzureAppConfigurationOptions : IDisposable
```
jimmyca15, line 23: When is dispose called on this in a customer app?

amerjusupovic: Missed this, looks like we could call `_options.Dispose` in `AzureAppConfigurationProvider.Dispose`?

jimmyca15: Whoever creates a disposable object should be the one to dispose it. In our case `AzureAppConfigurationOptions` is created internally and doesn't make sense to be disposed where it's created.

So I'm not sure we should be doing it this way at all. I'm leaning towards us doing this via a policy instead of passing an HttpClient.

amerjusupovic: I've looked into it, and it seems like the [`RetryOptions.NetworkTimeout`](https://learn.microsoft.com/en-us/dotnet/api/azure.core.retryoptions.networktimeout?view=azure-dotnet) property also throws a [`TaskCanceledException`](https://stackoverflow.com/questions/78649680/how-to-change-networktimeout-for-defaultazurecredential#:~:text=To%20resolve%20the%20Network%20Timeout%20error%2C%20use%20Retry,and%20used%20below%20code%20to%20generate%20access%20token.) on timeout. It's difficult to find actual documentation on this property, but it appears to function the same.

jimmyca15: Did you check the exception composition? Because right now a check is being added for TaskCanceled with Timeout as an inner exception.

amerjusupovic: Tested it locally, looks like it's not a `TimeoutException` as the inner exception, just another `TaskCanceledException`. I'll update this

jimmyca15: Is there a test that can be added to validate?

amerjusupovic: I've added a test to check that the provider fails over and checks the exception type if there's no replica for failover.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -529,6 +530,7 @@ private static ConfigurationClientOptions GetDefaultClientOptions()
             clientOptions.Retry.MaxRetries = MaxRetries;
             clientOptions.Retry.MaxDelay = MaxRetryDelay;
             clientOptions.Retry.Mode = RetryMode.Exponential;
+            clientOptions.Retry.NetworkTimeout = NetworkTimeout;
```
RichardChen820, line 533: I remember that when I implemented this, I also tried to achieve the goal by setting this property — but I believe there was an issue. If I recall correctly, the reason is that it only applies to retries. The problem is, during app startup, if the very first request times out due to abnormal network operation and exhausts the entire 100s startup timeout, failover still cannot get a chance to be performed.

amerjusupovic: When I run the test I just added, I can look at the `AggregateException` that's inside the `TimeoutException` from startup and see all of the exceptions thrown. When I run it with an invalid connection string like in the test and 2 max retries, I would expect to see 1 `RequestFailedException` due to not finding the store, then 2 requests failing on retry because of network timeout. Instead, I just see 3 requests failing due to network timeout of 1 tick. I can add to the test to be more explicit about checking what all of the exceptions are, let me know if I missed something about how it works.

RichardChen820: After verifying on my side, I think you are right. It applys to each single network operation, no matter it's retry or not.

RichardChen820 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1234,7 +1234,7 @@ private bool IsFailOverable(AggregateException ex)
         {
             TaskCanceledException tce = ex.InnerExceptions?.LastOrDefault(e => e is TaskCanceledException) as TaskCanceledException;
 
-            if (tce != null && tce.InnerException is TaskCanceledException)
+            if (tce != null)
```
jimmyca15, line 1237: This variable isn't needed. You could change LastOrDefault to Any above.

jimmyca15 (approved):

jimmyca15 (approved):

Issue #627, opened 2025-02-24
Defining features using MicrosoftSchema in AzureAppConfiguration and appsetting.json does not work when not using variants


emilssonn:
When AzureAppConfiguration loads features/config from Azure it checks for the existence of Variants, Allocation or Telemetry to decide what feature flag schema to use. If none of these are found it loads the config/feature in the DotnetSchema instead of the MicrosoftSchema. This makes merging features defined using the MicrosoftSchema in other config sources and the MicrosoftSchema impossible if not using a variant flag.

When following https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#feature-flag-declaration and then adding AzureAppConfiguration the feature flags defined in appsettings.json using MicrosoftSchema cannot be overridden using AzureAppConfiguration if not using variants.

Steps to reproduce:
1. Add feature flags in appsettings.json using the MicrosoftSchema and not using variants.
2. Connect to AzureAppConfiguration.
3. Add the same feature flag (not variant)
4. In AzureAppConfiguration, enabled/disable the feature flag.
5. Nothing changes in the application (restart to reload if needed)

There should be some way (or the library should have more logic) to force the library to always call ProcessMicrosoftSchemaFeatureFlag in this section:
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/7dc547c594f579d2c7755bba90b0babd6f182af0/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs#L34

zhiyuanliang-ms:
Hi, @emilssonn Thanks for bringing up this issue.

You are right. The root cause of this bug is that if there is no variant or telemetry related fields in the feature flag, the flag will be converted into .NET schema.

![Image](https://github.com/user-attachments/assets/f54228e2-8e39-4665-92c2-809a87eb1e9f)

And the feature management library will prioritize using Microsoft schema [code](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L49)

zhiyuanliang-ms:
[...]

Issue #536, opened 2024-03-25
Add new API to reload on changes to key-values


amerjusupovic:
Currently, the provider only has the `AzureAppConfigurationRefreshOptions.Register` method as a way to monitor individual key-values for refresh. By adding a new API, we can enable monitoring for all of the key-values to be loaded by the provider. Then, when a refresh is triggered, the provider will reload the configuration if any of the selected key-values have changed in App Configuration. You can only call one of `RegisterAll` or `Register`, not both.

### Proposed API: RegisterAll
```csharp
config.AddAzureAppConfiguration(options => {
     options.ConfigureRefresh(refresh =>
     {
         refresh.RegisterAll();
     });
});
```

In the example above, since the provider selects all key-values with label `null` by default if no `Select` or `SelectSnapshot` statement is present, all of the key-values that have no label will be monitored when `RegisterAll` is called. If any of them have changed since the last refresh attempt, the configuration will be reloaded.

```csharp
config.AddAzureAppConfiguration(options => {
     options.Select("TestApp*", "prod");
     options.ConfigureRefresh(refresh =>
     {
         refresh.RegisterAll();
     });
});
```

Now, changes to key-values that start with `TestApp` and have the label `prod` will trigger a configuration reload.

PR #658, opened 2025-04-30
Add index offset to Microsoft schema feature flags

amerjusupovic:
Because issues like #627 have mentioned problems combining feature management from multiple sources like json files, this PR adds an index offset to all feature flags parsed by the provider. There is an additional offset between each provider to ensure there is no overlap between those sources as well.


zhiyuanliang-ms:
The new solution is to have feature manager merge `feature_flags` array from different configuration sources.

[#536](https://github.com/microsoft/FeatureManagement-Dotnet/pull/536)

PR #661, opened 2025-05-12
Add .gitattributes and normalize line endings

amerjusupovic:


zhiyuanliang-ms (approved):

zhiyuanliang-ms:
We should also add this for other repos. Could you help do this for .net FM? I will do it for JS repos.

PR #665, opened 2025-05-15
Add testcase for refresh activity

zhiyuanliang-ms:
Add testcase for #645


zhiyuanliang-ms:
The testcases related to ActivitySource needs to be executed in a standalone process.

amerjusupovic (approved):

PR #676, opened 2025-06-17
Add environment variable to disable feature management schema compatibility logic

amerjusupovic:
This PR adds the environment variable `AZURE_APP_CONFIGURATION_FM_SCHEMA_COMPATIBILITY_DISABLED`. When set to `true`, the provider will process and output all feature flags as the newer [Microsoft schema](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFlag.v2.0.0.schema.json). If false, the provider will function as it does today where a flag is only processed in the Microsoft schema if any of the `variants`, `allocation`, or `telemetry` properties are present.


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -18,12 +19,20 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManage
...
         private int _featureFlagIndex = 0;
+        private bool _fmSchemaCompatibilityDisabled = false;
 
         public FeatureManagementKeyValueAdapter(FeatureFlagTracing featureFlagTracing)
         {
             _featureFlagTracing = featureFlagTracing ?? throw new ArgumentNullException(nameof(featureFlagTracing));
+
+            try
+            {
+                _fmSchemaCompatibilityDisabled = bool.TryParse(Environment.GetEnvironmentVariable(DisableFmSchemaCompatibilityEnvironmentVariable), out bool disabled) ? disabled : false;
```
zhiyuanliang-ms, line 33: I think we should only put the Environment.GetEnvironmentVariable in the try {}.
like [this](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L149)

zhiyuanliang-ms (dismissed):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
```diff
@@ -18,12 +19,24 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.FeatureManage
 {
     internal class FeatureManagementKeyValueAdapter : IKeyValueAdapter
     {
+        private const string DisableFmSchemaCompatibilityEnvironmentVariable = "AZURE_APP_CONFIGURATION_FM_SCHEMA_COMPATIBILITY_DISABLED";
```
jimmyca15, line 22: I think it would be a good idea if we put all the environment variables that customers can use in a single file/class. `EnvironmentVariables`. I think we have at least one other for disabling telemetry.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs
zhiyuanliang-ms: Added a new public class `AzureAppConfigurationEnvironmentVariables`.

Should we also move `public const string RequestTracingDisabledEnvironmentVariable = "AZURE_APP_CONFIGURATION_TRACING_DISABLED";` there?

Now it is internal: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs#L8

jimmyca15: Yes. The class should be internal. My suggestion is to not expose these as public API surface from the library, but rather to have them more easily referenceable in source.

jimmyca15 (approved):

PR #689, opened 2025-08-20
Adding Snapshot References Feature

t-vvidyasaga:
**Overview**
Adds Snapshot References to the .NET configuration provider for dynamic updates via the cloud portal, with automatic resolution and no code changes.

**Key Changes**
**AzureAppConfigurationProvider.cs**
- Detects snapshot references via custom content type.
- Implements ParseSnapshotReference() for JSON parsing.
- Adds Resolve() for snapshot resolution.
- Updates LoadSelected() to auto-resolve references.
- Triggers refreshAll for snapshot references.

**SnapshotReference.cs (New)**
Model for snapshot reference: name, client, cancellation token.
**SnapshotReferenceConstants.cs (New)**
Shared constants for content type and JSON fields.

**Test Coverage**
Includes unit and integration tests for resolution, refresh behavior, and edge cases.

**Key Callouts**
Resolution: Auto-detected by content type; lexicographic “last wins” for duplicates.
Exceptions:
- Invalid JSON → FormatException
- Invalid composition → InvalidOperationException

Refresh: Always triggers refreshAll, regardless of settings.
JSON Handling: Ignores extra fields, validates snapshot_name, handles empty values.
Edge Cases: Non-existent references and empty names return empty config without exceptions.

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
@@ -1284,7 +1284,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
                         do
                         {
-                            UpdateClientBackoffStatus(previousEndpoint, success);
+                            UpdateClientBackoffStatus(_configClientManager.GetEndpointForClient(currentClient), success);
 
                             clientEnumerator.MoveNext();
 
@@ -1400,6 +1400,8 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                     _requestTracingOptions.FeatureManagementVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAssemblyName);
 
                     _requestTracingOptions.FeatureManagementAspNetCoreVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.FeatureManagementAspNetCoreAssemblyName);
+
+                    _requestTracingOptions.AspireComponentVersion = TracingUtils.GetAssemblyVersion(RequestTracingConstants.AspireComponentAssemblyName);
 
                     if (TracingUtils.GetAssemblyVersion(RequestTracingConstants.SignalRAssemblyName) != null)
                     {
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
@@ -24,6 +24,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         public const string EnvironmentKey = "Env";
         public const string FeatureManagementVersionKey = "FMVer";
         public const string FeatureManagementAspNetCoreVersionKey = "FMANCVer";
+        public const string AspireComponentVersionKey = "DNACVer";
         public const string DevEnvironmentValue = "Dev";
         public const string KeyVaultConfiguredTag = "UsesKeyVault";
         public const string KeyVaultRefreshConfiguredTag = "RefreshesKeyVault";
@@ -53,6 +54,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
         public const string FeatureManagementAssemblyName = "Microsoft.FeatureManagement";
         public const string FeatureManagementAspNetCoreAssemblyName = "Microsoft.FeatureManagement.AspNetCore";
+        public const string AspireComponentAssemblyName = "Aspire.Microsoft.Extensions.Configuration.AzureAppConfiguration";
         public const string SignalRAssemblyName = "Microsoft.AspNetCore.SignalR";
 
         public const string Delimiter = "+";
--- a/src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
+++ b/src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
@@ -21,7 +21,7 @@
   <!-- Nuget Package Version Settings -->
 
   <PropertyGroup>
-    <OfficialVersion>8.3.0</OfficialVersion>
+    <OfficialVersion>8.4.0</OfficialVersion>
   </PropertyGroup>
   
   <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'=='Official'">
--- a/src/Microsoft.Azure.AppConfiguration.Functions.Worker/Microsoft.Azure.AppConfiguration.Functions.Worker.csproj
+++ b/src/Microsoft.Azure.AppConfiguration.Functions.Worker/Microsoft.Azure.AppConfiguration.Functions.Worker.csproj
@@ -24,7 +24,7 @@
   <!-- Nuget Package Version Settings -->
 
   <PropertyGroup>
-    <OfficialVersion>8.3.0</OfficialVersion>
+    <OfficialVersion>8.4.0</OfficialVersion>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'=='Official'">
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/JsonKeyValueAdapter.cs
@@ -15,6 +15,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
     internal class JsonKeyValueAdapter : IKeyValueAdapter
     {
+        private static readonly JsonDocumentOptions JsonParseOptions = new JsonDocumentOptions
+        {
+            CommentHandling = JsonCommentHandling.Skip
+        };
+
         public Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, Uri endpoint, Logger logger, CancellationToken cancellationToken)
         {
             if (setting == null)
@@ -28,7 +33,7 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 
             try
             {
-                using (JsonDocument document = JsonDocument.Parse(rootJson))
+                using (JsonDocument document = JsonDocument.Parse(rootJson, JsonParseOptions))
                 {
                     keyValuePairs = new JsonFlattener().FlattenJson(document.RootElement);
                 }
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj
@@ -38,7 +38,7 @@
   <!-- Nuget Package Version Settings -->
   
   <PropertyGroup>
-    <OfficialVersion>8.3.0</OfficialVersion>
+    <OfficialVersion>8.4.0</OfficialVersion>
   </PropertyGroup>
   
   <PropertyGroup Condition="'$(CDP_PATCH_NUMBER)'!='' AND '$(CDP_BUILD_TYPE)'=='Official'">
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/RequestTracingOptions.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/RequestTracingOptions.cs
@@ -52,6 +52,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
         public string FeatureManagementAspNetCoreVersion { get; set; }
 
         /// <summary>
+        /// Version of the Aspire.Microsoft.Extensions.Configuration.AzureAppConfiguration assembly, if present in the application.
+        /// </summary>
+        public string AspireComponentVersion { get; set; }
+
+        /// <summary>
         /// Flag to indicate whether Microsoft.AspNetCore.SignalR assembly is present in the application.
         /// </summary>
         public bool IsSignalRUsed { get; set; } = false;
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
@@ -181,6 +181,11 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
                 correlationContextKeyValues.Add(new KeyValuePair<string, string>(RequestTracingConstants.FeatureManagementAspNetCoreVersionKey, requestTracingOptions.FeatureManagementAspNetCoreVersion));
             }
 
+            if (requestTracingOptions.AspireComponentVersion != null)
+            {
+                correlationContextKeyValues.Add(new KeyValuePair<string, string>(RequestTracingConstants.AspireComponentVersionKey, requestTracingOptions.AspireComponentVersion));
+            }
+
             if (requestTracingOptions.UsesAnyTracingFeature())
             {
                 correlationContextKeyValues.Add(new KeyValuePair<string, string>(RequestTracingConstants.FeaturesKey, requestTracingOptions.CreateFeaturesString()));
--- a/tests/Tests.AzureAppConfiguration/Unit/FailoverTests.cs
+++ b/tests/Tests.AzureAppConfiguration/Unit/FailoverTests.cs
@@ -415,5 +415,89 @@ namespace Tests.AzureAppConfiguration
                 ae2.InnerExceptions.All(ex => ex is TaskCanceledException) &&
                 ae2.InnerException is TaskCanceledException tce);
         }
+
+        [Fact]
+        public async Task FailOverTests_AllClientsBackedOffAfterNonFailoverableException()
+        {
+            IConfigurationRefresher refresher = null;
+            var mockResponse = new Mock<Response>();
+
+            // Setup first client - succeeds on startup, fails with 404 (non-failoverable) on first refresh
+            var mockClient1 = new Mock<ConfigurationClient>();
+            mockClient1.SetupSequence(c => c.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()))
+                       .Returns(new MockAsyncPageable(Enumerable.Empty<ConfigurationSetting>().ToList()))
+                       .Throws(new RequestFailedException(412, "Request failed."))
+                       .Throws(new RequestFailedException(412, "Request failed."));
+            mockClient1.SetupSequence(c => c.GetConfigurationSettingAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()))
+                        .Returns(Task.FromResult(Response.FromValue<ConfigurationSetting>(kv, mockResponse.Object)))
+                        .Throws(new RequestFailedException(412, "Request failed."))
+                        .Throws(new RequestFailedException(412, "Request failed."));
+            mockClient1.SetupSequence(c => c.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()))
+                       .Throws(new RequestFailedException(412, "Request failed."))
+                       .Throws(new RequestFailedException(412, "Request failed."));
+            mockClient1.Setup(c => c.Equals(mockClient1)).Returns(true);
+
+            // Setup second client - succeeds on startup, should not be called during refresh
+            var mockClient2 = new Mock<ConfigurationClient>();
+            mockClient2.Setup(c => c.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()))
+                       .Returns(new MockAsyncPageable(Enumerable.Empty<ConfigurationSetting>().ToList()));
+            mockClient2.Setup(c => c.GetConfigurationSettingAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()))
+                       .Returns(Task.FromResult(Response.FromValue<ConfigurationSetting>(kv, mockResponse.Object)));
+            mockClient2.Setup(c => c.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()))
+                       .Returns(Task.FromResult(Response.FromValue<ConfigurationSetting>(kv, mockResponse.Object)));
+            mockClient2.Setup(c => c.Equals(mockClient2)).Returns(true);
+
+            ConfigurationClientWrapper cw1 = new ConfigurationClientWrapper(TestHelpers.PrimaryConfigStoreEndpoint, mockClient1.Object);
+            ConfigurationClientWrapper cw2 = new ConfigurationClientWrapper(TestHelpers.SecondaryConfigStoreEndpoint, mockClient2.Object);
+
+            var clientList = new List<ConfigurationClientWrapper>() { cw1, cw2 };
+            var configClientManager = new ConfigurationClientManager(clientList);
+
+            // Verify 2 clients are available
+            Assert.Equal(2, configClientManager.GetClients().Count());
+
+            // Act & Assert - Build configuration successfully with both clients
+            var config = new ConfigurationBuilder()
+                .AddAzureAppConfiguration(options =>
+                {
+                    options.ClientManager = configClientManager;
+                    options.Select("TestKey*");
+                    options.ConfigureRefresh(refreshOptions =>
+                    {
+                        refreshOptions.Register("TestKey1", "label")
+                            .SetRefreshInterval(TimeSpan.FromSeconds(1));
+                    });
+
+                    options.ReplicaDiscoveryEnabled = false;
+                    refresher = options.GetRefresher();
+                }).Build();
+
+            // First refresh - should call client 1 and fail with non-failoverable exception
+            // This should cause all clients to be backed off
+            await Task.Delay(1500);
+            await refresher.TryRefreshAsync();
+
+            // Verify that client 1 was called during the first refresh
+            mockClient1.Verify(mc => mc.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()), Times.Exactly(1));
+            mockClient1.Verify(mc => mc.GetConfigurationSettingAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()), Times.Exactly(1));
+            mockClient1.Verify(mc => mc.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()), Times.Exactly(1));
+
+            // Verify that client 2 was not called during the first refresh
+            mockClient2.Verify(mc => mc.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()), Times.Never);
+            mockClient2.Verify(mc => mc.GetConfigurationSettingAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()), Times.Never);
+            mockClient2.Verify(mc => mc.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()), Times.Never);
+
+            // Second refresh - no clients should be called as all are backed off
+            await Task.Delay(1500);
+            await refresher.TryRefreshAsync();
+
+            // Verify that no additional calls were made to any client during the second refresh
+            mockClient1.Verify(mc => mc.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()), Times.Exactly(1));
+            mockClient1.Verify(mc => mc.GetConfigurationSettingAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()), Times.Exactly(1));
+            mockClient1.Verify(mc => mc.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()), Times.Exactly(1));
+            mockClient2.Verify(mc => mc.GetConfigurationSettingsAsync(It.IsAny<SettingSelector>(), It.IsAny<CancellationToken>()), Times.Never);
+            mockClient2.Verify(mc => mc.GetConfigurationSettingAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()), Times.Never);
+            mockClient2.Verify(mc => mc.GetConfigurationSettingAsync(It.IsAny<ConfigurationSetting>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()), Times.Never);
+        }
     }
 }
--- a/tests/Tests.AzureAppConfiguration/Unit/JsonContentTypeTests.cs
+++ b/tests/Tests.AzureAppConfiguration/Unit/JsonContentTypeTests.cs
@@ -276,6 +276,90 @@ namespace Tests.AzureAppConfiguration
             Assert.False(jsonKeyValueAdapter.CanProcess(setting));
         }
 
+        [Fact]
+        public void JsonContentTypeTests_LoadJsonValuesWithComments()
+        {
+            List<ConfigurationSetting> _kvCollection = new List<ConfigurationSetting>
+            {
+                // Test various comment styles and positions
+                ConfigurationModelFactory.ConfigurationSetting(
+                    key: "MixedCommentStyles",
+                    value: @"{
+                        // Single line comment at start
+                        ""ApiSettings"": {
+                            ""BaseUrl"": ""https://api.example.com"", // Inline single line
+                            /* Multi-line comment
+                               spanning multiple lines */
+                            ""ApiKey"": ""secret-key"",
+                            ""Endpoints"": [
+                                // Comment before array element
+                                ""/users"",
+                                /* Comment between elements */
+                                ""/orders"",
+                                ""/products"" // Comment after element
+                            ]
+                        },
+                        // Test edge cases
+                        ""StringWithSlashes"": ""This is not a // comment"",
+                        ""StringWithStars"": ""This is not a /* comment */"",
+                        ""UrlValue"": ""https://example.com/path"", // This is a real comment
+                        ""EmptyComment"": ""value"", //
+                        /**/
+                        ""AfterEmptyComment"": ""value2""
+                        /* Final multi-line comment */
+                    }",
+                    contentType: "application/json"),
+                // Test invalid JSON with comments
+                ConfigurationModelFactory.ConfigurationSetting(
+                    key: "InvalidJsonWithComments",
+                    value: @"// This is a comment
+                        { invalid json structure
+                        // Another comment
+                        missing quotes and braces",
+                    contentType: "application/json"),
+                // Test only comments (should be invalid JSON)
+                ConfigurationModelFactory.ConfigurationSetting(
+                    key: "OnlyComments",
+                    value: @"
+                        // Just comments
+                        /* No actual content */
+                    ",
+                    contentType: "application/json")
+            };
+
+            var mockClientManager = GetMockConfigurationClientManager(_kvCollection);
+
+            var config = new ConfigurationBuilder()
+                .AddAzureAppConfiguration(options => options.ClientManager = mockClientManager)
+                .Build();
+
+            // Verify mixed comment styles are properly parsed
+            Assert.Equal("https://api.example.com", config["MixedCommentStyles:ApiSettings:BaseUrl"]);
+            Assert.Equal("secret-key", config["MixedCommentStyles:ApiSettings:ApiKey"]);
+            Assert.Equal("/users", config["MixedCommentStyles:ApiSettings:Endpoints:0"]);
+            Assert.Equal("/orders", config["MixedCommentStyles:ApiSettings:Endpoints:1"]);
+            Assert.Equal("/products", config["MixedCommentStyles:ApiSettings:Endpoints:2"]);
+
+            // Verify edge cases where comment-like text appears in strings
+            Assert.Equal("This is not a // comment", config["MixedCommentStyles:StringWithSlashes"]);
+            Assert.Equal("This is not a /* comment */", config["MixedCommentStyles:StringWithStars"]);
+            Assert.Equal("https://example.com/path", config["MixedCommentStyles:UrlValue"]);
+            Assert.Equal("value", config["MixedCommentStyles:EmptyComment"]);
+            Assert.Equal("value2", config["MixedCommentStyles:AfterEmptyComment"]);
+
+            // Invalid JSON should fall back to string value
+            Assert.Equal(@"// This is a comment
+                        { invalid json structure
+                        // Another comment
+                        missing quotes and braces", config["InvalidJsonWithComments"]);
+
+            // Only comments should be treated as string value (invalid JSON)
+            Assert.Equal(@"
+                        // Just comments
+                        /* No actual content */
+                    ", config["OnlyComments"]);
+        }
+
         private IConfigurationClientManager GetMockConfigurationClientManager(List<ConfigurationSetting> _kvCollection)
         {
             var mockResponse = new Mock<Response>();
--- a/tests/Tests.AzureAppConfiguration/Unit/Tests.cs
+++ b/tests/Tests.AzureAppConfiguration/Unit/Tests.cs
@@ -352,10 +352,12 @@ namespace Tests.AzureAppConfiguration
         [Fact]
         public void TestActivitySource()
         {
+            string activitySourceName = Guid.NewGuid().ToString();
+
             var _activities = new List<Activity>();
             var _activityListener = new ActivityListener
             {
-                ShouldListenTo = source => source.Name == "Microsoft.Extensions.Configuration.AzureAppConfiguration",
+                ShouldListenTo = source => source.Name == activitySourceName,
                 Sample = (ref ActivityCreationOptions<ActivityContext> options) => ActivitySamplingResult.AllData,
                 ActivityStarted = activity => _activities.Add(activity),
             };
@@ -371,7 +373,11 @@ namespace Tests.AzureAppConfiguration
                 .ReturnsAsync(Response.FromValue(_kv, mockResponse.Object));
 
             var config = new ConfigurationBuilder()
-                .AddAzureAppConfiguration(options => options.ClientManager = TestHelpers.CreateMockedConfigurationClientManager(mockClient.Object))
+                .AddAzureAppConfiguration(options =>
+                {
+                    options.ClientManager = TestHelpers.CreateMockedConfigurationClientManager(mockClient.Object);
+                    options.ActivitySourceName = activitySourceName;
+                })
                 .Build();
 
             Assert.Contains(_activities, a => a.OperationName == "Load");


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
```diff
@@ -0,0 +1,27 @@
+using Azure.Data.AppConfiguration;
```
austintolani, line 1: Missing copyright info

tests/Tests.AzureAppConfiguration/Unit/SnapshotReferencesTests.cs
```diff
@@ -0,0 +1,538 @@
+using Azure;
```
austintolani, line 1: Missing copyright info

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Threading;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Models
```
austintolani, line 4: Generally, the namespace should match the file path

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
```diff
@@ -0,0 +1,27 @@
+using Azure.Data.AppConfiguration;
+using System.Threading;
```
austintolani, line 2: I think we can drop "Azure" from the "AzureSnapshotReference" in this filepath. "SnapshotReference" or "AzureAppConfigurationSnapshotReference" makes more sense to me

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
```diff
@@ -0,0 +1,27 @@
...
+        public string SnapshotName { get; set; }
+
+        public ConfigurationClient Client { get; set; }
+
+        public CancellationToken CancellationToken { get; set; }
+
+        /// <param name="snapshotName">The name of the snapshot.</param>
+        /// <param name="client">The configuration client.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        public SnapshotReference(string snapshotName, ConfigurationClient client, CancellationToken cancellationToken)
```
austintolani, line 20: Seems like this could be simplified to use object initializer syntax or a primary constructor

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
-                        throw new InvalidOperationException($"Could not find snapshot with name '{loadOption.SnapshotName}'.", rfe);
+                        data[resolvedSetting.Key] = resolvedSetting.Value;
                     }
+                }
+            }
 
-                    if (snapshot.SnapshotComposition != SnapshotComposition.Key)
+            return data;
+        }
+        private string ParseSnapshotReference(ConfigurationSetting setting)
```
austintolani, line 930: Nit: newline between method declarations

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
+                        // Skip unknown properties
+                        reader.Skip();
+                    }
                 }
+
+                return null; // Snapshot name property not found
+            }
+            catch (JsonException jsonEx)
+            {
+                // Handles malformed JSON with appropriate exception
```
austintolani, line 981: I think this comment can be removed, the intention is clear from the code

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
+                throw new FormatException($"Invalid snapshot reference format. The value is not valid JSON.", jsonEx);
             }
+        }
 
-            return data;
+        private async Task<Dictionary<string, ConfigurationSetting>> Resolve(SnapshotReference snapshotReference)
```
austintolani, line 986: async method should take in a `CancellationToken`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
+            if (snapshotReference.SnapshotName == null)
+            {
+                return resolvedSettings;
+            }
+
+            ConfigurationSnapshot snapshot;
+
+            try
+            {
+                snapshot = await snapshotReference.Client.GetSnapshotAsync(snapshotReference.SnapshotName).ConfigureAwait(false);
```
austintolani, line 999: We should pass a `CancellationToken` to async methods

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReferenceConstants.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
```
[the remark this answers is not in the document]
jimmyca15: It is common to see namespaces match directory structure. But it is not a requirement. Namespace for a type should be decided first, directory structure is secondary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// Represents a reference to a configuration snapshot that needs to be resolved.
+    /// </summary>
+    internal class SnapshotReference
+    {
+        public string SnapshotName { get; set; }
+
+        public ConfigurationClient Client { get; set; }
+
+        public CancellationToken CancellationToken { get; set; }
```
jimmyca15, line 15: I am surprised to see CancellationToken as a property on the snapshot reference. Why is that?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ContentTypeExtensions.cs
```diff
@@ -81,5 +81,10 @@ public static bool IsKeyVaultReference(this ContentType contentType)
         {
             return contentType.MediaType.Equals(KeyVaultConstants.ContentType);
         }
+
+        public static bool IsSnapshotReference(this ContentType contentType)
+        {
+            return contentType.MediaType.Equals(SnapshotReferenceConstants.ContentType);
```
jimmyca15, line 87: It's a bit odd that `IsAi` and `IsAiChatCompletion` above are more complex. Did you verify this works as expected?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
zhiyuanliang-ms, line 1: The folder name should be `SnapshotReference` instead of `AzureSnapshotReference`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
zhiyuanliang-ms: Vote for `SnapshotReference`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
zhiyuanliang-ms: Suggest using: `Microsoft.Extensions.Configuration.AzureAppConfiguration.SnapshotReference`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
```diff
@@ -0,0 +1,27 @@
+using Azure.Data.AppConfiguration;
+using System.Threading;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Models
+{
+    /// <summary>
+    /// Represents a reference to a configuration snapshot that needs to be resolved.
+    /// </summary>
+    internal class SnapshotReference
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Agree.

It doesn't make sense to couple ConfigurationClient with snapshot reference.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
+        private string ParseSnapshotReference(ConfigurationSetting setting)
+        {
+            if (setting == null || string.IsNullOrWhiteSpace(setting.Value))
+            {
+                return null;
+            }
+
+            try
+            {
+                Utf8JsonReader reader = new Utf8JsonReader(System.Text.Encoding.UTF8.GetBytes(setting.Value));
```
zhiyuanliang-ms, line 939:
```suggestion
                var reader = new Utf8JsonReader(System.Text.Encoding.UTF8.GetBytes(setting.Value));
```

Please also follow this convention in other place: var xxx = new SomeClass();

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
-                        cancellationToken);
-
-                    await CallWithRequestTracing(async () =>
+                    if (reader.GetString() == SnapshotReferenceConstants.SnapshotReferenceJsonPropertyName)
                     {
-                        await foreach (ConfigurationSetting setting in settingsEnumerable.ConfigureAwait(false))
+                        if (reader.Read() && reader.TokenType == JsonTokenType.String)
                         {
-                            data[setting.Key] = setting;
+                            string snapshotName = reader.GetString();
```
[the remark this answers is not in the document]
zhiyuanliang-ms: We had the similar code in [`AzureKeyVaultKeyValueAdapter.cs`](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs#L119)

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhiyuanliang-ms: How about renaming this method to ResolveSnapshotReference?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
+
+            ConfigurationSnapshot snapshot;
+
+            try
+            {
+                snapshot = await snapshotReference.Client.GetSnapshotAsync(snapshotReference.SnapshotName).ConfigureAwait(false);
+            }
+            catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.NotFound)
+            {
+
```
zhiyuanliang-ms, line 1003: Should we log some warning here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
+            if (snapshot.SnapshotComposition != SnapshotComposition.Key)
+            {
+                throw new InvalidOperationException($"{nameof(snapshot.SnapshotComposition)} for the selected snapshot with name '{snapshot.Name}' must be 'key', found '{snapshot.SnapshotComposition}'.");
+            }
+
+            IAsyncEnumerable<ConfigurationSetting> settingsEnumerable = snapshotReference.Client.GetConfigurationSettingsForSnapshotAsync(
+                snapshotReference.SnapshotName,
+                snapshotReference.CancellationToken);
+
+            await CallWithRequestTracing(async () =>
```
zhiyuanliang-ms, line 1016: Why is `CallWithRequestTracing` used here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
austintolani: AzureKeyVault is the name of an (external) service. In this case, a snapshot reference is a resource internal AppConfig so in my view it doesn't need to be qualified further

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReferenceConstants.cs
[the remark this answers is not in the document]
t-vvidyasaga: Updated namespace based on type

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureSnapshotReference/SnapshotReference.cs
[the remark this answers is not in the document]
t-vvidyasaga: Fixed code so now `ConfigurationClient` and `CancellationToken` directly pass into `Resolve(SnapshotReference snapshotReference, ConfigurationClient client, CancellationToken cancellationToken)`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
t-vvidyasaga: > How about renaming this method to ResolveSnapshotReference?
I chose to name the method Resolve instead of ResolveSnapshotReference because it loads snapshot values for both a snapshot reference and .SelectSnapshot("<SnapshotName>"). That said, if the name feels misleading or unclear, I’m happy to rename it!

jimmyca15: > How about renaming this method to ResolveSnapshotReference?

It takes a snapshot reference as a parameter. There's no need to put it in the method name.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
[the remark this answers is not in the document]
austintolani: Looks like this moved to `SnapshotReferenceParser`. Resolving.

tests/Tests.AzureAppConfiguration/Unit/SnapshotReferencesTests.cs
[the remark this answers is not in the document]
t-vvidyasaga: Yes, ofc! I believe my latest commit should already have this change.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +913,61 @@ await CallWithRequestTracing(async () =>
                 }
                 else
                 {
-                    ConfigurationSnapshot snapshot;
+                    // Load snapshot data
+                    SnapshotReference snapshotReference = new SnapshotReference(loadOption.SnapshotName);
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Using `var` when the type is obvious. Same comment for other places. e.g. line 880

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ContentTypeExtensions.cs
avanigupta: I just realized that we only match the profile for AI and AIChatCompletion content types. But we match the entire content type for feature flags and key vault refs. Is that because we want to be more relaxed for AI content type tracking?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ContentTypeExtensions.cs
```diff
@@ -81,5 +82,10 @@ public static bool IsKeyVaultReference(this ContentType contentType)
         {
             return contentType.MediaType.Equals(KeyVaultConstants.ContentType);
         }
+
+        public static bool IsSnapshotReference(this ContentType contentType)
```
avanigupta, line 86: This is not being used currently. I'm guessing the request tracing logic is not updated in this PR yet?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        /// <summary>
+        /// Parses a snapshot name from a snapshot reference configuration setting.
+        /// /// </summary>
+        /// <param name="setting">The configuration setting containing the snapshot reference JSON.</param>
+        /// <returns>The snapshot name if found and valid; otherwise, null.</returns>
+        /// <exception cref="FormatException">Thrown when the setting contains invalid JSON or invalid snapshot reference format.</exception>
+        public static string ParseSnapshotName(ConfigurationSetting setting)
+        {
+            if (setting == null || string.IsNullOrWhiteSpace(setting.Value))
```
avanigupta, line 23: If setting is null, we should throw `ArgumentNullException`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
```diff
@@ -0,0 +1,76 @@
...
+                            string snapshotName = reader.GetString();
+                            if (!string.IsNullOrEmpty(snapshotName))
+                            {
+                                return snapshotName;
+                            }
+                        }
+                        else
+                        {
+                            // Invalid snapshot name
+                            throw new FormatException($"Invalid snapshot reference format. The 'snapshot_name' property must be a string value, but found {reader.TokenType}.");
```
avanigupta, line 58: Same here, error message to be defined in `ErrorMessages` class

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
```diff
@@ -0,0 +1,76 @@
...
+                    {
+                        continue;
+                    }
+
+                    if (reader.GetString() == SnapshotReferenceConstants.SnapshotReferenceJsonPropertyName)
+                    {
+                        if (reader.Read() && reader.TokenType == JsonTokenType.String)
+                        {
+                            string snapshotName = reader.GetString();
+                            if (!string.IsNullOrEmpty(snapshotName))
```
avanigupta, line 50: This check is not needed. If snapshotName is null, we still end up returning null at the end of this method. Simply return ` reader.GetString()` and check for null/empty at caller site

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
```diff
@@ -0,0 +1,76 @@
...
+                        // Skip unknown properties
+                        reader.Skip();
+                    }
+                }
+
+                return null; // Snapshot name property not found
+            }
+            catch (JsonException jsonEx)
+            {
+                throw new FormatException($"Invalid snapshot reference format. The value is not valid JSON.", jsonEx);
```
avanigupta, line 72: Const error message

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +913,61 @@ await CallWithRequestTracing(async () =>
...
+                await foreach (ConfigurationSetting setting in settingsEnumerable.ConfigureAwait(false))
+                {
+                    resolvedSettings[setting.Key] = setting;
+                }
+            }).ConfigureAwait(false);
+
+            return resolvedSettings;
         }
 
         private async Task<Dictionary<KeyValueIdentifier, ConfigurationSetting>> LoadKeyValuesRegisteredForRefresh(
```
avanigupta, line 973: If a snapshot reference is not loaded via normal Select statement but is registered for refresh as a sentinel key, that snapshot reference will be loaded in `LoadKeyValuesRegisteredForRefresh` method. We need to resolve such references too.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Not necessary - non-existent snapshot is treated the same as non-existent key-value

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -874,6 +875,19 @@ await CallWithRequestTracing(async () =>
 
                             foreach (ConfigurationSetting setting in page.Values)
                             {
+                                if (setting.ContentType == SnapshotReferenceConstants.ContentType)
+                                {
+                                    SnapshotReference snapshotReference = new SnapshotReference(SnapshotReferenceParser.ParseSnapshotName(setting));
+                                    Dictionary<string, ConfigurationSetting> resolvedSettings = await Resolve(snapshotReference, client, cancellationToken).ConfigureAwait(false);
```
avanigupta, line 881: If you pass `data` dictionary to the helper function, you can update it in the helper method directly instead of doing it here.
The logic here can be simplified to the following:
```
if (setting.ContentType == SnapshotReferenceConstants.ContentType && setting.Value != null && setting.Value.TryParseSnapshotName(out string snapshotName))
{
    await LoadSnapshotData(snapshotName, data, client, cancellationToken).ConfigureAwait(false);

    continue;
}
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// </summary>
+    internal static class SnapshotReferenceParser
+    {
+        /// <summary>
+        /// Parses a snapshot name from a snapshot reference configuration setting.
+        /// /// </summary>
+        /// <param name="setting">The configuration setting containing the snapshot reference JSON.</param>
+        /// <returns>The snapshot name if found and valid; otherwise, null.</returns>
+        /// <exception cref="FormatException">Thrown when the setting contains invalid JSON or invalid snapshot reference format.</exception>
+        public static string ParseSnapshotName(ConfigurationSetting setting)
```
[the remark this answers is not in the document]
avanigupta: The idea is to not throw while parsing snapshot reference. If there is a format exception, instead of throwing, we can return false. This way, the caller can treat the invalid snapshot reference as a regular string value.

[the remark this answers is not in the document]
jimmyca15: > The idea is to not throw while parsing snapshot reference.

Is that what's done for key vault references?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
+                throw new InvalidOperationException($"{nameof(snapshot.SnapshotComposition)} for the selected snapshot with name '{snapshot.Name}' must be 'key', found '{snapshot.SnapshotComposition}'.");
+            }
+
+            IAsyncEnumerable<ConfigurationSetting> settingsEnumerable = snapshotReference.Client.GetConfigurationSettingsForSnapshotAsync(
+                snapshotReference.SnapshotName,
+                snapshotReference.CancellationToken);
+
+            await CallWithRequestTracing(async () =>
+            {
+                await foreach (ConfigurationSetting setting in settingsEnumerable.ConfigureAwait(false))
```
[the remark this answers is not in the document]
t-vvidyasaga: Now a cancellationToken is passed into settingsEnumerable

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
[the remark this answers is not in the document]
avanigupta: Ok, that makes sense. Also, to support your point, we break the application for invalid feature flags too. So we should be consistent and throw for invalid snapshot references.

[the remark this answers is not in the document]
avanigupta: Updated suggestion: 
Rename the method here to `Parse` since the class already has snapshot reference in the name. And use `setting.Key` and `setting.Label` in error messages to indicate which snapshot reference caused the format exception

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: Since we're not adding `TryParseSnapshotName`, this gets updated to:

```
if (setting.ContentType == SnapshotReferenceConstants.ContentType)
{
    await LoadSnapshotData(SnapshotReferenceParser.Parse(setting), data, client, cancellationToken).ConfigureAwait(false);

    continue;
}
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
+        }
 
-            return data;
+        private async Task<Dictionary<string, ConfigurationSetting>> Resolve(SnapshotReference snapshotReference)
+        {
+            var resolvedSettings = new Dictionary<string, ConfigurationSetting>();
+
+            if (snapshotReference.SnapshotName == null)
+            {
+                return resolvedSettings;
```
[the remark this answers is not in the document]
t-vvidyasaga: Updated code to throw an exception if snapshot name is null. Below is exception being thrown 
` if (snapshotReference.SnapshotName == null)
  {
       throw new InvalidOperationException("Snapshot reference contains a null snapshot name. The snapshot reference configuration setting must contain a valid 'snapshot_name' property.");
  }`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (setting == null || string.IsNullOrWhiteSpace(setting.Value))
+            {
+                return null;
+            }
+
+            try
+            {
+                Utf8JsonReader reader = new Utf8JsonReader(System.Text.Encoding.UTF8.GetBytes(setting.Value));
+
+                // Ensure the JSON begins with an object '{'
```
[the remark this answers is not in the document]
t-vvidyasaga: Removed comment

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
```diff
@@ -0,0 +1,76 @@
...
+                if (reader.Read() && reader.TokenType != JsonTokenType.StartObject)
+                {
+                    throw new FormatException($"Invalid snapshot reference format. Expected JSON object but found {reader.TokenType}.");
```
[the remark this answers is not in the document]
t-vvidyasaga: Modified code by defining the error messages as constants.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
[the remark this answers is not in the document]
t-vvidyasaga: Updated the method name to Parse and added the` setting.Key` and `setting.Label` to the error messages.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhiyuanliang-ms: Should we update [here](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L867) to make it consistent?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
t-vvidyasaga: Updated the `LoadKeyValuesRegisteredForRefresh ` method to handle the case of if Snapshot Reference was registered then resolve and add the snapshot's configuration settings. But only the snapshot reference will be added to `watchedIndividualKvs`.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReference.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.SnapshotReferences
+{
+    internal class SnapshotReference
```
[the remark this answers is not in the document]
avanigupta: Discussed offline. We are adding this class to maintain future extensibility if we decide to add other snapshot reference properties. cc @jimmyca15

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
jimmyca15, line 21:
```suggestion
        public static SnapshotReference ParseSnapshotReference(ConfigurationSetting setting)
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ContentTypeExtensions.cs
jimmyca15: > Flag to mark if an application has any snapshot reference (eg. tag it with UsesSnapshotRef) 

This is what I would opt to track.

t-vvidyasaga: Added request tracing to keep track of `UsesSnapshotRef`

t-vvidyasaga:
@microsoft-github-policy-service agree company="Microsoft"

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +913,61 @@ await CallWithRequestTracing(async () =>
...
-                    {
-                        throw new InvalidOperationException($"{nameof(snapshot.SnapshotComposition)} for the selected snapshot with name '{snapshot.Name}' must be 'key', found '{snapshot.SnapshotComposition}'.");
-                    }
+            return data;
+        }
 
-                    IAsyncEnumerable<ConfigurationSetting> settingsEnumerable = client.GetConfigurationSettingsForSnapshotAsync(
-                        loadOption.SnapshotName,
-                        cancellationToken);
+        private async Task<Dictionary<string, ConfigurationSetting>> Resolve(SnapshotReference snapshotReference,
```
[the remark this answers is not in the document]
t-vvidyasaga: Have updated the name of the method to `LoadSnapshotData` and also changed the argument to take string snapshotName instead of the SnapshotReference.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -1700,6 +1723,104 @@ await _configClient.SetConfigurationSettingAsync(
...
+            await SetUpSnapshotReferences(testContext);
+
+            await _configClient.SetConfigurationSettingAsync(
+                new ConfigurationSetting($"{testContext.KeyPrefix}:Setting1", "UpdatedAfterSnapshot"));
+
+            var config = new ConfigurationBuilder()
+                .AddAzureAppConfiguration(options =>
+                {
+                    options.Connect(_connectionString);
+                    options.Select($"{testContext.KeyPrefix}:*");
```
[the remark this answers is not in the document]
t-vvidyasaga: Great point to bring up! I updated the test case so `Select` is called twice to explicitly check is snapshot references are resolved correctly.

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -1700,6 +1723,104 @@ await _configClient.SetConfigurationSettingAsync(
...
+            await _configClient.ArchiveSnapshotAsync(testContext.SnapshotName);
+        }
+
+        [Fact]
+        public async Task SnapshotReference_HandleNonExistentSnapshot()
+        {
+            TestContext testContext = CreateTestContext("SnapshotRefNonExistent");
+            await SetupKeyValues(testContext);
+
+            string nonExistentSnapshotName = $"snapshot-does-not-exist-{Guid.NewGuid()}";
```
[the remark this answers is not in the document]
t-vvidyasaga: Thanks for the feedback! I've updated the test to use `testContext.SnapshotName` instead of `Guid.NewGuid()` for better consistency with our test patterns while still ensuring uniqueness.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ContentTypeExtensions.cs
jimmyca15: > Did you verify this works as expected?

?

jimmyca15: Or are you saying you will be removing this function?

tests/Tests.AzureAppConfiguration/Integration/IntegrationTests.cs
```diff
@@ -1700,6 +1723,104 @@ await _configClient.SetConfigurationSettingAsync(
...
+        [Fact]
+        public async Task SnapshotReference_WithKeyVaultReference()
+        {
+            TestContext testContext = CreateTestContext("SnapshotRefKeyVault");
+            await SetupKeyValues(testContext);
+            await SetupKeyVaultReferences(testContext);
+
+            await CreateSnapshot(testContext.SnapshotName, new List<ConfigurationSettingsFilter>
+            {
+                new ConfigurationSettingsFilter(testContext.KeyPrefix + "*")
```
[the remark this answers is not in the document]
t-vvidyasaga: You're absolutely right! The Key Vault reference with `KeyVaultReferenceLabel` wasn't being included in the snapshot. I've fixed it by explicitly adding the Key Vault reference key to the snapshot creation and removing the manual selection in `ConfigurationBuilder()`.

tests/Tests.AzureAppConfiguration/Unit/SnapshotReferencesTests.cs
```diff
@@ -0,0 +1,538 @@
...
+                contentType: SnapshotReferenceConstants.ContentType);
+
+        // Reference with invalid JSON
+        ConfigurationSetting _snapshotReferenceInvalidJson = ConfigurationModelFactory.ConfigurationSetting(
+                key: "SnapshotRefInvalidJson",
+                value: "{invalid json, missing quotes}",
+                eTag: new ETag("snapshot-ref-etag-999"),
+                contentType: SnapshotReferenceConstants.ContentType);
+
+        ConfigurationSetting _snapshotReferenceInvalidJson2 = ConfigurationModelFactory.ConfigurationSetting(
```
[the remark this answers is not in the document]
t-vvidyasaga: Thank you for catching! I have added this case to testing now.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs
```diff
@@ -27,6 +27,7 @@ internal class RequestTracingConstants
         public const string DevEnvironmentValue = "Dev";
         public const string KeyVaultConfiguredTag = "UsesKeyVault";
         public const string KeyVaultRefreshConfiguredTag = "RefreshesKeyVault";
+        public const string SnapshotReferenceTag = "UsesSnapshotRef";
```
jimmyca15, line 30:
```suggestion
        public const string SnapshotReferenceTag = "SnapshotRef";
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ErrorMessages.cs
```diff
@@ -10,5 +10,10 @@ internal class ErrorMessages
         public const string FeatureFlagInvalidJsonProperty = "Invalid property '{0}' for feature flag. Key: '{1}'. Found type: '{2}'. Expected type: '{3}'.";
         public const string FeatureFlagInvalidFormat = "Invalid json format for feature flag. Key: '{0}'.";
         public const string InvalidKeyVaultReference = "Invalid Key Vault reference.";
+        public const string SnapshotReferenceInvalidFormat = "Invalid snapshot reference format for key '{0}' (label: '{1}'). Expected JSON object but found {2}.";
+        public const string SnapshotReferenceInvalidJsonProperty = "Invalid snapshot reference format for key '{0}' (label: '{1}'). The 'snapshot_name' property must be a string value, but found {2}.";
+        public const string SnapshotReferenceNull = "Invalid snapshot reference format. The 'snapshot_name' property must not be null.";
+        public const string SnapshotReferenceInvalidJson = "Invalid snapshot reference format for key '{0}' (label: '{1}'). The value is not valid JSON.";
```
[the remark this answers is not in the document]
t-vvidyasaga: Yes, I just fixed this in the code to remove the redundant error message.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,39 +921,69 @@ await CallWithRequestTracing(async () =>
...
-                    catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.NotFound)
-                    {
-                        throw new InvalidOperationException($"Could not find snapshot with name '{loadOption.SnapshotName}'.", rfe);
-                    }
-
-                    if (snapshot.SnapshotComposition != SnapshotComposition.Key)
-                    {
-                        throw new InvalidOperationException($"{nameof(snapshot.SnapshotComposition)} for the selected snapshot with name '{snapshot.Name}' must be 'key', found '{snapshot.SnapshotComposition}'.");
-                    }
+                    var snapshotReference = new SnapshotReference { SnapshotName = loadOption.SnapshotName };
```
jimmyca15, line 927: Creating a reference here is unnecessary.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,39 +921,69 @@ await CallWithRequestTracing(async () =>
...
             }
 
             return data;
         }
 
+        private async Task<Dictionary<string, ConfigurationSetting>> LoadSnapshotData(string snapshotName, ConfigurationClient client, CancellationToken cancellationToken)
+        {
+            var resolvedSettings = new Dictionary<string, ConfigurationSetting>();
+
+            if (snapshotName == null)
```
jimmyca15, line 945: Check where the snapshot reference is first obtained. Here we can Debug.Assert if desired.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,39 +921,69 @@ await CallWithRequestTracing(async () =>
                 }
                 else
                 {
-                    ConfigurationSnapshot snapshot;
+                    // Load snapshot data
+                    // SnapshotReference snapshotReference = new SnapshotReference(loadOption.SnapshotName);
```
jimmyca15, line 925: Comment should be removed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ContentTypeExtensions.cs
```diff
@@ -6,6 +6,7 @@
 using System.Net.Mime;
```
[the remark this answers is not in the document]
t-vvidyasaga: removed unnecessary code additions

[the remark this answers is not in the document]
jimmyca15: I still see the additions?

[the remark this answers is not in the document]
jimmyca15: image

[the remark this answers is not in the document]
t-vvidyasaga: I just pushed those changes, please let me know if you still see it?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceConstants.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.SnapshotReferences
+{
+    internal class SnapshotReferenceConstants
+    {
+        public const string ContentType = "application/json; profile=\"https://azconfig.io/mime-profiles/snapshot-ref\"; charset=utf-8";
+
+        public const string SnapshotReferenceJsonPropertyName = "snapshot_name";
```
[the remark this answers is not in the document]
t-vvidyasaga: Would you suggest creating this type as a new file to be added and also consolidate Json properties into this file. For example, Key Vault references uri should also be added to this file, or just make this differentiation for snapshot references?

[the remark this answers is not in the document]
jimmyca15: Yes, every class in it's own file.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,39 +921,69 @@ await CallWithRequestTracing(async () =>
...
+        private async Task<Dictionary<string, ConfigurationSetting>> LoadSnapshotData(string snapshotName, ConfigurationClient client, CancellationToken cancellationToken)
+        {
+            var resolvedSettings = new Dictionary<string, ConfigurationSetting>();
+
+            if (snapshotName == null)
+            {
+                throw new InvalidOperationException(ErrorMessages.SnapshotReferenceNull);
+            }
+
+            if (snapshotName == string.Empty)
```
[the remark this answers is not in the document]
t-vvidyasaga: Yes, this would result in throwing (from a bad request) if this check is not in place

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -899,37 +914,114 @@ await CallWithRequestTracing(async () =>
...
+                    if (reader.GetString() == SnapshotReferenceConstants.SnapshotReferenceJsonPropertyName)
                     {
-                        await foreach (ConfigurationSetting setting in settingsEnumerable.ConfigureAwait(false))
+                        if (reader.Read() && reader.TokenType == JsonTokenType.String)
                         {
-                            data[setting.Key] = setting;
+                            string snapshotName = reader.GetString();
+                            if (!string.IsNullOrEmpty(snapshotName))
+                            {
+                                return snapshotName;
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I mean make the code more similar. But making a common helper is definitly better.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReferenceParser.cs
```diff
@@ -0,0 +1,75 @@
...
+        public static SnapshotReference Parse(ConfigurationSetting setting)
+        {
+            if (setting == null)
+            {
+                throw new ArgumentNullException(nameof(setting));
+            }
+
+            if (string.IsNullOrWhiteSpace(setting.Value))
+            {
+                return null;
```
avanigupta, line 30: We need to make sure that `SnapshotReference.SnapshotName` is never null or empty before `LoadSnapshotData` is invoked. There are two ways to make sure that empty or invalid reference values are caught and surfaced as FormatException. 
1. This `Parse` method always throws FormatException if the return value is null or empty.
2. The callers in provider class always check the result of `Parse` for null/empty result. If it is null/empty, provider throws `FormatException`.

Either way, `FormatException` needs to be thrown when we see snapshot ref content type but cannot parse a valid (non-empty) value from it.
I think we should opt for the first strategy because that confines the parsing/validation logic in this parser class instead of having to do it wherever `Parse` is called in provider class.

Thoughts @jimmyca15?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReferences/SnapshotReference.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.SnapshotReferences
```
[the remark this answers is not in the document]
t-vvidyasaga: Fixed the error by using your suggestion of updating `SnapshotReference` with `SnapshotReference.SnapshotReference` in the `AzureAppConfigurationProvider.cs`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ContentTypeExtensions.cs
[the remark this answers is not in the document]
t-vvidyasaga: Hi Jimmy, thank you for telling. I actually did not know what best practice was, I will push my changes before responding to any comments from now. Also, on that note there should be no more whitespace issues with this file now.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -968,8 +1015,31 @@ private async Task<Dictionary<KeyValueIdentifier, ConfigurationSetting>> LoadKey
...
-                    existingSettings[watchedKey] = watchedKv;
+                    if (watchedKv.ContentType == SnapshotReferenceConstants.ContentType)
+                    {
+                        // Track snapshot reference usage for telemetry
+                        if (_requestTracingEnabled && _requestTracingOptions != null)
+                        {
+                            _requestTracingOptions.UpdateSnapshotReferenceTracing(watchedKv.ContentType);
+                        }
+
+                        var snapshotReference = SnapshotReferenceParser.Parse(watchedKv);
```
avanigupta, line 1026: Explicit type instead of var

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -968,8 +1007,31 @@ private async Task<Dictionary<KeyValueIdentifier, ConfigurationSetting>> LoadKey
...
+
+                        Dictionary<string, ConfigurationSetting> resolvedSettings = await LoadSnapshotData(snapshotReference.SnapshotName, client, cancellationToken).ConfigureAwait(false);
+
+                        foreach (KeyValuePair<string, ConfigurationSetting> resolvedSetting in resolvedSettings)
+                        {
+                            existingSettings[resolvedSetting.Key] = resolvedSetting.Value;
+                        }
+
+                        // Only track the snapshot reference itself for refresh monitoring, since snapshots are immutable
+                        watchedIndividualKvs[watchedKeyLabel] = new ConfigurationSetting(watchedKv.Key, watchedKv.Value, watchedKv.Label, watchedKv.ETag);
```
avanigupta, line 1028: Can you take out this line and put it outside the if/else block since its common for both blocks?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ErrorMessages.cs
```diff
@@ -10,5 +12,9 @@ internal class ErrorMessages
         public const string FeatureFlagInvalidJsonProperty = "Invalid property '{0}' for feature flag. Key: '{1}'. Found type: '{2}'. Expected type: '{3}'.";
         public const string FeatureFlagInvalidFormat = "Invalid json format for feature flag. Key: '{0}'.";
         public const string InvalidKeyVaultReference = "Invalid Key Vault reference.";
+        public const string SnapshotReferenceInvalidFormat = "Invalid snapshot reference format for key '{0}' (label: '{1}').";
+        public const string SnapshotReferenceInvalidJsonProperty = "Invalid snapshot reference format for key '{0}' (label: '{1}'). The '" + JsonFields.SnapshotName + "' property must be a string value, but found {2}.";
```
avanigupta, line 16: Can you pass the value of `JsonFields.SnapshotName` to the error message in SnapshotReferenceParser instead of adding this dependency in ErrorMessages.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
t-vvidyasaga: Moved this line outside of the if/else block

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReference/SnapshotReferenceParser.cs
```diff
@@ -0,0 +1,77 @@
...
+                    if (reader.TokenType != JsonTokenType.PropertyName)
+                    {
+                        continue;
+                    }
+
+                    if (reader.GetString() == JsonFields.SnapshotName)
+                    {
+                        if (reader.Read() && reader.TokenType == JsonTokenType.String)
+                        {
+                            if (string.IsNullOrWhiteSpace(reader.GetString()))
```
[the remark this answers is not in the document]
t-vvidyasaga: made this change!

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -968,8 +1011,29 @@ private async Task<Dictionary<KeyValueIdentifier, ConfigurationSetting>> LoadKey
...
+                    {
+                        // Track snapshot reference usage for telemetry
+                        if (_requestTracingEnabled && _requestTracingOptions != null)
+                        {
+                            _requestTracingOptions.UsesSnapshotReference = true;
+                        }
+
+                        SnapshotReference.SnapshotReference snapshotReference = SnapshotReferenceParser.Parse(watchedKv);
+
+                        Dictionary<string, ConfigurationSetting> resolvedSettings = await LoadSnapshotData(snapshotReference.SnapshotName, client, cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
avanigupta: You are setting it correctly in `LoadSelected` method. But I dont see `UsesSnapshotReference` being set to false in `LoadKeyValuesRegisteredForRefresh` method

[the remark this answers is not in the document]
t-vvidyasaga: Oh I'm so sorry, I thought you meant is LoadSelected(). I just made this change!

avanigupta (dismissed):

avanigupta (approved):

Issue #699, opened 2025-09-18
Health check does not respect `AZURE_APP_CONFIGURATION_PROVIDER_DISABLED` and returns `Unhealthy`


julealgon:
I just added the newly introduced health check from version 8.3.0 in one of our projects and immediately went to test the results, only to get an `Unhealthy` response locally in a project that explicitly configures `AZURE_APP_CONFIGURATION_PROVIDER_DISABLED` for local development.

On the console, this is logged:
```
fail: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
      Health check Microsoft.Extensions.Configuration.AzureAppConfiguration with status Unhealthy completed after 1.0959ms with message 'No configuration provider is found.'
```

I think this behavior is not good. Clearly, the health check implementation is ignoring the fact that the project has explicitly opted to turn off the provider, which should be a standard scenario for local development when using the `AZURE_APP_CONFIGURATION_PROVIDER_DISABLED` flag.

I think the team should reconsider the behavior here and return `Healthy` instead if the provider is not present but it has been explicitly turned off.

julealgon:
That seems totally reasonable @zhiyuanliang-ms . I'd definitely be ok with it.

julealgon:
Thank you for looking into this @zhiyuanliang-ms . The reasoning for adding a null-pattern provider is interesting and makes sense. I had not thought about it from that perspective (we don't enumerate the checks ourselves but I can see that happening for sure).

zhiyuanliang-ms:
Done #700

PR #700, opened 2025-09-25
Add AlwaysHealthy health check when provider is disabled

zhiyuanliang-ms:
## Why this PR?

#699

### DIFF
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AlwaysHealthyHealthCheck.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AlwaysHealthyHealthCheck.cs
@@ -1,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal class AlwaysHealthyHealthCheck : IHealthCheck
+    {
+        public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+        {
+            return Task.FromResult(HealthCheckResult.Healthy());
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AlwaysHealthyHealthCheck.cs
+++ b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AlwaysHealthyHealthCheck.cs
@@ -7,11 +7,13 @@ using System.Threading;
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
-    internal class AlwaysHealthyHealthCheck : IHealthCheck
+    internal sealed class AlwaysHealthyHealthCheck : IHealthCheck
     {
+        private static readonly Task<HealthCheckResult> _healthyResult = Task.FromResult(HealthCheckResult.Healthy());
+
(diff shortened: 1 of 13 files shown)


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AlwaysHealthyHealthCheck.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using System.Threading.Tasks;
+using System.Threading;
```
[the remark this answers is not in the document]
zhiyuanliang-ms: We didn't use implicit usings in the entire project before. So I won't use it here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Utils.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+
+using System;
+using System.Security;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal static class Utils
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Change class name to `ProviderToggleChecker`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/EnvironmentVariables.cs
```diff
@@ -14,12 +14,18 @@ internal static class EnvironmentVariables
         /// When set to "true", schema compatibility checks for feature flags are disabled,
         /// and all feature flags will be interpreted using the Microsoft Feature Flags schema.
         /// </summary>
-        public const string DisableFmSchemaCompatibility = "AZURE_APP_CONFIGURATION_FM_SCHEMA_COMPATIBILITY_DISABLED";
+        public const string DisableFmSchemaCompatibilityKey = "AZURE_APP_CONFIGURATION_FM_SCHEMA_COMPATIBILITY_DISABLED";
```
jimmyca15, line 17: Change the class name to EnvironmentVariableNames and then `Key` isn't needed at the end of all the fields.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ProviderToggleChecker.cs
```diff
@@ -0,0 +1,23 @@
...
+//
+
+using System;
+using System.Security;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal static class ProviderToggleChecker
+    {
+        public static bool IsProviderDisabled()
```
jimmyca15, line 12: This class seems more complex than just inlining. The only value seems to be the safe read of an environment variable. If anything, add an extension to try to read an environment variable, and keep it clear in the caller that we rely on environment variable rather than adding another indirection.

jimmyca15: Then that helper can be used elsewhere that we read environment variables as well.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/EnvironmentVariables.cs
jimmyca15: But if we avoid any changes ([comment](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/700#discussion_r2388728958)), then I see no need for the change

PR #706, opened 2025-10-15
Support Azure Front Door

zhiyuanliang-ms:
## Why this PR?

Introduce a new API: `ConnectAzureFrontDoor` which takes the Azure Front Door endpoint as parameter. The configuration provider will load configuration settings from the Azure Front Door.

## Behavior change

When `ConnectAzureFrontDoor` is used:
1. replica discovery and load balancing features will be disabled.
2. watch request (for refresh) will not be conditional request.
3. `authorization` header will be removed from the request
4. `sync-token` will not be attached to the request.
5.  Registering watched settings(sentinel key) for refresh is forbidden when connecting to Azure Front Door. With the consistency model of CDN, Sentinel key based refresh may not meet expectation of customers. Only allow collection monitoring based refresh for CDN scenario.

## Refresh logic

- Use "`x-ms-date` response header from AppConfig - this is the time when server generated the response 

- The provider will now track the page etags and the `x-ms-date` of each page response.

- A refresh is triggered only when the refresh engine detects that the page’s ETag has changed and the accompanying x-ms-date is later than the previous server response time. Once a refresh occurs, the system simply reloads whatever content is currently available from the CDN. This means it may temporarily load outdated pages, but because the provider serving the outdated content will still show a detectable change during subsequent refresh checks, the system will eventually update all pages. The underlying philosophy is that a refresh should, at minimum, achieve what an application restart would. Therefore, our refresh mechanism effectively reloads all configuration settings.

### DIFF
--- a/tests/Tests.AzureAppConfiguration/Unit/FeatureManagementTests.cs
+++ b/tests/Tests.AzureAppConfiguration/Unit/FeatureManagementTests.cs
@@ -1072,7 +1072,7 @@ namespace Tests.AzureAppConfiguration
                 options.UseFeatureFlags();
             }).Build();
 
-            bool performedDefaultQuery = mockTransport.Requests.Any(r => r.Uri.PathAndQuery.Contains("/kv?key=%2A&label=%00"));
+            bool performedDefaultQuery = mockTransport.Requests.Any(r => r.Uri.PathAndQuery.Contains("key=%2A&label=%00"));
             bool queriedFeatureFlags = mockTransport.Requests.Any(r => r.Uri.PathAndQuery.Contains(Uri.EscapeDataString(FeatureManagementConstants.FeatureFlagMarker)));
 
             Assert.True(performedDefaultQuery);
@@ -1100,7 +1100,7 @@ namespace Tests.AzureAppConfiguration
                 })
                 .Build();
 
-            bool performedDefaultQuery = mockTransport.Requests.Any(r => r.Uri.PathAndQuery.Contains("/kv?key=%2A&label=%00"));
+            bool performedDefaultQuery = mockTransport.Requests.Any(r => r.Uri.PathAndQuery.Contains("key=%2A&label=%00"));
             bool queriedFeatureFlags = mockTransport.Requests.Any(r => r.Uri.PathAndQuery.Contains(Uri.EscapeDataString(FeatureManagementConstants.FeatureFlagMarker)));
 
             Assert.True(performedDefaultQuery);


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -980,27 +1074,69 @@ private async Task<Dictionary<KeyValueIdentifier, ConfigurationSetting>> LoadKey
...
-                    await CallWithRequestTracing(async () => watchedKv = await client.GetConfigurationSettingAsync(watchedKey, watchedLabel, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
+                    Response<ConfigurationSetting> response = null;
+                    await CallWithRequestTracing(async () =>
+                    {
+                        response = await client.GetConfigurationSettingAsync(watchedKey, watchedLabel, cancellationToken).ConfigureAwait(false);
+                    }).ConfigureAwait(false);
+
+                    using Response rawResponse = response.GetRawResponse();
+                    DateTimeOffset responseDate = rawResponse.GetDate();
+                    if (_watchedIndividualKvChangeDetectedTime.TryGetValue(watchedKeyLabel, out DateTimeOffset lastChangeDetectedTime))
```
jimmyca15: Date comparison shouldn't be necessary for loading. Only for detecting if a refresh/reload is needed.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ResponseExtensions.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Azure;
+using System;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+    internal static class ResponseExtensions
+    {
+        public static DateTimeOffset GetDate(this Response response)
+        {
+            return response.Headers.Date ?? DateTimeOffset.UtcNow;
```
jimmyca15: I thought we were using x-ms-date?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -416,12 +433,40 @@ public AzureAppConfigurationOptions Connect(IEnumerable<Uri> endpoints, TokenCre
...
+            }
+
+            if (endpoint == null)
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+
+            Credential ??= new EmptyTokenCredential();
+
+            Endpoints = new List<Uri>() { endpoint };
```
jimmyca15: Better to use an array since you only need IEnumerable.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
```diff
@@ -211,6 +211,11 @@ private static string CreateCorrelationContextHeader(RequestType requestType, Re
                 correlationContextTags.Add(RequestTracingConstants.PushRefreshTag);
             }
 
+            if (requestTracingOptions.IsAfdUsed)
+            {
+                correlationContextTags.Add(RequestTracingConstants.AfdTag);
```
jimmyca15: This seems like it should be part of `RequestTracingOptions.CreateFeaturesString`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -35,8 +35,8 @@ internal class AzureAppConfigurationProvider : ConfigurationProvider, IConfigura
         private Dictionary<string, ConfigurationSetting> _mappedData;
         private Dictionary<KeyValueIdentifier, ConfigurationSetting> _watchedIndividualKvs = new Dictionary<KeyValueIdentifier, ConfigurationSetting>();
         private HashSet<string> _ffKeys = new HashSet<string>();
-        private Dictionary<KeyValueSelector, IEnumerable<MatchConditions>> _kvEtags = new Dictionary<KeyValueSelector, IEnumerable<MatchConditions>>();
-        private Dictionary<KeyValueSelector, IEnumerable<MatchConditions>> _ffEtags = new Dictionary<KeyValueSelector, IEnumerable<MatchConditions>>();
+        private Dictionary<KeyValueSelector, IEnumerable<PageWatcher>> _kvPageWatchers = new Dictionary<KeyValueSelector, IEnumerable<PageWatcher>>();
```
jimmyca15:
```suggestion
        private Dictionary<KeyValueSelector, IEnumerable<WatchedPage>> _watchedKvPages_ = new Dictionary<KeyValueSelector, IEnumerable<WatchedPage>>();
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1427,33 +1447,36 @@ private void UpdateClientBackoffStatus(Uri endpoint, bool successful)
...
         {
-            bool haveCollectionsChanged = false;
+            bool havePageChanged = false;
 
             foreach (KeyValueSelector selector in selectors)
             {
-                if (pageEtags.TryGetValue(selector, out IEnumerable<MatchConditions> matchConditions))
+                Debug.Assert(pageWatchers.ContainsKey(selector));
+
+                if (pageWatchers.TryGetValue(selector, out IEnumerable<PageWatcher> watchers))
```
jimmyca15: This sequence doesn't make sense. If there is a guarantee that pageWatchers contains selector, then we shouldn't use TryGetValue afterward.

If there isn't a guarantee of that, then we shouldn't be doing a Debug.Assert.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1427,33 +1447,36 @@ private void UpdateClientBackoffStatus(Uri endpoint, bool successful)
             _configClientBackoffs[endpoint] = clientBackoffStatus;
         }
 
-        private async Task<bool> HaveCollectionsChanged(
+        private async Task<bool> HavePageChange(
```
jimmyca15: I don't understand the motivation for the rename, I find the previous name clearer.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
```diff
@@ -89,23 +91,29 @@ public static async Task<bool> HaveCollectionsChanged(this ConfigurationClient c
 
             AsyncPageable<ConfigurationSetting> pageable = client.GetConfigurationSettingsAsync(selector, cancellationToken);
 
-            using IEnumerator<MatchConditions> existingMatchConditionsEnumerator = matchConditions.GetEnumerator();
+            using IEnumerator<PageWatcher> existingPageWatcherEnumerator = pageWatchers.GetEnumerator();
 
-            await foreach (Page<ConfigurationSetting> page in pageable.AsPages(pageIterator, matchConditions).ConfigureAwait(false))
+            IAsyncEnumerable<Page<ConfigurationSetting>> pages = makeConditionalRequest
+                ? pageable.AsPages(pageIterator, pageWatchers.Select(p => p.Etag))
```
jimmyca15: Is there some limitation preventing us from using the efficient if-not-modified collection scanning approach with AFD? I don't remember discussins such a limitation.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -1000,8 +1010,9 @@ private async Task<Dictionary<KeyValueIdentifier, ConfigurationSetting>> LoadKey
                 }
 
                 // If the key-value was found, store it for updating the settings
-                if (watchedKv != null)
+                if (watchedKv != null && existingSettings != null)
```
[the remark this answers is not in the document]
jimmyca15: Also considering we dereference existingSettings on line 993 above, this check seems questionable or we have a null-ref waiting to happen.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ResponseExtensions.cs
zhiyuanliang-ms: image

Yes, response.Headers.Date will use the `x-ms-date` header

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
zhiyuanliang-ms: Could you check @avanigupta's [comment](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/706#discussion_r2478971223)? I am ok with both. But once we decide, it will be consistent across all libraries.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhiyuanliang-ms: Reverted

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
zhiyuanliang-ms: Discussed with @avanigupta 

AFD does not send conditional requests to origin. So even if we send conditional requests, AFD will send a non-conditional request to AppConfig and cache the 200 response from server. Then AFD decides whether to respond with 304 not modified or 200 modified based on what's present in AFD cache for future requests. 
So the application has no benefit of using conditional requests and it makes the implementation more complicated because now the change decision depends on the state of AFD cache.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs
```diff
@@ -64,11 +66,11 @@ public static async Task<KeyValueChange> GetKeyValueChange(this ConfigurationCli
             };
         }
 
-        public static async Task<bool> HaveCollectionsChanged(this ConfigurationClient client, KeyValueSelector keyValueSelector, IEnumerable<MatchConditions> matchConditions, IConfigurationSettingPageIterator pageIterator, CancellationToken cancellationToken)
+        public static async Task<bool> HaveCollectionsChanged(this ConfigurationClient client, KeyValueSelector keyValueSelector, IEnumerable<WatchedPage> pageWatchers, IConfigurationSettingPageIterator pageIterator, bool makeConditionalRequest, CancellationToken cancellationToken)
```
jimmyca15:
```suggestion
        public static async Task<bool> HaveCollectionsChanged(
            this ConfigurationClient client,
            KeyValueSelector keyValueSelector,
            IEnumerable<WatchedPage> watchedPages,
            IConfigurationSettingPageIterator pageIterator,
            bool makeConditionalRequest,
            CancellationToken cancellationToken)
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs
avanigupta: Ok, let's put it with Features. We can update it in JS provider PR too.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationSource.cs
```diff
@@ -34,13 +36,34 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
             {
                 AzureAppConfigurationOptions options = _optionsProvider();
 
+                IAzureClientFactory<ConfigurationClient> clientFactory = options.ClientFactory;
+
+                if (options.IsAfdUsed)
+                {
+                    if (options.LoadBalancingEnabled)
```
avanigupta: Can you set ReplicaDiscoveryEnabled to false - just to be explicit. I know we cant throw an error because the default is true, but these options are passed on to `AzureAppConfigurationProvider` class. Better to set it to false even if we dont use it later.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/ErrorMessages.cs
```diff
@@ -14,5 +14,9 @@ internal class ErrorMessages
         public const string SnapshotReferenceInvalidJsonProperty = "Invalid snapshot reference format for key '{0}' (label: '{1}'). The '{2}' property must be a string value, but found {3}.";
         public const string SnapshotReferencePropertyMissing = "Invalid snapshot reference format for key '{0}' (label: '{1}'). The '{2}' property is required.";
         public const string SnapshotInvalidComposition = "{0} for the selected snapshot with name '{1}' must be 'key', found '{2}'.";
+        public const string ConnectionConflict = "Cannot connect to both Azure App Configuration and Azure Front Door at the same time.";
+        public const string AfdConnectionConflict = "Cannot connect to multiple Azure Front Doors.";
+        public const string AfdLoadBalancingUnsupported = "Load balancing is not supported when connecting to Azure Front Door.";
```
avanigupta: Error message can be updated to match https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/223#discussion_r2500578331

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -854,15 +856,16 @@ private async Task<Dictionary<string, ConfigurationSetting>> LoadSelected(
...
 
                     await CallWithRequestTracing(async () =>
                     {
                         AsyncPageable<ConfigurationSetting> pageableSettings = client.GetConfigurationSettingsAsync(selector, cancellationToken);
 
                         await foreach (Page<ConfigurationSetting> page in pageableSettings.AsPages(_options.ConfigurationSettingPageIterator).ConfigureAwait(false))
                         {
-                            using Response response = page.GetRawResponse();
+                            using Response rawResponse = page.GetRawResponse();
+                            DateTimeOffset responseDate = rawResponse.GetDate();
```
avanigupta:
```suggestion
                            DateTimeOffset serverResponseTime = rawResponse.GetMsDate();
```

jimmyca15 (approved):

avanigupta (approved):

avanigupta (approved):

PR #731, opened 2026-03-19
Nullable reference type

linglingye001:


zhiyuanliang-ms:
Nullable mark should not be abused. [NRT should be used to express the design intent](https://learn.microsoft.com/en-us/dotnet/csharp/tutorials/nullable-reference-type).

All of our C# code bases don't use `string?`, this is an established convention. We should keep it.

What we should do is to add more null check and throw the correct exception (intead of NRE).

src/Microsoft.Azure.AppConfiguration.AspNetCore/AzureAppConfigurationRefreshExtensions.cs
```diff
@@ -25,7 +25,7 @@ public static IApplicationBuilder UseAzureAppConfiguration(this IApplicationBuil
                 throw new ArgumentNullException(nameof(builder));
             }
 
-            IConfigurationRefresherProvider refresherProvider = (IConfigurationRefresherProvider)builder.ApplicationServices.GetService(typeof(IConfigurationRefresherProvider));
+            IConfigurationRefresherProvider? refresherProvider = (IConfigurationRefresherProvider?)builder.ApplicationServices.GetService(typeof(IConfigurationRefresherProvider));
```
zhiyuanliang-ms, line 28: I don't think we need to make `IConfigurationRefresherProvider` nullable here. We have null check below, correct exception will be thrown

src/Microsoft.Azure.AppConfiguration.AspNetCore/Microsoft.Azure.AppConfiguration.AspNetCore.csproj
```diff
@@ -3,6 +3,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
+    <Nullable>enable</Nullable>
```
zhiyuanliang-ms, line 6: None of our C# code bases and even .NET azure SDKs don't set Nullable to enable (the default value is disable). If we want to set it explicitly, it should be disable

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/CachedKeyVaultSecret.cs
```diff
@@ -10,7 +10,7 @@ internal class CachedKeyVaultSecret
         ///// <summary>
         ///// The value of the Key Vault secret.
         ///// </summary>
-        public string SecretValue { get; set; }
+        public string? SecretValue { get; set; }
```
zhiyuanliang-ms, line 13: Adding `?` doesn't correctly express the design intent, the SecretValue should not be null for `CachedKeyValueSecret`

linglingye001:
Discussed offline, not add nullable annotation to keep consistent with the Azure .NET SDK ecosystem

Issue #735, opened 2026-05-19
Support for Parallel Secret Resolution


janmarien:
As already available for the JavaScript SDK, the option for parallel secret resolution is highly desired.
Currently App configuration resolution times are running up to several minutes which could be drastically reduced if Key Vault secrets would be resolved in parallel instead of sequentially.

linglingye001:
Hi @janmarien, thanks for raising this and we're actively working on it. Could you share a bit more about your scenario? How many Key Vault references are being resolved in the configuration that currently takes up to several minutes?

janmarien:
Hi @linglingye001, several minutes might have been a bit exaggerated but we do have App Configs which include over 10+ Key Vault references. Suppose on average all those references take around 200ms to resolve, we easily reach 2 minutes for the whole app configuration resolution (including authentication calls etc..) to have completed. For most jobs, this is still acceptable, but we do have some time-sensitive jobs that would highly benefit from the parallel resolution.

zhiyuanliang-ms:
Hey, @janmarien Thanks for reply!
Did you get the number of key vault reference wrong? With only 10+ secret reference (each takes 200ms), how can whole configuration resolution take 2 minutes? Do you mean 2 seconds? How many key value do you have in total?

linglingye001:
Hi @janmarien, parallel secret resolution is available in [v8.6.0](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration/8.6.0)

janmarien:
> Hey, [@janmarien](https://github.com/janmarien) Thanks for reply! Did you get the number of key vault reference wrong? With only 10+ secret reference (each takes 200ms), how can whole configuration resolution take 2 minutes? Do you mean 2 seconds? How many key value do you have in total?

Hi @zhiyuanliang-ms, apologies for the horrific math there 🤦 Indeed the whole configuration does not take minutes, but does take multiple seconds. In our case, for certain time-sensitive matters, we would really benefit from reducing this loading time by being able to resolve the secrets in parallel.

@linglingye001 Thank you very much for implementing this feature. We will go ahead and start testing with it!

PR #736, opened 2026-05-20
Resolve key vault references concurrently

linglingye001:
This PR introduces an opt-in capability to resolve Azure Key Vault references concurrently during Azure App Configuration load, aiming to reduce startup time when many Key Vault references are present. https://github.com/Azure/AppConfiguration-DotnetProvider/issues/735

**Changes:**
- Add `ParallelSecretResolutionEnabled` option under Key Vault configuration and plumb it into provider options.
- Update configuration loading to optionally process adapter resolution concurrently.
- Add unit tests covering parallel resolution behavior and default sequential behavior


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -16,6 +16,7 @@ internal class AzureKeyVaultSecretProvider
         private readonly AzureAppConfigurationKeyVaultOptions _keyVaultOptions;
         private readonly IDictionary<string, SecretClient> _secretClients;
         private readonly Dictionary<Uri, CachedKeyVaultSecret> _cachedKeyVaultSecrets;
+        private readonly object _cacheLock = new object();
```
zhiyuanliang-ms: Instead of using lock, we should use ConcurrentDictionary for atomic operation

linglingye001: discussed offline

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -81,42 +80,35 @@ public async Task<string> GetSecretValue(KeyVaultSecretIdentifier secretIdentifi
...
+            foreach (KeyValuePair<Uri, CachedKeyVaultSecret> secret in _cachedKeyVaultSecrets)
+            {
+                if (secret.Value.RefreshAt.HasValue && secret.Value.RefreshAt.Value < utcNow)
+                {
+                    return true;
+                }
+            }
+
+            return false;
         }
```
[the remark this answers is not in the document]
linglingye001: `ShouldRefreshKeyVaultSecrets` is called once per `RefreshAsync` invocation (gated by the refresh interval), not per config read, and N is the number of unique KV references — practically tens. The scan reads two fields per entry from a ConcurrentDictionary and runs in microseconds; it's dominated by the network I/O that follows whenever it returns true.

Reintroducing the min-RefreshAt tracker is also not as cheap as it was before. Without `_cacheLock`, every write would need a CAS update on a shared field, and whenever the secret holding the current minimum gets its RefreshAt pushed later (the common case after a successful refresh) or is removed, we still need an O(N) rescan to find the new minimum. So we'd be moving an O(N) scan from the cold check path onto the parallel write path that this PR is specifically trying to keep contention-free.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -625,17 +626,44 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
...
+            if (pendingKeyVaultTasks?.Count > 0)
+            {
+                IEnumerable<KeyValuePair<string, string>>[] keyVaultResults =
+                    await Task.WhenAll(pendingKeyVaultTasks).ConfigureAwait(false);
+
+                results.AddRange(keyVaultResults);
+            }
+
+            foreach (IEnumerable<KeyValuePair<string, string>> keyValuePairs in results)
+            {
```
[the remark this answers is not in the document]
linglingye001: it's not very likely that the key vault reference and key values share the same key after key prefix trimming.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -625,17 +626,44 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
                 _requestTracingOptions.ResetAiConfigurationTracing();
             }
 
+            bool parallelSecretResolution = _options.IsParallelSecretResolutionEnabled;
+
+            // Only Key Vault references perform network I/O during adapter processing; other
+            // adapters complete synchronously. When parallel resolution is enabled, Key Vault
+            // references are dispatched concurrently while non-Key Vault settings are processed
+            // inline. Results are merged once at the end.
+            var results = new List<IEnumerable<KeyValuePair<string, string>>>(data.Count);
```
zhiyuanliang-ms: No need to temporarily store all results. The previous implementation sets applicationData directly is fine.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -656,6 +684,12 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
             return applicationData;
         }
 
+        private static bool IsKeyVaultReference(ConfigurationSetting setting)
+        {
```
zhiyuanliang-ms: The same code also exists in AzureKeyVaultAdapter. We should extract them. Maybe adding ConfigurationSettingExtensions.cs

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -144,6 +144,11 @@ internal IEnumerable<IKeyValueAdapter> Adapters
         /// </summary>
         internal bool IsKeyVaultRefreshConfigured { get; private set; } = false;
 
+        /// <summary>
+        /// Flag to indicate whether Key Vault references should be resolved in parallel.
+        /// </summary>
+        internal bool IsParallelSecretResolutionEnabled { get; private set; } = false;
```
zhiyuanliang-ms: Boolean's default value is false. No need to specify it. Also remove the value for the two booleans above, make them consistent with KeyVaultOptions

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -39,10 +37,10 @@ public async Task<string> GetSecretValue(KeyVaultSecretIdentifier secretIdentifi
         {
             string secretValue = null;
 
-            if (_cachedKeyVaultSecrets.TryGetValue(secretIdentifier.SourceId, out CachedKeyVaultSecret cachedSecret) &&
-                    (!cachedSecret.RefreshAt.HasValue || DateTimeOffset.UtcNow < cachedSecret.RefreshAt.Value))
+            if (_cachedKeyVaultSecrets.TryGetValue(secretIdentifier.SourceId, out CachedKeyVaultSecret cachedHit) &&
```
zhiyuanliang-ms: I think `cachedSecret` is more clear. I don't think we should rename it

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -81,42 +80,35 @@ public async Task<string> GetSecretValue(KeyVaultSecretIdentifier secretIdentifi
 
         public bool ShouldRefreshKeyVaultSecrets()
         {
-            return _nextRefreshTime.HasValue && _nextRefreshTime.Value < DateTimeOffset.UtcNow;
+            DateTimeOffset utcNow = DateTimeOffset.UtcNow;
```
zhiyuanliang-ms: no need to have this variable

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -81,42 +80,35 @@ public async Task<string> GetSecretValue(KeyVaultSecretIdentifier secretIdentifi
...
+
+            return false;
         }
 
         public void ClearCache()
         {
-            var sourceIdsToRemove = new List<Uri>();
-
-            var utcNow = DateTimeOffset.UtcNow;
+            DateTimeOffset utcNow = DateTimeOffset.UtcNow;
```
zhiyuanliang-ms: no need to have this variable. Just use it inline

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -14,16 +14,14 @@ namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault
     internal class AzureKeyVaultSecretProvider
     {
         private readonly AzureAppConfigurationKeyVaultOptions _keyVaultOptions;
-        private readonly IDictionary<string, SecretClient> _secretClients;
-        private readonly Dictionary<Uri, CachedKeyVaultSecret> _cachedKeyVaultSecrets;
-        private Uri _nextRefreshSourceId;
```
zhiyuanliang-ms, line 19: The reason we remove the `_nextRefreshSourceId` and `_nextRefreshTime` is to simplify the implementation. Otherwise, we have to add lock here, because they will be updated in the `GetSecretValue` code path (the `SetSecretInCache` call)

The benefit of maintaining the `_netRefresh` pair is that the `ShouldRefreshKeyVaultSecrets` call will be an O(1) operation. The implementation in this PR makes it an O(n) operation. But I think there won't be too many cached secrets and iterate the whole concurrent dictionary is not that expensive. So I think it is fine.


cc @avanigupta @jimmyca15

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -15,6 +15,7 @@
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Net.Mime;
```
zhiyuanliang-ms: what is this?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -625,17 +626,44 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
...
+                    results.Add(await ProcessAdapters(kvp.Value, cancellationToken).ConfigureAwait(false));
+                }
+            }
 
+            if (pendingKeyVaultTasks?.Count > 0)
+            {
+                IEnumerable<KeyValuePair<string, string>>[] keyVaultResults =
+                    await Task.WhenAll(pendingKeyVaultTasks).ConfigureAwait(false);
+
+                results.AddRange(keyVaultResults);
```
avanigupta: This change breaks config composition when multiple settings share the same key. Key Vault references are now always appended at the end, causing them to override existing key-values and effectively ignore the selection precedence order. We need to avoid this behavior.

avanigupta:
I'm thinking of a different approach for this PR. What if we added a new method to the `IKeyValueAdapter` interface which pre-loads data before `ProcessKeyValue` is called. The pre-loader would be responsible for parallelization of KeyVault calls, and it would update the secret cache so that by the time `ProcessKeyValue` is invoked, it just reads from cache (no I/O in ProcessKeyValue). This would preserve the precedence order for config composition. Other adapters dont have anything to preload, so they are no-op.

`Task PreloadAsync(IEnumerable<ConfigurationSetting> settings, Logger logger, CancellationToken cancellationToken);`

Semantics: "Given all the settings about to be processed, pre-warm your caches so that `ProcessKeyValue` can return instantly."

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: `System.Net.Mime` is the BCL namespace that contains `ContentType`,  we also have this in ContentTypeExtensions.cs

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
zhiyuanliang-ms: Are loading configuration settings selector by selector? My understanding is that under one selector, there will not be duplicated key name, so the override behavior only happens across selectors.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
```diff
@@ -31,6 +31,13 @@ public class AzureAppConfigurationKeyVaultOptions
         internal TimeSpan? DefaultSecretRefreshInterval = null;
         internal bool IsKeyVaultRefreshConfigured = false;
 
+        /// <summary>
+        /// Specifies whether Key Vault references should be resolved in parallel.
+        /// Default value is false. Enabling this can reduce the time required to resolve Key Vault references
```
zhiyuanliang-ms:
```suggestion
        /// Enabling this can reduce the time required to resolve Key Vault references
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
avanigupta: That is correct. We set the order during load and need to maintain the same precedence order during adapter processing.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationKeyVaultOptions.cs
avanigupta: Default value is good to specify in public facing docs

avanigupta: It can be worded differently though. How about this:

"Flag to indicate whether Key Vault references should be resolved in parallel. Disabled by default."

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IKeyValueAdapter.cs
```diff
@@ -13,6 +13,10 @@ internal interface IKeyValueAdapter
     {
         Task<IEnumerable<KeyValuePair<string, string>>> ProcessKeyValue(ConfigurationSetting setting, Uri endpoint, Logger logger, CancellationToken cancellationToken);
 
+        // Pre-warm any per-setting state (e.g. Key Vault secret cache) before ProcessKeyValue is invoked
+        // on each setting. Adapters with no pre-fetchable state can return a completed task.
```
avanigupta: comment not needed on interface to keep it consistent with other methods

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -625,16 +625,25 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
...
 
+            // When parallel secret resolution is enabled, let each adapter pre-warm its caches
+            // (Key Vault references are dispatched concurrently here) so the sequential loop below
+            // can process settings in the original order without losing precedence on key collisions.
+            if (_options.IsParallelSecretResolutionEnabled)
+            {
+                await Task.WhenAll(
+                    _options.Adapters.Select(adapter =>
+                        adapter.PreloadAsync(data.Values, _logger, cancellationToken)))
+                    .ConfigureAwait(false);
```
avanigupta:
```suggestion
                foreach (IKeyValueAdapter adapter in _options.Adapters)
                {
                    await adapter.PreloadAsync(data.Values, _logger, cancellationToken).ConfigureAwait(false);
                }
```

avanigupta: Parallelization applies only to secret resolution

linglingye001: Updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationSettingExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+//
+using Azure.Data.AppConfiguration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.AzureKeyVault;
+using System.Net.Mime;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+    internal static class ConfigurationSettingExtensions
+    {
+        public static bool IsKeyVaultReference(this ConfigurationSetting setting)
```
avanigupta: New extension adds more redirection without much benefit since we already have ContentTypeExtensions. We can reuse `contentType.IsKeyVaultReference()` wherever we need to check for key vault reference.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +112,79 @@ public bool NeedsRefresh()
             return _secretProvider.ShouldRefreshKeyVaultSecrets();
         }
 
+        public async Task PreloadAsync(IEnumerable<ConfigurationSetting> settings, Logger logger, CancellationToken cancellationToken)
+        {
+            if (settings == null || !_secretProvider.IsParallelSecretResolutionEnabled)
```
avanigupta: Preloading is not limited to parallel operations. We can always pre-warm the cache regardless of parallelization, to make `ProcessKeyValue` always read from cache only.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -137,12 +137,17 @@ internal IEnumerable<IKeyValueAdapter> Adapters
...
         /// <summary>
         /// Flag to indicate whether Key Vault secret values will be refreshed automatically.
         /// </summary>
-        internal bool IsKeyVaultRefreshConfigured { get; private set; } = false;
+        internal bool IsKeyVaultRefreshConfigured { get; private set; }
+
+        /// <summary>
+        /// Flag to indicate whether Key Vault references should be resolved in parallel.
+        /// </summary>
+        internal bool IsParallelSecretResolutionEnabled { get; private set; }
```
avanigupta: Are we using this anywhere?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
linglingye001: Yes, `ProcessKeyValue` read from cache. The problem now is that how we pre-warm the cache. I think we still need a boolean to control the parallelization during the pre-warm process. Or are you saying that we resolve the KVR in parallel by default?

avanigupta: Yes, we still need the boolean to decide whether to load in parallel or in sequence. But the load operation should still happen inside `PreloadAsync` regardless of the boolean. Dont return early if `!_secretProvider.IsParallelSecretResolutionEnabled`. Let the flag decide if you load using `Task.WhenAll` parallelly, or load using `await` for each secret sequentially. Then ProcessKeyValue is a pure cache reader in both modes.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +112,79 @@ public bool NeedsRefresh()
...
+
+            var tasks = new Task[toFetch.Count];
+
+            for (int i = 0; i < toFetch.Count; i++)
+            {
+                (KeyVaultSecretIdentifier identifier, string key, string label) = toFetch[i];
+                tasks[i] = PreloadSecretAsync(identifier, key, label, logger, cancellationToken);
+            }
+
+            await Task.WhenAll(tasks).ConfigureAwait(false);
```
avanigupta: We need to define a max concurrency limit.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +112,79 @@ public bool NeedsRefresh()
...
+                    continue;
+                }
+
+                string secretRefUri = ParseSecretReferenceUri(setting);
+
+                if (string.IsNullOrEmpty(secretRefUri) ||
+                    !Uri.TryCreate(secretRefUri, UriKind.Absolute, out Uri secretUri) ||
+                    !KeyVaultSecretIdentifier.TryCreate(secretUri, out KeyVaultSecretIdentifier secretIdentifier))
+                {
+                    // Invalid references are surfaced from ProcessKeyValue with full exception context.
```
avanigupta: How do we propagate secret resolution errors when resolving in parallel? If one secret fails to resolve. do we end early or finish all tasks before throwing the error?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
linglingye001: No, removed

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
linglingye001: how about 16? do you have any suggestion?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
linglingye001: Currently the parallel preload always waits for every task to finish before throw the error. And  `ProcessKeyValue` will re-fetch the first failing reference (cache was evicted), hits the same error, and throws a properly populated `KeyVaultReferenceException` with the Key/Label/Etag/SecretIdentifier info. 

If fail fast and throw in parallel preload, it would be generic exception without key/label context. 

Do we need to end early in parallel phase？

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
linglingye001: Got it. Updated the `PreloadAsync`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
linglingye001: looks like there's no recommended number from the doc https://learn.microsoft.com/en-us/azure/key-vault/general/overview-throttling#how-to-throttle-your-app-in-response-to-service-limits

avanigupta: KeyVault has a limit of 4000 requests per 10 seconds per vault: https://learn.microsoft.com/en-us/azure/key-vault/general/service-limits#secrets-managed-storage-account-keys-and-vault-transactions

What if we made the concurrency configurable by user?
Instead of exposing a boolean option to turn on/off parallel secret resolution, we could offer something like `int MaxSecretResolutionConcurrency`. The default value will be 1 (i.e., sequential load from KeyVault). Anything higher than 1 will enable parallel secret resolution. 
Users can configure it to whatever max concurrency they want based on their Key Vault usage.

cc @jimmyca15

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -75,9 +75,7 @@ KeyVaultReferenceException CreateKeyVaultReferenceException(string message, Conf
 
         public bool CanProcess(ConfigurationSetting setting)
         {
-            if (setting == null ||
-                string.IsNullOrWhiteSpace(setting.Value) ||
-                string.IsNullOrWhiteSpace(setting.ContentType))
+            if (setting == null || string.IsNullOrWhiteSpace(setting.Value))
```
avanigupta: This change can be reverted

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
avanigupta: With the I/O operations moving to `PreloadAsync`, `ProcessKeyValue` should not need to make a request to KeyVault just to get the error again. 
We can avoid losing the key/label context during parallel execution by having each task wrap its own failure into a `KeyVaultReferenceException` internally (where its Key/Label are still in scope), so the exception surfaced by `Task.WhenAll` is already fully populated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
zhiyuanliang-ms: This topic has been discussed for JS provider. See [comment](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/192#discussion_r2079026707)

I think there are several things we should align:
1. For server side scenario, it will be very difficult to hit the key vault maximum transactions. If there are 100 secrets in a store, there must be more than 40 instances start/refresh within 10 seconds.
2. For the most critical configuration initial load path, we have startup retry on. So even we are throttled, it will be retried.
3. The [customer](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/735#issuecomment-4955807981) said he only had 10+ secret references.

I think instead of managing the maximum concurrency by ourselves, we should take advantage of key vault sdk's built-in retry mechanism. For those uses who use secret reference very heavily, they should pass secret client instance directly with retry policy configured in secret client options.

zhiyuanliang-ms: Discussed offline with @jimmyca15 

We aligned on that the purpose of limit the max concurrency is not to avoid key vault throttling, but to be responsible for memory usage.

If there are thousands of secret references and we create a task for each one, memory usage could explode.

Therefore, instead of exposing a configurable `MaxSecretResolutionConcurrency` option, we just need to enforce an internal concurrency limit to ensure that parallel secret resolution uses a manageable amount of memory.

avanigupta: Btw, there is precedent for other Azure services exposing the concurrency limit in parallel resolution scenarios like these. 
For eg: 
- [StorageTransferOptions.MaximumConcurrency](https://learn.microsoft.com/en-us/dotnet/api/azure.storage.storagetransferoptions.maximumconcurrency?view=azure-dotnet) 
- [ServiceBusProcessorOptions.MaxConcurrentCalls](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusprocessoroptions.maxconcurrentcalls?view=azure-dotnet)
- [LogsUploadOptions.MaxConcurrency](https://learn.microsoft.com/en-us/dotnet/api/azure.monitor.ingestion.logsuploadoptions.maxconcurrency?view=azure-dotnet)

We can always define our upper limit for this config. But if we already exposed boolean option in JS provider, I'm good with keeping boolean option here too. If we have to keep it internal, 16 sounds like a good limit to me.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +116,98 @@ public bool NeedsRefresh()
...
+        {
+            if (semaphore != null)
+            {
+                await semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+            }
+
+            try
+            {
+                await _secretProvider.GetSecretValue(identifier, setting.Key, setting.Label, logger, cancellationToken).ConfigureAwait(false);
+            }
```
zhiyuanliang-ms: The current implementation will still create a `Task` instance for each secret references. We can have a fixed size task array and use worker pattern to resolve secret in parallel.

```C#
int workerCount = Math.Min(KeyVaultConstants.MaxParallelSecretResolution, toFetch.Count);

int nextIndex = -1;
Task[] workers = new Task[workerCount];

for (int i = 0; i < workerCount; i++)
{
    workers[i] = ((Func<Task>)(async () =>
    {
        while (true)
        {
            int index = Interlocked.Increment(ref nextIndex);

            if (index >= toFetch.Count)
            {
                return;
            }

            var secretToFetch = toFetch[index];

            await load(secretToFetch);
        }
    }))
```

linglingye001: The worker pool would be the better if there're thousands of references per load. While in most scenarios KV refs < 100,

- The workload (N = number of KV references) is small; the allocation savings of the worker pool is not obvious.
- The semaphore version is more readable and each item maps cleanly to one task, which makes the `KeyVaultReferenceException` propagation (already carefully wired to carry `Key`/`Label`/`SecretIdentifier`) straightforward.
- The worker pool needs extra hand-written machinery (exception capture so a dying worker doesn't silently drop items), more surface for subtle bugs.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +88,103 @@ public bool NeedsRefresh()
...
         }
 
+        public async Task PreloadAsync(IEnumerable<ConfigurationSetting> settings, Logger logger, CancellationToken cancellationToken)
+        {
+            if (settings == null)
+            {
+                return;
+            }
+
+            HashSet<Uri> seen = new HashSet<Uri>();
```
zhiyuanliang-ms:
```suggestion
            var seen = new HashSet<Uri>();
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +88,103 @@ public bool NeedsRefresh()
...
+            List<(KeyVaultSecretIdentifier Identifier, ConfigurationSetting Setting, string SecretRefUri)> toFetch = new List<(KeyVaultSecretIdentifier, ConfigurationSetting, string)>();
```
zhiyuanliang-ms:
```suggestion
            var toFetch = new List<(KeyVaultSecretIdentifier, ConfigurationSetting, string)>();
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +88,103 @@ public bool NeedsRefresh()
...
+
+            foreach (ConfigurationSetting setting in settings)
+            {
+                if (!CanProcess(setting))
+                {
+                    continue;
```
zhiyuanliang-ms, line 104: In JS, we throw if `Preload` takes settings that the current adapater cannot process. I think it is ok to just skip the settings

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
zhiyuanliang-ms: The reason we need ConfigurationSetting and SecretRefUri here is to build error message. In my option, these two information can be optional when building key vault reference exception.


The hash set and the list are kind of duplicated. In theory, we only need one copy of all unique identifiers.

linglingye001: `KeyVaultSecretIdentifier Identifier, ConfigurationSetting Setting, string SecretRefUri` are used when call `secretProvider.GetSecretValue`, besides build error message, they're also used in logger

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +88,103 @@ public bool NeedsRefresh()
...
+                            try
+                            {
+                                await _secretProvider.GetSecretValue(identifier, setting, secretRefUri, logger, failFastCts.Token).ConfigureAwait(false);
+                            }
+                            catch (Exception e)
+                            {
+                                // Stop other workers from picking up new work and remember the first failure so it can be surfaced below.
+                                Interlocked.CompareExchange(ref firstException, e, null);
+                                failFastCts.Cancel();
+                                return;
```
zhiyuanliang-ms: I think we can just throw here.

```C#
    failFastCts.Cancel();
    throw;
```

In this case, we can save the firstException.

linglingye001: If every worker does catch `{ failFastCts.Cancel(); throw; }`, multiple worker tasks can end up faulted simultaneously: the one that hit the real error, plus any siblings whose in-flight `GetSecretValue` call observes `failFastCts.Token` cancellation and throws `OperationCanceledException`.

The problem is which exception await `Task.WhenAll(workers)` surfaces isn't based on which happened first chronologically — `Task.WhenAll`'s aggregated exceptions are ordered by the position of the task in the workers array, not fault order. So if `workers[0]` is a sibling that got cancelled as a side effect, and `workers[5]` is the one that hit the actual KeyVaultReferenceException, `await Task.WhenAll` would throw workers[0]'s `OperationCanceledException` — masking the real root cause with a meaningless cancellation exception.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -49,74 +54,76 @@ public async Task<string> GetSecretValue(KeyVaultSecretIdentifier secretIdentifi
...
                 else if (_keyVaultOptions.SecretResolver != null)
                 {
                     secretValue = await _keyVaultOptions.SecretResolver(secretIdentifier.SourceId).ConfigureAwait(false);
                 }
 
-                cachedSecret = new CachedKeyVaultSecret(secretValue, secretIdentifier.SourceId);
+                updatedCachedSecret = new CachedKeyVaultSecret(secretValue, secretIdentifier.SourceId);
                 success = true;
             }
+            catch (OperationCanceledException)
```
zhiyuanliang-ms: Why catch this error?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -49,74 +54,76 @@ public async Task<string> GetSecretValue(KeyVaultSecretIdentifier secretIdentifi
 
             if (client == null && _keyVaultOptions.SecretResolver == null)
             {
-                throw new UnauthorizedAccessException("No key vault credential or secret resolver callback configured, and no matching secret client could be found.");
+                throw KeyVaultReferenceException.Create("No key vault credential or secret resolver callback configured, and no matching secret client could be found.", setting, null, secretRefUri);
```
zhiyuanliang-ms: I don't think we should change the exception type here. This is a general exception unrelated to specific setting

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
linglingye001: If the cancellation happens while `Azure.Identity` code is acquiring a token (e.g. `client.GetSecretAsync` gets cancelled mid-auth), the resulting `OperationCanceledException`'s Source can legitimately be `Azure.Identity`— which matches the filter and would get it wrongly wrapped into a `KeyVaultReferenceException` instead of propagating as a plain cancellation.

That distinction matters directly for the new `PreloadAsync` fail-fast logic: when one worker fails and calls `failFastCts.Cancel()`, sibling workers' in-flight `GetSecretValue` calls should throw a clean `OperationCanceledException` that gets discarded (only the real root-cause exception should win the `Interlocked.CompareExchange` race). If cancellation got misclassified as a `KeyVaultReferenceException` here, it could incorrectly win that race and mask the actual error that triggered the fail-fast — the exact bug we discussed avoiding.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
linglingye001: Before this error will be catched in adapter and re-throw with KeyVaultReferenceException, since now we move that part error handing into GetSecret, I think throw with ` KeyVaultReferenceException` here is fine.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
linglingye001: Discussed offline, use `semaphoreSlim` pattern for code readability and simplicity.

zhiyuanliang-ms (dismissed):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +87,83 @@ public bool NeedsRefresh()
...
+                {
+                    throw KeyVaultReferenceException.Create("Invalid Key vault secret identifier.", setting, null, secretRefUri);
+                }
+
+                if (!seen.Add(secretIdentifier.SourceId))
+                {
+                    continue;
+                }
+
+                toFetch.Add((secretIdentifier, setting, secretRefUri));
```
avanigupta:
```suggestion
                if (seen.Add(secretIdentifier.SourceId))
                {
                    toFetch.Add((secretIdentifier, setting, secretRefUri));
                }
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
```diff
@@ -35,7 +40,7 @@ public AzureKeyVaultSecretProvider(AzureAppConfigurationKeyVaultOptions keyVault
             }
         }
 
-        public async Task<string> GetSecretValue(KeyVaultSecretIdentifier secretIdentifier, string key, string label, Logger logger, CancellationToken cancellationToken)
+        public async Task<string> GetSecretValue(KeyVaultSecretIdentifier secretIdentifier, ConfigurationSetting setting, string secretRefUri, Logger logger, CancellationToken cancellationToken)
```
avanigupta: I dont think we need `secretRefUri` as an argument for `GetSecretValue`. That information is already present in `secretIdentifier.SourceId`, and we only use it for the exception messages anyway.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +87,83 @@ public bool NeedsRefresh()
...
 
+        public async Task PreloadAsync(IEnumerable<ConfigurationSetting> settings, Logger logger, CancellationToken cancellationToken)
+        {
+            if (settings == null)
+            {
+                return;
+            }
+
+            var seen = new HashSet<Uri>();
+            var toFetch = new List<(KeyVaultSecretIdentifier, ConfigurationSetting, string)>();
```
avanigupta:
```suggestion
            var toFetch = new List<(KeyVaultSecretIdentifier, ConfigurationSetting)>();
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs
```diff
@@ -116,6 +87,83 @@ public bool NeedsRefresh()
...
+            if (toFetch.Count == 0)
+            {
+                return;
+            }
+
+            if (_secretProvider.IsParallelSecretResolutionEnabled)
+            {
+                using (var semaphore = new SemaphoreSlim(KeyVaultConstants.MaxParallelSecretResolution))
+                {
+                    async Task ResolveSecretAsync(KeyVaultSecretIdentifier identifier, ConfigurationSetting setting, string secretRefUri)
```
avanigupta: `secretRefUri` argument can be dropped from all new methods.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultSecretProvider.cs
linglingye001: Make sense, dropped `secretRefUri`

avanigupta (approved):

PR #738, opened 2026-06-12
Loading feature flags from new endpoint

linglingye001:
Adds support for loading feature flags from Azure App Configuration's new feature-flag endpoint , rather than only reading "classic" flags stored as key-values prefixed with `.appconfig.featureflag/`.

**Changes:**

- For each `FeatureFlagSelector`, fetches flags via the dedicated SDK endpoint, with change detection/refresh support.
- Default service version bumped to `V2026_05_01_Preview`.
- Introduces `ToConfiguration(...)` to convert a `FeatureFlag` into configuration key-values. Removed adapter pattern for feature flags.
- Index handling — classic flags occupy array indices `0..N-1`; new standalone flags continue from the classic flag count (`feature_management:feature_flags:{index}`). 
- New feature flag supersedes classic flag with the same name; superseded classic flags are dropped from mapped data.


src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -323,14 +323,46 @@ public AzureAppConfigurationOptions UseFeatureFlags(Action<FeatureFlagOptions> c
...
+
+                var newFfSelector = new KeyValueSelector
+                {
+                    KeyFilter = nameFilter,
+                    LabelFilter = featureFlagSelector.LabelFilter,
+                    TagFilters = featureFlagSelector.TagFilters,
+                    IsNewFeatureFlagSelector = true
+                };
+
+                _selectors.AppendUnique(newFfSelector);
```
jimmyca15: I think we should separate key value selectors and feature flag selectors.

And I don't believe we need to duplicate the feature flag selectors and watchers.

Whenever we are processing feature flag selectors we know we must

1. take the name filter, append the classic feature flag prefix, query classic feature flags
2. Take the name filter, use it to query standalone (new) feature flags.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -930,6 +946,60 @@ await CallWithRequestTracing(async () =>
...
+            {
+                AsyncPageable<SdkFeatureFlag> pageable = client.GetFeatureFlagsAsync(selector, cancellationToken);
+
+                await foreach (Page<SdkFeatureFlag> page in pageable.AsPages().ConfigureAwait(false))
+                {
+                    using Response response = page.GetRawResponse();
+
+                    foreach (SdkFeatureFlag ff in page.Values)
+                    {
+                        ConfigurationSetting synthesized = FeatureFlagSettingConverter.ToConfigurationSetting(ff);
```
jimmyca15: The new feature flags shouldn't be converted to configuration setting. When we load a new feature flag, we should break it down into it's associated key-value pairs, like the feature flag adapter does for classic feature flags, and inject that into the configuration directly.

jimmyca15:
This needs to go into the preview branch.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -937,6 +936,130 @@ await CallWithRequestTracing(async () =>
...
+            // (later writes to `data` overwrite earlier ones).
+            var classicPagesBySelector = new Dictionary<FeatureFlagSelector, IEnumerable<WatchedPage>>();
+
+            if (!_options.ExcludeClassicFeatureFlags)
+            {
+                foreach (FeatureFlagSelector ffSelector in featureFlagSelectors)
+                {
+                    var selector = new SettingSelector()
+                    {
+                        KeyFilter = FeatureManagementConstants.FeatureFlagMarker + (ffSelector.NameFilter ?? "*"),
```
jimmyca15: This fall back to "*" is interesting. How would we end up with a feature flag selector with a null name filter. would that be

```
ffOptions.Select(null);
```

I would've thought that we would throw if someone passes null for the name filter.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: I saw some updates in the PR. Is this one still being worked on?

linglingye001: Yes, working in progess

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: Right, provider will throw if `null` is passed. The fall back is not needed, updated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
linglingye001: Added internal type `FeatureFlagSelector` , `Select()` stores the raw name filter (no .appconfig.featureflag/ prefix).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: Updated, new feature flags now map to the internal model (provider FeatureFlag) and share the ff adapter's logic.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs
```diff
@@ -605,5 +625,30 @@ private static ConfigurationClientOptions GetDefaultClientOptions()
 
             return clientOptions;
         }
+
+        /// <summary>
+        /// Builds <see cref="FeatureFlagClientOptions"/> that mirror the relevant settings of the
+        /// configured <see cref="ClientOptions"/> so that the feature-flag client communicates with the
+        /// same store, audience, transport and retry behavior as the configuration client.
+        /// </summary>
+        internal FeatureFlagClientOptions GetFeatureFlagClientOptions()
```
jimmyca15: If we're going to separate the options, I think it will be better to have FeatureFlagClientOptions as a property just like ClientOptions.

And whenever ConfigureClientOptions is used, we can reflect the changes on FeatureFlagClientOptions.

jimmyca15: Makes consumption equal.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ClientWrapper.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Azure.Data.AppConfiguration;
+using System;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal class ClientWrapper
```
jimmyca15: How about we introduce `IAppConfigurationClient` which exposes the methods that we need to use from both FeatureFlagClient and ConfigurationClient. And the implementation `AppConfigurationClient` can contain both internally.

jimmyca15: That way the provider can still work with a single client.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -640,24 +669,56 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
...
-                    string key = kv.Key;
+                    AddWithTrimmedPrefix(applicationData, kv.Key, kv.Value);
+                }
+            }
 
-                    foreach (string prefix in _options.KeyPrefixes)
+            // Emit configuration for feature flags loaded from the standalone feature-flag endpoint.
+            // These are mapped directly from the SDK model to feature-management key-values without
+            // round-tripping through a synthesized ConfigurationSetting.
+            if (sdkFFs != null && sdkFFs.Count > 0)
```
jimmyca15: I think that there should be a separate method to add feature flags into data.

```
private void ProcessFeatureFlags(
    Dictionary<string, ConfigurationSetting> data,
    IEnumerable<FeatureFlag> featureFlags);
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -640,24 +669,56 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
...
+                    .FirstOrDefault();
+
+                if (featureFlagAdapter != null)
+                {
+                    foreach (KeyValuePair<string, SdkFeatureFlag> kvp in sdkFFs)
                     {
-                        if (key.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+                        SdkFeatureFlag featureFlag = kvp.Value;
+
+                        FeatureManagement.FeatureFlag model = SdkFeatureFlagModelMapper.MapToModel(featureFlag);
```
jimmyca15: There shouldn't be an intermediate state. We should go straight from `FeatureFlag` to the `IEnumerable<KeyValuePair<string, string>>` that should go into configuration.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/FeatureFlagClientExtensions.cs
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Azure;
+using Azure.Data.AppConfiguration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration.Models;
+using SdkFeatureFlag = Azure.Data.AppConfiguration.FeatureFlag;
```
jimmyca15: Let's avoid introducing this concept of 'SdkFeatureFlag'.

Going forward 'FeatureFlag' refers to the new flags.

All previous naming for old flags should be renamed to ClassicFeatureFlag. Including the type [here](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/preview/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlag.cs).

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -937,6 +998,138 @@ await CallWithRequestTracing(async () =>
             return data;
         }
 
+        private async Task LoadFeatureFlags(
```
jimmyca15: I'd like to keep classic feature flags and standalone feature flags separate.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/FeatureFlagClientExtensions.cs
jimmyca15: If we do need some separation, we can use the terms classic feature flag and standalone feature flag.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AppConfigurationClient.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// </summary>
+    internal class AppConfigurationClient : IAppConfigurationClient
+    {
+        private readonly ConfigurationClient _configurationClient;
+        private readonly FeatureFlagClient _featureFlagClient;
+
+        public AppConfigurationClient(Uri endpoint, ConfigurationClient configurationClient)
+        {
+            Endpoint = endpoint;
+            _configurationClient = configurationClient;
```
jimmyca15: Check for null.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AppConfigurationClient.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        public AppConfigurationClient(Uri endpoint, ConfigurationClient configurationClient, FeatureFlagClient featureFlagClient)
+        {
+            Endpoint = endpoint;
+            _configurationClient = configurationClient;
+            _featureFlagClient = featureFlagClient;
```
jimmyca15: I believe we should always have a feature flag client. Even if the caller of the provider doesn't use feature flags. Otherwise, this class is in a partial state that callers have to be aware they can't call feature flag methods.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -673,36 +668,38 @@ private async Task<Dictionary<string, string>> PrepareData(
...
+        {
+            if (featureFlags == null)
+            {
+                return;
+            }
 
-                        ETag etag = featureFlag.Etag ?? default;
+            FeatureManagementKeyValueAdapter featureFlagAdapter = _options.Adapters
+                .OfType<FeatureManagementKeyValueAdapter>()
+                .FirstOrDefault();
```
jimmyca15: We shouldn't have the code looking like we may not have what we need to convert a feature flag. Here's what I think we can do to make it more explicit.

Add a static type `static class FeatureFlagConverter` that has a public method

`public IEnumerable<KeyValuePair<string, string>> ToConfiguration(FeatureFlag flag)`.

Adapters are for processing configuration settings. The new feature flags aren't configuration settings so we don't need an adapter.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AppConfigurationClient.cs
linglingye001: Removed two-arg constructor.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: .NET config binds `feature_management:feature_flags` as an array,  classic flags get indices `0..N-1`, then standalone continue `N..N+M-1`. For FeatureFlagConverter, we need `_featureFlagIndex`, `_featureFlagTracing` and `Endpoint` for FeatureFlagReference. 

How about 
```
public static IEnumerable<KeyValuePair<string, string>> ToConfiguration(
            FeatureFlag flag,
            ref int featureFlagIndex,
            Uri endpoint,
            FeatureFlagTracing tracing)
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagConverter.cs
```diff
@@ -0,0 +1,497 @@
...
+        /// <param name="featureFlagIndex">
+        /// The next available index in the "feature_management:feature_flags" array. It is advanced when the
+        /// flag is emitted using the Microsoft schema, so that a caller converting multiple flags produces a
+        /// contiguous set of indices.
+        /// </param>
+        /// <param name="endpoint">The endpoint used to build the feature flag reference for telemetry.</param>
+        /// <param name="tracing">Tracing used to record feature filter, variant, seed and telemetry usage.</param>
+        public static IEnumerable<KeyValuePair<string, string>> ToConfiguration(
+            FeatureFlag flag,
+            ref int featureFlagIndex,
```
jimmyca15: The caller should manage the index. Ref introduces complexities of how to handle mutations when exceptions occur.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagConverter.cs
```diff
@@ -0,0 +1,497 @@
...
+            Uri endpoint,
+            FeatureFlagTracing tracing)
```
jimmyca15: Tracing updating should be separated from conversion to configuration.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -640,22 +664,58 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
...
+            {
+                return;
+            }
+
+            // Continue the "feature_management:feature_flags" array index after any classic feature flags
+            // already emitted (by the feature-management adapter) in this load, so that the combined array
+            // uses a single contiguous set of indices.
+            int featureFlagIndex = _options.Adapters
+                .OfType<FeatureManagementKeyValueAdapter>()
+                .FirstOrDefault()?.FeatureFlagIndex ?? 0;
```
jimmyca15: What about on refresh, does the index keep incrementing? It seems like we should track index only in the provider. I think we need to move away from the adapter pattern for feature flags.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -937,7 +993,176 @@ await CallWithRequestTracing(async () =>
...
+
+            Dictionary<FeatureFlagSelector, IEnumerable<WatchedPage>> newPagesBySelector =
+                await LoadFeatureFlags(client, featureFlagSelectors, featureFlags, ffKeys, cancellationToken).ConfigureAwait(false);
+
+            foreach (FeatureFlagSelector ffSelector in featureFlagSelectors)
+            {
+                classicPagesBySelector.TryGetValue(ffSelector, out IEnumerable<WatchedPage> classicPages);
+                newPagesBySelector.TryGetValue(ffSelector, out IEnumerable<WatchedPage> newPages);
+
+                ffPageWatchers[ffSelector] = new FeatureFlagWatchedPages
```
jimmyca15: Let's track _classicFeatureFlagWatchedPages and _featureFlagWatchedPages separately so we don't have to introduce a new type.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: Discussed offline. Use converter for the classic and new ffs. When process classic ff, there're Dotnet ff schema and Microsoft ff schema, so the ff index is not equal to the classicFf.count() and invisible to the caller after the conversion.  Working in progress for this part.

jimmyca15: > so the ff index is not equal to the classicFf.count()

But it's fine to start at classicFfCount right? The only issue would be if we started at an index that was already occupied. No issue if we skip an index.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -627,35 +680,103 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
...
+            if (featureFlags == null)
+            {
+                return processedFeatureFlags;
+            }
+
+            foreach (FeatureFlag featureFlag in featureFlags)
+            {
+                _options.FeatureFlagTracing.Update(featureFlag);
+
+                string featureFlagPath = $"{FeatureManagementConstants.FeatureManagementSectionName}:{FeatureManagementConstants.FeatureFlagsSectionName}:{featureFlagIndex}";
```
jimmyca15: Why isn't this part of FeatureFlagConverter.ToConfiguration?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -627,35 +680,103 @@ private async Task<Dictionary<string, string>> PrepareData(Dictionary<string, Co
...
-                        {
-                            key = key.Substring(prefix.Length);
-                            break;
-                        }
-                    }
+            // The number of classic feature flags emitted using the Microsoft schema. Standalone feature flags
+            // continue the "feature_management:feature_flags" array from this index so that the combined array
+            // uses a single contiguous set of indices. Classic flags emitted using the .NET schema do not
+            // occupy an array slot and therefore do not advance this count.
+            _classicFeatureFlagCount = featureFlagIndex;
```
jimmyca15: I thought we would be able to use the classic feature flag count and simplify things. The worse thing I could think was an index might be skipped which isn't an issue.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -937,7 +1063,199 @@ await CallWithRequestTracing(async () =>
             return data;
         }
 
-        private async Task<Dictionary<string, ConfigurationSetting>> LoadSnapshotData(string snapshotName, ConfigurationClient client, CancellationToken cancellationToken)
+        class ClassicFeatureFlagLoadResult
```
jimmyca15: Class definitions should be private and brought to the top.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -937,7 +1063,199 @@ await CallWithRequestTracing(async () =>
...
+        }
+
+        // Merges the classic feature flags loaded from the ".appconfig.featureflag/" key-value namespace into
+        // the key-value data so that they are mapped and emitted alongside the regular key-values. Standalone
+        // feature flags are kept separate and emitted directly (see ProcessFeatureFlags).
+        private static Dictionary<string, ConfigurationSetting> MergeClassicFeatureFlags(
+            Dictionary<string, ConfigurationSetting> data,
+            ClassicFeatureFlagLoadResult classicFeatureFlagLoadResult)
+        {
+            if (classicFeatureFlagLoadResult?.ClassicFeatureFlags != null)
```
jimmyca15: Based off when we call this, is it ever expected to be null? I would expect no.

jimmyca15: It's not clear to me that this is worth a helper method.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -937,7 +1063,199 @@ await CallWithRequestTracing(async () =>
...
+                foreach (FeatureFlag featureFlag in featureFlagLoadResult.FeatureFlags)
+                {
+                    ffKeys.Add(FeatureManagementConstants.FeatureFlagMarker + (featureFlag.Name ?? string.Empty));
+                }
+            }
+
+            return ffKeys;
+        }
+
+        private static FlagSelector BuildFlagSelector(FeatureFlagSelector ffSelector)
```
jimmyca15: This looks like a good candiate for an extension method on FeatureFlagSelector

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -409,12 +428,21 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
                                 _mappedData.Remove(key);
                             }
 
-                            Dictionary<string, ConfigurationSetting> mappedFfData = await MapConfigurationSettings(ffCollectionData).ConfigureAwait(false);
+                            // Drop any classic feature flag in the mapped data that is now superseded by a
```
jimmyca15: Even with the comment I'm having trouble understanding what this is for.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: Oh I see. I thought the index must be contiguous. Yes, we can start at classicFFCount. Will update the code

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: Sorry about that, this loop is redundant. Will remove.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: It should be in ToConfiguration(). Updated.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: Updated, use classic feature flag count now.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: Discussed offline.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: Yes, it won't be null.  
Removed helper method.
```
    ffKeys = new HashSet<string>(
        classicFeatureFlagLoadResult.ClassicFeatureFlags.Select(ff => ff.Key)
        .Concat(featureFlagLoadResult.FeatureFlags.Select(ff => FeatureManagementConstants.FeatureFlagMarker + ff.Name))
    );

```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IClientManager.cs
```diff
@@ -8,14 +8,12 @@
 
 namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
 {
-    internal interface IConfigurationClientManager
+    internal interface IClientManager
```
zhiyuanliang-ms: How about `IAppConfigurationClientManager`

We also use SecretClient

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/FeatureFlagClientExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Extensions
+{
+    internal static class FeatureFlagClientExtensions
+    {
+        public static async Task<bool> HaveFeatureFlagsChanged(
+            this IAppConfigurationClient client,
+            Models.FeatureFlagSelector featureFlagSelector,
```
zhiyuanliang-ms, line 22:
```suggestion
            FeatureFlagSelector featureFlagSelector,
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IFeatureFlagPageIterator.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Azure;
+using Azure.Data.AppConfiguration;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration
+{
+    internal interface IFeatureFlagPageIterator
```
zhiyuanliang-ms, line 10: Can we add some comment to explain the purpose of this interface and `IConfigurationSettingPageIterator` is to facilitate testing.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -453,7 +490,18 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
                         // PrepareData makes calls to KeyVault and may throw exceptions. But, we still update watchers before
                         // SetData because repeating appconfig calls (by not updating watchers) won't help anything for keyvault calls.
                         // As long as adapter.NeedsRefresh is true, we will attempt to update keyvault again the next time RefreshAsync is called.
-                        SetData(await PrepareData(_mappedData, cancellationToken).ConfigureAwait(false));
+                        Dictionary<string, string> preparedData = await PrepareData(_mappedData, _featureFlags, cancellationToken).ConfigureAwait(false);
+
+                        int classicFeatureFlagCount = classicFeatureFlagLoadResult?.ClassicFeatureFlags.Count() ?? 0;
```
jimmyca15: Isn't classicFeatureFlagLoadResult guaranteed to be non-null here?

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -937,7 +1061,133 @@ await CallWithRequestTracing(async () =>
             return data;
         }
 
-        private async Task<Dictionary<string, ConfigurationSetting>> LoadSnapshotData(string snapshotName, ConfigurationClient client, CancellationToken cancellationToken)
+        // Loads classic feature flags (from the ".appconfig.featureflag/" key-value namespace) into `data`
+        // as configuration settings. Returns the watched pages per selector.
+        private async Task<ClassicFeatureFlagLoadResult> LoadClassicFeatureFlags(
```
jimmyca15: It seems this should return an empty result if classic feature flags are ignored.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -810,32 +931,47 @@ await ExecuteWithFailOverPolicyAsync(
...
+            // Merges the classic feature flags loaded from the ".appconfig.featureflag/" key-value namespace into the key-value data.
+            foreach (ConfigurationSetting setting in classicFeatureFlagLoadResult.ClassicFeatureFlags)
+            {
+                data[setting.Key] = setting;
+            }
+
+            Dictionary<string, ConfigurationSetting> mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
 
-                SetData(await PrepareData(mappedData, cancellationToken).ConfigureAwait(false));
+            Dictionary<string, string> preparedData = await PrepareData(mappedData, featureFlagLoadResult.FeatureFlags, cancellationToken).ConfigureAwait(false);
```
jimmyca15: I think we should make a change. Instead of passing in `featureFlagLoadResult.FeatureFlags` to PrepareData to remove the keys that are superseded by standalone feature flags, I think we should filter out these keys before we call 'MapConfigurationSettings`. We can do it at this level. we'll need to do it in a few places. But the idea is that classic feature flags that aren't going to be superseded won't appear in the `MapConfigurationSettings` flow.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -331,44 +349,52 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
...
-                                KeyFilter = watcher.Key,
-                                LabelFilter = watcher.Label,
-                                TagFilters = watcher.Tags,
-                                IsFeatureFlagSelector = true
-                            }),
-                            _watchedFfPages,
-                            client,
-                            cancellationToken).ConfigureAwait(false);
+                        refreshFeatureFlag = await HaveClassicFeatureFlagsChanged(_options.FeatureFlagSelectors, _watchedClassicFeatureFlagPages, appConfigClient, cancellationToken).ConfigureAwait(false)
+                                            || await HaveFeatureFlagsChanged(_options.FeatureFlagSelectors, _watchedFeatureFlagPages, appConfigClient, cancellationToken).ConfigureAwait(false);
```
jimmyca15:
```suggestion
                            || await HaveFeatureFlagsChanged(_options.FeatureFlagSelectors, _watchedFeatureFlagPages, appConfigClient, cancellationToken).ConfigureAwait(false);
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -937,7 +1061,133 @@ await CallWithRequestTracing(async () =>
...
+                    AsyncPageable<FeatureFlag> pageable = client.GetFeatureFlagsAsync(selector, cancellationToken);
+
+                    await foreach (Page<FeatureFlag> page in pageable.AsPages(_options.FeatureFlagPageIterator).ConfigureAwait(false))
+                    {
+                        using Response rawResponse = page.GetRawResponse();
+                        DateTimeOffset serverResponseTime = rawResponse.GetMsDate();
+
+                        foreach (FeatureFlag ff in page.Values)
+                        {
+                            string key = FeatureManagementConstants.FeatureFlagMarker + (ff.Name ?? string.Empty);
```
jimmyca15: This seems unused.

jimmyca15 (changes requested):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: Also, if any classic feature flags get into the system from the configuration selectors, we should exclude them from the system if ExcludeClassicFeatureFlags is enabled.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: it won't be null if `refreshAll` or `refreshFeatureFlag` is `true`

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -379,9 +405,28 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
...
+
                     if (refreshAll)
                     {
+                        // Merges the classic feature flags loaded from the ".appconfig.featureflag/" key-value namespace into the key-value data,
+                        // excluding any that are superseded by a standalone feature flag with the same name.
+                        var featureFlagKeys = new HashSet<string>(
+                            featureFlagLoadResult.FeatureFlags.Select(ff => FeatureManagementConstants.FeatureFlagMarker + ff.Name));
+
+                        List<ConfigurationSetting> classicFeatureFlags = classicFeatureFlagLoadResult.ClassicFeatureFlags
+                            .Where(setting => !featureFlagKeys.Contains(setting.Key))
```
zhiyuanliang-ms: For classic feature flag, configuration settings key is not necessarily the same as the feature flag id.
We should dedup based on feature flag id instead of settings key.

But, in fact, as long as we ensure that new feature flag is added to configuration after classic feature flag:
``` json
[
{ "id": "classic-ff-1"},
....
{ "id": "new-ff-1"},
]
```

The new feature flag will always override the old feature flag because of the last one win behavior in feature management library.

So I think we don't need to dedup here.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -810,32 +949,52 @@ await ExecuteWithFailOverPolicyAsync(
...
-                Dictionary<string, ConfigurationSetting> mappedData = await MapConfigurationSettings(data).ConfigureAwait(false);
+            // Merges the classic feature flags loaded from the ".appconfig.featureflag/" key-value namespace into the key-value data,
+            // excluding any that are superseded by a standalone feature flag with the same name.
+            var featureFlagKeys = new HashSet<string>(
+                featureFlagLoadResult.FeatureFlags.Select(ff => FeatureManagementConstants.FeatureFlagMarker + ff.Name));
 
-                SetData(await PrepareData(mappedData, cancellationToken).ConfigureAwait(false));
+            List<ConfigurationSetting> classicFeatureFlags = classicFeatureFlagLoadResult.ClassicFeatureFlags
+                .Where(setting => !featureFlagKeys.Contains(setting.Key))
+                .ToList();
```
jimmyca15: No need to allocate a list, keeping as an IEnumerable is fine.

linglingye001: updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -331,44 +350,52 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
...
+                                appConfigClient,
+                                _options.FeatureFlagSelectors,
+                                cancellationToken).ConfigureAwait(false);
+
+                            featureFlagLoadResult = await LoadFeatureFlags(
+                                appConfigClient,
+                                _options.FeatureFlagSelectors,
+                                cancellationToken).ConfigureAwait(false);
+
+                            ffKeys = new HashSet<string>(
```
jimmyca15: ff keys should only need to hold classic feature flag keys. It's only used to remove classic feature flags from mapped data.

jimmyca15: Then _featureFlagIndex shouldn't be needed because it can always be obtained from _ffKeys.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -277,47 +298,45 @@ public async Task RefreshAsync(CancellationToken cancellationToken)
                     //
                     // Avoid instance state modification
                     Dictionary<KeyValueSelector, IEnumerable<WatchedPage>> kvEtags = null;
-                    Dictionary<KeyValueSelector, IEnumerable<WatchedPage>> ffEtags = null;
                     HashSet<string> ffKeys = null;
```
jimmyca15: ffKeys can be declared after ExecuteWithFailoverPolicy. Since it doesn't rely on client, there's no need to initialize it inside that block.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
jimmyca15: As a result, _ffKeys can be renamed to _classicFfKeys.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -403,18 +444,33 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
...
                         {
+                            // Exclude any classic feature flags that are superseded by a standalone feature flag with the same name.
+                            var featureFlagKeys = new HashSet<string>(
+                                featureFlagLoadResult.FeatureFlags.Select(ff => FeatureManagementConstants.FeatureFlagMarker + ff.Name));
+
+                            IEnumerable<ConfigurationSetting> classicFeatureFlags = classicFeatureFlagLoadResult.ClassicFeatureFlags
+                                .Where(setting => !featureFlagKeys.Contains(setting.Key));
+
                             // Remove all feature flag keys that are not present in the latest loading of feature flags, but were loaded previously
                             foreach (string key in _ffKeys.Except(ffKeys))
```
jimmyca15: It seems we don't have to be too smart here. We can just remove all `_ffKeys` from `_mappedData` because we add them back right after.

Given we take the change for ff keys to only have classic feature flag keys.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -934,10 +1080,154 @@ await CallWithRequestTracing(async () =>
...
+        private async Task<ClassicFeatureFlagLoadResult> LoadClassicFeatureFlags(
+            IAppConfigurationClient client,
+            IEnumerable<FeatureFlagSelector> featureFlagSelectors,
+            CancellationToken cancellationToken)
+        {
+            var classicFeatureFlags = new Dictionary<string, ConfigurationSetting>();
+
+            var pages = new Dictionary<FeatureFlagSelector, IEnumerable<WatchedPage>>();
+
+            if (_options.ExcludeClassicFeatureFlags)
```
jimmyca15: Keeping a single return statement makes things easier to review. I'd suggest keeping the single return statement at the bottom and putting the central piece in the if statement.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -35,6 +35,14 @@ internal TimeSpan RefreshInterval
         /// </summary>
         public string Label { get; set; }
 
+        /// <summary>
+        /// When set to true, classic feature flags (key-values whose key is prefixed with
+        /// ".appconfig.featureflag/") are not loaded from the configuration store. Only feature flags
+        /// returned by the new feature-flag endpoint will be loaded.
+        /// Defaults to false; classic flags are loaded alongside new flags for backward compatibility.
+        /// </summary>
```
jimmyca15: No need for default value (` = false;`) for booleans.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -381,7 +408,21 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
 
                     if (refreshAll)
                     {
+                        // Merges the classic feature flags loaded from the ".appconfig.featureflag/" key-value namespace into the key-value data,
+                        // excluding any that are superseded by a standalone feature flag with the same name.
+                        var featureFlagKeys = new HashSet<string>(
+                            featureFlagLoadResult.FeatureFlags.Select(ff => FeatureManagementConstants.FeatureFlagMarker + ff.Name));
+
+                        IEnumerable<ConfigurationSetting> classicFeatureFlags = classicFeatureFlagLoadResult.ClassicFeatureFlags
```
jimmyca15:
```suggestion
                        IEnumerable<ConfigurationSetting> eligibleClassicFeatureFlags = classicFeatureFlagLoadResult.ClassicFeatureFlags
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
```diff
@@ -403,18 +444,33 @@ await ExecuteWithFailOverPolicyAsync(clients, async (client) =>
...
+                            }
+
+                            Dictionary<string, ConfigurationSetting> mappedFfData = await MapConfigurationSettings(classicFeatureFlags.ToDictionary(x => x.Key, x => x)).ConfigureAwait(false);
 
                             foreach (KeyValuePair<string, ConfigurationSetting> kvp in mappedFfData)
                             {
                                 _mappedData[kvp.Key] = kvp.Value;
                             }
+
+                            _featureFlags = featureFlagLoadResult.FeatureFlags;
```
jimmyca15:
```suggestion
                            //
                            // Remove all previous classic feature flag keys
                            // We'll will add them back below if they are still in the latest batch
                            foreach (string key in _classicFfKeys)
                            {
                                _mappedData.Remove(key);
                            }

                            //
                            // Exclude any classic feature flags that are superseded by a standalone feature flag with the same name.
                            var ineligibleClassicFfKeys = new HashSet<string>(
                                featureFlagLoadResult.FeatureFlags.Select(ff => FeatureManagementConstants.FeatureFlagMarker + ff.Name));

                            IEnumerable<ConfigurationSetting> eligibleClassicFeatureFlags = classicFeatureFlagLoadResult.ClassicFeatureFlags
                                .Where(setting => !ineligibleClassicFfKeys.Contains(setting.Key));

                            Dictionary<string, ConfigurationSetting> mappedFfData = await MapConfigurationSettings(eligibleClassicFeatureFlags.ToDictionary(x => x.Key, x => x)).ConfigureAwait(false);

                            foreach (KeyValuePair<string, ConfigurationSetting> kvp in mappedFfData)
                            {
                                _mappedData[kvp.Key] = kvp.Value;
                            }

                            _featureFlags = featureFlagLoadResult.FeatureFlags;
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs
linglingye001: Discussed offline. Will keep the dedup handling here which can make all the classic flags disappear from system and provider map callback.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
```diff
@@ -35,6 +35,14 @@ internal TimeSpan RefreshInterval
         /// </summary>
         public string Label { get; set; }
 
+        /// <summary>
+        /// When set to true, classic feature flags (key-values whose key is prefixed with
+        /// ".appconfig.featureflag/") are not loaded from the configuration store. Only feature flags
+        /// returned by the new feature-flag endpoint will be loaded.
+        /// Defaults to false; classic flags are loaded alongside new flags for backward compatibility.
```
jimmyca15:
```suggestion
        /// When set to true, only enhanced feature flags will be loaded.
        /// Defaults to false; All feature flag types are loaded.
```

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureFlagOptions.cs
jimmyca15: After further thought, it may be better to skip this API for now, and we can re-introduce it post-preview. When trying out enhanced feature flags, it is most likely to not have any of the concerns that this API aims to resolve, such as spending requests to query empty sets of classic feature flags.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/ClassicClientFilter.cs
yuanqu72, line 1: We need to rename the files: the official decision is to keep the old feature flag as "feature flag" and new feature flag as "enhanced feature flag". We should not have "classic feature flag" anywhere in our code/doc.

examples/ConsoleAppWithFailOver/Program.cs
yuanqu72, line 1: We should add samples to show how the enhanced feature flag is loaded using selector, i.e. use it to actually filter new ff by name/label/tag. Based on the code change in this pr, it should behave the same as the old ff.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/ClassicClientFilter.cs
linglingye001: updated

examples/ConsoleAppWithFailOver/Program.cs
linglingye001: discussed offline

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationFeatureFlagClientFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+    internal class AzureAppConfigurationFeatureFlagClientFactory : IAzureClientFactory<FeatureFlagClient>
+    {
+        private readonly FeatureFlagClientOptions _clientOptions;
+
+        private readonly TokenCredential _credential;
+        private readonly IEnumerable<string> _connectionStrings;
+
+        public AzureAppConfigurationFeatureFlagClientFactory(
+            IEnumerable<string> connectionStrings,
+            FeatureFlagClientOptions clientOptions)
```
zhiyuanliang-ms, line 22: I am curious about why does [`FeatureFlagClientOptions`](https://github.com/Azure/azure-sdk-for-net/blob/appconfig-ff-endpoint-preview/sdk/appconfiguration/Azure.Data.AppConfiguration/src/FeatureFlags/FeatureFlagClientOptions.cs) support all api version we have.

Should we prevent user from modifying api version for feature flag client?

@jimmyca15 @mrm9084

mrm9084: The feature flag client is just the configuration client with the 'configuration' methods removed. We can remove the values. One thing to keep in mind is that our clients are just clients sending pipeline requests. Technically a `FeatureFlagClient` has the ability to send a json payload to app config asking for whatever, and it uses the authentication setup in the client. So, using a pipeline request the customer could use an old api version to access configurations from the feature flag client.

I don't know why anyone would do this, but they could.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/EnhancedFeatureFlagConverter.cs
```diff
@@ -362,17 +370,10 @@ private static void WriteParameterValue(Utf8JsonWriter writer, string value)
 
             if (trimmed.Length > 0 && (trimmed[0] == '{' || trimmed[0] == '['))
             {
-                try
-                {
-                    using JsonDocument doc = JsonDocument.Parse(value);
-                    doc.RootElement.WriteTo(writer);
+                using JsonDocument doc = JsonDocument.Parse(value);
```
jimmyca15: For feature filter parameters we should preserve the fallback to string. Feature filter parameters don't have the content type property, so parsing as JSON is always a best effort attempt. The customer may actually intend for it to be a string.

This is in contrast to variant value's that set content type to "application/json". In that case, the customer is declaring that they want that value to be treated as json.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/EnhancedFeatureFlagConverter.cs
```diff
@@ -0,0 +1,401 @@
...
+            }
+
+            using var stream = new MemoryStream();
+
+            using (var writer = new Utf8JsonWriter(stream))
+            {
+                writer.WriteStringValue(value);
+            }
+
+            using JsonDocument stringDoc = JsonDocument.Parse(stream.ToArray());
```
jimmyca15: Is it possible to write it like this to avoid ToArray

```
byte[] bytes = JsonSerializer.SerializeToUtf8Bytes(value);
using JsonDocument doc = JsonDocument.Parse(bytes);
return doc.RootElement.Clone();
```

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/EnhancedFeatureFlagConverter.cs
linglingye001: updated, preserve the fallback to string.

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/EnhancedFeatureFlagConverter.cs
linglingye001: updated

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationFeatureFlagClientFactory.cs
zhiyuanliang-ms: To me, the new feature flag change is not something opt-in. If a customer configured a specific api-version, and he upgraded to the latest provider version. That customer has no knowledge about the new feature flag change (e.g. new api version, new endpoint, etc)
Suddenly, the provider will fail at every request sent by provider to get enhanced feature flag.

This case doesn't apply to .NET provider, because we don't have any public api to set client factory for `FaetureFlagClient`. But in JS provider, feature flag client options is [configurable](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/341/changes/BASE..a80420fe3b7427a0ead655a45ab7fc504aabaae2#r3819306399).

zhiyuanliang-ms: In .NET provider, there are actually two ways to configure SDK client:
1. `AzureAppConfigurationOptions.ConfigureClientOptions`

Image
This method will modify the `ConfigurationClientOptions` and then we will sync retry and audience configuration to `FeatureFlagClientOptions` only (api version is not included)

2. `AzureAppConfigurationOptions.SetClientFactory` method which allows user to set a `IAzureClientFactory<ConfigurationClient>`. It only works for `ConfigurationClient`

Image

We will not sync the configuration change made by the custom ConfigurationClient factory to `FeatureFlagClient`. (I am not sure whether this implementation is intended)

In theory, both ways could touch api version for `ConfigurationClient`. But in our current implementation for .NET provider, the wrong api version issue for feature flag client will never happen. (This is different from JS provider's current implementation)

linglingye001: > To me, the new feature flag change is not something opt-in. If a customer configured a specific api-version, and he upgraded to the latest provider version. That customer has no knowledge about the new feature flag change (e.g. new api version, new endpoint, etc) Suddenly, the provider will fail at every request sent by provider to get enhanced feature flag.
> 
> This case doesn't apply to .NET provider, because we don't have any public api to set client factory for `FaetureFlagClient`. But in JS provider, feature flag client options is [configurable](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/341/changes/BASE..a80420fe3b7427a0ead655a45ab7fc504aabaae2#r3819306399).

We can have doc/release note to cover the api version change and enhanced feature flag only available in `2026-05-01-Preview`.

Since all the SDK/Providers will support enhanced feature flags in preview version, if users do not want to use this feature, they can remain on the existing stable version.

For JS users, if they declare a dependency like `"@azure/app-configuration-provider": "^2.5.0"`, it will not automatically upgrade to a preview version. For users who explicitly set the API version, there is no risk of their applications being broken by an automatic upgrade. In other words, adopting Enhanced Feature Flags requires an intentional move to a preview version rather than happening implicitly through dependency updates.

jimmyca15 (approved):

src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationFeatureFlagClientFactory.cs
linglingye001: For version validation. 
1. The provider creates the default client options beforehand with V2026_05_01_Preview. Since `Version` is internal and read-only, the callback `ConfigureClientOptions` cannot change it.
image
2.  We can document that SetFeatureFlagClientFactory need to return clients using latest version V2026_05_01_Preview. The hard validation needs reflection which I don't think is necessary.

